PR #21, opened 2021-02-18
Add integration test script

abhipth:
*Description of changes:*
Add E2E Integration test script that runs integration test on existing cluster.

The script assumes that you have a Cluster that supports Pod Security Group Feature and have an instance type that supports ENI Trunking.

### DIFF
--- a/test/integration/run.sh
+++ b/test/integration/run.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+echo "Running VPC Resource Controller integration test with the following variables
+KUBE CONFIG: $KUBE_CONFIG_PATH
+CLUSTER_NAME: $CLUSTER_NAME
+REGION: $REGION"
+
+CLUSTER_INFO=$(aws eks describe-cluster --name $CLUSTER_NAME --region $REGION)
+
+VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
+SERVICE_ROLE_ARN=$(echo $CLUSTER_INFO | jq -r '.cluster.roleArn')
+ROLE_NAME=${SERVICE_ROLE_ARN##*/}
+ 
+echo "VPC ID: $VPC_ID, Service Role ARN: $SERVICE_ROLE_ARN, Role Name: $ROLE_NAME"
+
+# Set up local resources
+echo "Attaching IAM Policy to Cluster Service Role"
+aws iam attach-role-policy \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSVPCResourceController \
+    --role-name "$ROLE_NAME" > /dev/null
+
+echo "Enabling Pod ENI on aws-node"
+kubectl set env daemonset aws-node -n kube-system ENABLE_POD_ENI=true
+
+#Start the test
+echo "Starting the ginkgo test suite" 
+
+(cd $SCRIPT_DIR/perpodsg && CGO_ENABLED=0 GOOS=linux ginkgo -v -timeout 15m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
+
+#Tear down local resources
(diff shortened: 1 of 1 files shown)


test/integration/run.sh
```diff
@@ -0,0 +1,42 @@
...
+(cd $SCRIPT_DIR/perpodsg && CGO_ENABLED=0 GOOS=linux ginkgo -v -timeout 15m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
+
+#Tear down local resources
+echo "Detaching the IAM Policy from Cluster Service Role"
+aws iam detach-role-policy \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSVPCResourceController \
+    --role-name $ROLE_NAME > /dev/null
+
+echo "Disabling Pod ENI on aws-node"
+kubectl set env daemonset aws-node -n kube-system ENABLE_POD_ENI=false
```
haouc, line 40: My understanding is setting this to false wouldn't detach the trunk ENI and remove the label from instances. Maybe want to terminate the instances at the end of the tests? But adding this step can also add about 10 minutes to the test.

test/integration/run.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Set up local resources
+echo "Attaching IAM Policy to Cluster Service Role"
+aws iam attach-role-policy \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSVPCResourceController \
+    --role-name "$ROLE_NAME" > /dev/null
+
+echo "Enabling Pod ENI on aws-node"
+kubectl set env daemonset aws-node -n kube-system ENABLE_POD_ENI=true
+
```
haouc, line 27: Do we want to wait for at least 60 seconds to let aws-nodes restart?

test/integration/run.sh
```diff
@@ -0,0 +1,42 @@
...
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+echo "Running VPC Resource Controller integration test with the following variables
+KUBE CONFIG: $KUBE_CONFIG_PATH
+CLUSTER_NAME: $CLUSTER_NAME
+REGION: $REGION"
+
+CLUSTER_INFO=$(aws eks describe-cluster --name $CLUSTER_NAME --region $REGION)
```
haouc, line 11: Do we want to check if aws cli and kubectl are already installed?

test/integration/run.sh
abhipth: Setting this label to false would not detach the Trunk ENI. It will disable the vpc-resource-controller from creating new trunk interfaces and also not create any further branch interfaces on the pod. The purpose of this is to bring the state of the system to the closest state before starting the test. You are right re-scaling the node group would add a couple of minutes to the test which we want to avoid in our CI setup.

test/integration/run.sh
abhipth: Given that the Ginkgo suite only tests Pods using SGP. It will create the `Security Group Policy` and pods targeted by that SGP so in the worst case scenario when the `aws-node` doesn't start the Pods will be stuck in Pending State (as aws-node has not yet notified vpc-resource-controller to advertise the branch ENI capacity on the node) and once the aws-node finally restarts the Pod would go into Running state.

In short, the Ginkgo suite will not be impacted if the aws-node is still no started with the new environment variable.

test/integration/run.sh
abhipth: Our CI system which will use the script has these tools already installed.

test/integration/run.sh
haouc: Makes sense.

test/integration/run.sh
haouc: So test code in Ginkgo has the mechanism added to wait for pod till Running state?

test/integration/run.sh
haouc: Cool.

test/integration/run.sh
abhipth: Yes it waits for the Pod to go in Running state and if it doesn't then `-timeout 15m` when executing the tests will stop the timeout the failed test.

haouc (approved):
Looks good to me. Thanks.

PR #28, opened 2021-03-18
Annotate Fargate pod with matching security groups defined in SecurityGroupPolicy (CRD)

wangamel:
*Issue #, if available:*

*Description of changes:*

In VPC Resource Controller Mutating Webhook, Fargate pods will be filtered and annotated with matching security group defined in SecurityGroupPolicy (CRD) based on the pod selector or service account selector. It also means that after this change, VPC Resource Controller Mutating Webhook will only mutate non-fargate pods by injecting “vpc.amazonaws.com/pod-eni” into pod spec limit/request, except those with hostNetwork:True.

In this way, validation logic for pod security group on EKS Fargate is centralized with pods on node groups, which happens in VPC Resource Controller Mutating Webhook. The pod annotation here will support Fargate Scheduler to fetch pod security group information, and pass it to podENI when calling ECS runTask API.

### DIFF
--- a/webhook/core/annotation_validation_webhook.go
+++ b/webhook/core/annotation_validation_webhook.go
@@ -99,6 +99,13 @@ func (a *AnnotationValidator) handleUpdate(req admission.Request) admission.Resp
 		return admission.Allowed("")
 	}
 
+	if pod.Annotations[fargatePodSgAnnotKey] !=
+		oldPod.Annotations[fargatePodSgAnnotKey] {
+		logger.Info("denying annotation", "username", req.UserInfo.Username,
+			"annotation key", fargatePodSgAnnotKey)
+		return admission.Denied("annotation is not set by mutating webhook")
+	}
+
 	return admission.Allowed("")
 }
 
--- a/webhook/core/pod_webhook.go
+++ b/webhook/core/pod_webhook.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -29,6 +30,8 @@ import (
 )
 
 const resourceLimit = "1"
+const fargatePodSgAnnotKey = "fargate.amazonaws.com/pod-sg"
+const fargatePodLabel = "eks.amazonaws.com/fargate-profile"
 
 // +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=ignore,groups="",resources=pods,verbs=create,versions=v1,name=mpod.vpc.k8s.aws
 
@@ -76,14 +79,42 @@ func (prj *PodResourceInjector) Handle(ctx context.Context, req admission.Reques
 	//	pod.Spec.Containers[0].Resources.Limits[vpcresourceconfig.ResourceNameIPAddress] = resource.MustParse(resourceLimit)
 	//	pod.Spec.Containers[0].Resources.Requests[vpcresourceconfig.ResourceNameIPAddress] = resource.MustParse(resourceLimit)
 	//} else
+
+	isFargatePod := false
+	if _, ok := pod.ObjectMeta.Labels[fargatePodLabel]; ok {
+		isFargatePod = true
+	}
+	annotationMap := pod.ObjectMeta.Annotations
+	if annotationMap == nil {
+		annotationMap = make(map[string]string)
+	}
+
 	if sgList, cacheErr := prj.CacheHelper.GetPodSecurityGroups(pod); cacheErr != nil {
 		webhookLog.Error(cacheErr, "Webhook client failed to Get or List objects from cache.")
 		return admission.Denied("Webhood encountered error to Get or List object from k8s cache.")
-	} else if len(sgList) > 0 {
+	} else if len(sgList) > 0 && !isFargatePod {
 		webhookLog.Info("Injecting resource to the first container of the pod",
 			"resource name", vpcresourceconfig.ResourceNamePodENI, "resource count", resourceLimit)
 		pod.Spec.Containers[0].Resources.Limits[vpcresourceconfig.ResourceNamePodENI] = resource.MustParse(resourceLimit)
 		pod.Spec.Containers[0].Resources.Requests[vpcresourceconfig.ResourceNamePodENI] = resource.MustParse(resourceLimit)
+	} else if len(sgList) > 0 && isFargatePod {
+		// add pod-sg annotation if there is any matching security group in defined SGP (CRD)
+		annotationMap[fargatePodSgAnnotKey] = strings.Join(sgList, ",")
+		pod.ObjectMeta.Annotations = annotationMap
+		webhookLog.Info("Annotating fargate pod with security groups defined in CRD",
+			"Pod Security Group Annotation", annotationMap)
+	} else if len(sgList) == 0 && isFargatePod{
+		// Mutating Webhook catches the CREATE event when pod-sg annotation is added before this component
+		// Mutating Webhook will override pod-sg annotation if it is added before mutating webhook
+		// Annotation Validation Webhook catches the UPDATE event where pod-sg annotation is modified by any resource
+		if _, ok := pod.Annotations[fargatePodSgAnnotKey]; ok {
+			delete(annotationMap,fargatePodSgAnnotKey)
+			pod.ObjectMeta.Annotations = annotationMap
+			webhookLog.Info("Overriding pod-sg annotation added outside of mutating webhook",
+				"Pod Security Group Annotation", annotationMap)
+		} else {
+			return admission.Allowed("Fargate pod will not be annotated with security group.")
+		}
 	} else {
 		return admission.Allowed("Pod will not be injected with resources limits.")
 	}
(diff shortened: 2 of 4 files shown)


webhook/core/pod_webhook.go
```diff
@@ -76,14 +77,37 @@ func (prj *PodResourceInjector) Handle(ctx context.Context, req admission.Reques
...
 		pod.Spec.Containers[0].Resources.Limits[vpcresourceconfig.ResourceNamePodENI] = resource.MustParse(resourceLimit)
 		pod.Spec.Containers[0].Resources.Requests[vpcresourceconfig.ResourceNamePodENI] = resource.MustParse(resourceLimit)
+	} else if len(sgList) > 0 && isFargatePod {
+		// add pod-sg annotation if there is any matching security group in defined SGP (CRD)
+		annotationMap[fargatePodSgAnnotKey] = strings.Join(sgList, ",")
+		pod.ObjectMeta.Annotations = annotationMap
+		webhookLog.Info("Annotating fargate pod with security groups defined in CRD",
+			"Pod Security Group Annotation", annotationMap)
+	} else if len(sgList) == 0 && isFargatePod{
+		// override pod-sg annotation if it is not added outside of mutating webhook
```
wangamel, line 106: comment here is not correct, will be updated in the next revision. should be `// override pod-sg annotation if it is added before mutating webhook`

webhook/core/annotation_validation_webhook.go
```diff
@@ -84,6 +84,31 @@ func (av *AnnotationValidator) Handle(ctx context.Context, req admission.Request
 			}
 		}
 	}
+
+	// check if pod-sg annotation is updated by any resource
+	if podSgAnnot, ok := pod.Annotations["fargate.amazonaws.com/pod-sg"]; ok {
+		webhookLog.V(1).Info("Got annotation:", "fargate.amazonaws.com/pod-sg", podSgAnnot)
+		if req.Operation == v1beta1.Update {
+			webhookLog.V(1).Info("Operation is update, checking that pod-sg annotation wasn't modified")
+			oldPod, err := av.K8sWrapper.GetPodFromAPIServer(pod.Namespace, pod.Name)
```
abhipth, line 93: This adds a call to the API Server directly instead of getting the object from cache. Is there a reason why you wouldn't want to get the pod from cache instead?

Could you please also add the latency introduced by these changes under load. Example before this changes it was taking x milliseconds and after this change it was talking x+n milliseconds. Once we have the number n, we can be more confident the change is not introducing a large latency on updates.

We can sync up offline on how we can test this.

webhook/core/annotation_validation_webhook.go
```diff
@@ -84,6 +84,31 @@ func (av *AnnotationValidator) Handle(ctx context.Context, req admission.Request
...
+			if err != nil {
+				webhookLog.Error(err, "Failed to fetch pod in update request")
+				return admission.Denied("Validation failed. Trying to update annotation on a pod that can't be found")
+			}
+			if oldPodAnnotationValue, ok := oldPod.Annotations["fargate.amazonaws.com/pod-sg"]; ok {
+				// deny the request if pod-sg annotation is updated
+				if oldPodAnnotationValue != podSgAnnot {
```
abhipth, line 100: If you do decide to switch to get the pod object form cache instead. This is the sequence of event that may lead to an update to fail when it should go through.

Consider the sequence of events for a pod with same namespace/name. 

- Pod is terminating.
- Pod event received with deletion timestamp set.
- New Pod is created with same namespace/name.
- The cache could return the older pod (that is already deleted) which may have a different annotation then the current pod and then the event would fail.

Even though this is a rare scenario and not sure you could hit this in webhooks. But it would be worth adding extra guards like checking the UID when comparing older pod to a newer pod.

Reference: This kind of bug bit us in another controller and it's super hard to reproduce.

webhook/core/pod_webhook.go
abhipth: nit: can you also add a bit of details on the reason behind it in the comment if possible. Something like the validating webhook will catch all the update on the given annotation key and the mutating webhook will catch create events outside of the webhook.

webhook/core/pod_webhook_test.go
```diff
@@ -153,6 +174,15 @@ func TestPodResourceInjector_Handle(t *testing.T) {
 	}
 }
 
+// TestFargatePodSG_Handle test webhook annotating fargate pod with sg list.
+func TestFargatePodSG_Handle(t *testing.T) {
+	pod := NewFargatePod("test", "test_fp", "test_namespace")
+	assert.True(t, pod.ObjectMeta.Labels["eks.amazonaws.com/fargate-profile"] == "test_fp")
+	injector := getFargatePodResourceInjectorWithSGP()
+	resp := getResponse(pod, injector)
+	assert.True(t, resp.Allowed)
```
abhipth, line 183: Can we also verify the pod got annotated with the expected SG.

abhipth: If possible, could you also add test cases for following scenarios.

- If fargate pod's annotation is updated outside the mutating webhook. --> It should be denied
- If fargate pod is created with the sgp annotation. --> It should be denied
- All non fargate pod should never have this annotation.

abhipth (commented):
Added some initial review comments. Please let me know your thoughts on the changes.

webhook/core/pod_webhook.go
```diff
@@ -76,14 +79,38 @@ func (prj *PodResourceInjector) Handle(ctx context.Context, req admission.Reques
...
+		// Mutating Webhook catches the CREATE event when pod-sg annotation is added before this component
+		// Mutating Webhook will override pod-sg annotation if it is added before mutating webhook
+		// Annotation Validation Webhook catches the UPDATE event where pod-sg annotation is modified by any resource
+		delete(annotationMap,fargatePodSgAnnotKey)
+		pod.ObjectMeta.Annotations = annotationMap
+		webhookLog.Info("Overriding pod-sg annotation added outside of mutating webhook",
+			"Pod Security Group Annotation", annotationMap)
 	} else {
 		return admission.Allowed("Pod will not be injected with resources limits.")
 	}
```
haouc, line 116: Can you separate the logic to two parts each of which can take care of non Fargate or Fargate code flow?
Such as this may help readability.
```
sgList, cacheErr := prj.CacheHelper.GetPodSecurityGroups(pod); cacheErr != nil {
  return admission.Denied("...")
}
if !isFargatePod {
  if sgList > 0 {
  } else {
  }
}
if isFargatePod {
  if sgList > 0 {
  } else {
  }
}
```

webhook/core/pod_webhook_test.go
```diff
@@ -153,6 +192,48 @@ func TestPodResourceInjector_Handle(t *testing.T) {
 	}
 }
 
+// TestFargatePodResourceInjector_Handle test webhook annotating fargate pod with sg list.
+func TestFargatePodResourceInjector_Handle(t *testing.T) {
```
haouc, line 196: This is not necessary for this PR. But I think we should start using table driven tests to avoid repeated codes and improve readability and cases coverage. when we add or rewrite tests in this file. I think we have some in this project already.
https://github.com/golang/go/wiki/TableDrivenTests

webhook/core/pod_webhook.go
```diff
@@ -76,14 +79,40 @@ func (prj *PodResourceInjector) Handle(ctx context.Context, req admission.Reques
...
+	} else if len(sgList) == 0 && isFargatePod{
+		// Mutating Webhook catches the CREATE event when pod-sg annotation is added before this component
+		// Mutating Webhook will override pod-sg annotation if it is added before mutating webhook
+		// Annotation Validation Webhook catches the UPDATE event where pod-sg annotation is modified by any resource
+		if _, ok := pod.Annotations[fargatePodSgAnnotKey]; ok {
+			delete(annotationMap,fargatePodSgAnnotKey)
+			pod.ObjectMeta.Annotations = annotationMap
+			webhookLog.Info("Overriding pod-sg annotation added outside of mutating webhook",
+				"Pod Security Group Annotation", annotationMap)
+		}
```
abhipth, line 115: For the use case where len(sgList) == 0 & it's fargate pod & it doesn't have the  `fargate.amazonaws.com/pod-sg` we are still patching the pod even though we didn't do any modification.

Should we have another else condition here to allow admission if the annotation doesn't exist?

wangamel: I think here is where we annotate the fargate pods with `fargate.amazonaws.com/pod-sg` if len(sgList) > 0 & it's fargate pod. Does the use case you mentioned go to the `else` block below? which just allow the admission?

wangamel: Sorry, misunderstood the above comment, yes I will add another `else` to allow the admission

webhook/core/pod_webhook_test.go
```diff
@@ -153,6 +192,54 @@ func TestPodResourceInjector_Handle(t *testing.T) {
...
+		annotMap = make(map[string]string)
+	}
+	annotMap[fargatePodSgAnnotKey] = "sg-123"
+	pod.ObjectMeta.Annotations = annotMap
+	assert.True(t, pod.ObjectMeta.Labels[fargatePodLabel] == "test_fp")
+	injector := getFargatePodResourceInjectorWithoutSGP()
+	resp := getResponse(pod, injector)
+	assert.True(t, resp.Allowed)
+	for _, p := range resp.Patches {
+		assert.True(t, p.Operation == "remove")
```
abhipth, line 224: I think the test case is missing check that the annotation is successfully removed in the response?

abhipth (approved):
LGTM!

PR #32, opened 2021-04-05
Get old pod from request in annotation webhook instead of cache

abhipth:
*Issue #, if available:*
Switched to get the old pod from the webhook request instead of getting it from the cache.

*Description of changes:*
- get old pod from request instead of using cache
- add unit test cases for annotation webhook
- add integration test suite for validating annotation webhook

### DIFF
--- a/test/framework/manifest/pod.go
+++ b/test/framework/manifest/pod.go
@@ -28,6 +28,7 @@ type PodBuilder struct {
 	container              v1.Container
 	os                     string
 	labels                 map[string]string
+	annotations            map[string]string
 	terminationGracePeriod int
 	restartPolicy          v1.RestartPolicy
 }
@@ -35,9 +36,10 @@ type PodBuilder struct {
 func (p *PodBuilder) Build() (*v1.Pod, error) {
 	return &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      p.name,
-			Namespace: p.namespace,
-			Labels:    p.labels,
+			Name:        p.name,
+			Namespace:   p.namespace,
+			Labels:      p.labels,
+			Annotations: p.annotations,
 		},
 		Spec: v1.PodSpec{
 			ServiceAccountName:            p.serviceAccountName,
@@ -56,6 +58,7 @@ func NewDefaultPodBuilder() *PodBuilder {
 		container:              NewBusyBoxContainerBuilder().Build(),
 		os:                     "linux",
 		labels:                 map[string]string{},
+		annotations:            map[string]string{},
 		terminationGracePeriod: 0,
 		restartPolicy:          v1.RestartPolicyNever,
 	}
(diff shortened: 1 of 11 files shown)


webhook/core/annotation_validation_webhook.go
```diff
@@ -30,69 +30,80 @@ import (
...
-				// This is an update trying to add the pod annotation
-				if (req.UserInfo.Username != validUserInfo) && (req.UserInfo.Username != newValidUserInfo) {
-					webhookLog.Info("Denying adding annotation", "Username", req.UserInfo.Username)
-					return admission.Denied("Validation failed. Pod ENI annotation added outside of VPC Resource Controller")
-				}
-			}
+	// This will block any update on the specific annotation from non vpc resource controller
+	// service accounts
+	if pod.Annotations[config.ResourceNamePodENI] !=
+		oldPod.Annotations[config.ResourceNamePodENI] {
```
wangamel, line 93: Just to confirm, if `config.ResourceNamePodENI` does not exist in annotation, is it going to return empty string? so we don't check if it exist or not?

abhipth: This statement is supposed to cover the following update scenarios by a non vpc-resource-controler user.
- If the `config.ResourceNamePodENI` is added/updated by any user it should be blocked.
- If the `config.ResourceNamePodENI` is removed by any user it should be blocked.

We could definitely check if the annotation exists or not. Do you think with the current check we could miss some corner cases?

haouc: Could this case happen?
1, a pod update event coming
2, the pod has same annotations during the update
3, the pod has different usernames

test/framework/manifest/pod.go
```diff
@@ -101,6 +104,11 @@ func (p *PodBuilder) Labels(labels map[string]string) *PodBuilder {
 	return p
 }
 
+func (p *PodBuilder) Annotations(annotations map[string]string) *PodBuilder {
+	p.annotations = annotations
+	return p
+}
+
```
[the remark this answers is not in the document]
abhipth: This is the pattern so you could do something like
```
pod := PodBuilder{}.Name(name).Annotation(anotations-map).Labels(label-map).Build()
```

webhook/core/annotation_validation_webhook.go
abhipth: Yes this could happen. If the annotations are same and it's an update event from a different username, it should go through.

I'll add another integration test just to be on safe side.

haouc (approved):
lgtm

test/framework/resource/k8s/pod/manager.go
```diff
@@ -36,6 +36,7 @@ type Manager interface {
 	DeleteAndWaitTillPodIsDeleted(context context.Context, pod *v1.Pod) error
 	GetENIDetailsFromPodAnnotation(podAnnotation map[string]string) ([]*trunk.ENIDetails, error)
 	GetPodsWithLabel(context context.Context, namespace string, labelKey string, labelValue string) ([]v1.Pod, error)
+	PatchPod(context context.Context, oldPod *v1.Pod, newPod *v1.Pod) error
```
M00nF1sh, line 39: why not patch directly using k8sClient..since the logic is pretty simple, i don't see the value of add this wrapper :D

M00nF1sh (commented):
overall /lgtm

M00nF1sh (approved):

PR #39, opened 2021-07-23
Deprecate v1beta1 in crd and webhooks

haouc:
*Issue #, if available:*

*Description of changes:*
Since kubernetes are removing service to version v1beta1, we want to deprecate API version v1beta1 and start using v1 in CustomResourceDefinition and webhooks. 

Tests

- Applied updated CRD to existing cluster
- Applied updated webhooks to existing cluster

No errors were observed.


abhipth:
Will be addressed by #57

PR #40, opened 2021-07-28
add CI setup to run integration test on new PR

abhipth:
*Description of changes:*
Adds GitHub Actions for running Integration test using eksctl cluster on self hosted runners.

Each PR or a commit on existing PR will get `disable-testing` label added to it. Once the label is removed, then the integration tests can be run. 

Ideally, we should have used something like `ok-to-test` but all GitHub provided actions at this point don't support removing existing labels from PR. We need to do this when user commits to existing PR so we don't re run the tests without verifying it first.

See `scripts/test/README.md` for more details.


M00nF1sh (approved):
nice work :D

PR #45, opened 2021-08-17
Update to pull_request_target

vikasmb:
Restricted github_token permissions in integ_test workflow.


vikasmb:
Addressed in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/47

Issue #33, opened 2021-04-27
VPCResourceController should release branch-eni resources for terminated Pods.


M00nF1sh:
**What would you like to be enhanced**:
Pods created by Jobs will enter terminated state(Succeed/Failed), and the scheduler will treat resources used by terminated pods are released when making scheduling decision around resource limits.
For normal CNI workflow, kubelet will invoke DelNetwork for terminated pods, and our normal CNI will correctly reuse IP for new pods. 
However, for VPCResourceController, the branch-eni resource is still holding by terminated pods, and scheduler might schedule pod onto nodes where there are no branch-eni resource available. 

**Why is the change needed and what use case will it solve**:
We should change VPCResourceController to release the branch-eni resource for terminated pods as well.

Shaiou:
This would be really appreciated indeed. We've run into the issue many time where a pod would be scheduled but stuck in "creating" mode just because it was on a node with a terminated pod/job that did not release this eni.

0xkag:
Enterprise customer here.  Also experiencing this issue.

JacobHenner:
This is causing a decent bit of operational pain, and confusion among our developers. Is there an ETA for when this might be addressed?

abhipth:
@JacobHenner Sorry for the inconvenience. This will be fixed in v1.1.0, which is planned for release in October.

backjo:
hi all - is this issue expected to occur on the Windows VPC Resource Controller that is currently used for enabling windows in https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html - the same type of issue seems to be manifesting for Windows CronJob pods that we've got running.

abhipth:
@backjo yes the issue is present in the Windows VPC Resource Controller as well. The old Windows controller is being deprecated and the Windows functionality will now be available in this controller from v1.1.0 release. The release will also have the fix for the issue you mentioned.

Sorry for the inconvenience, if you are on v1.21 then you can consider the [TTL](https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically) for cleaning up the Succeeded/Failed Jobs.

backjo:
Hey @abhipth  - thanks for the confirmation, this is super helpful. We're already using the TTL setting to help mitigate but wanted to confirm that this was a known IP leak on Windows.

abhipth:
[...]

PR #48, opened 2021-08-24
remove networking for succeeded/failed pods

abhipth:
*Issue #, if available:* #33

*Description of changes:*

Removes networking for Pods in the phase `Succeeded` or `Failed` for Both Windows(Not enabled yet) and Linux Pods. [See](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/)

List of Changes
- Add code for bug fix
- Add new integration tests to test the same
- Run integration test in CI

### DIFF
--- a/controllers/core/pod_controller.go
+++ b/controllers/core/pod_controller.go
@@ -47,6 +47,8 @@ type PodReconciler struct {
 // if the resource is supported by the controller.
 func (r *PodReconciler) Reconcile(request custom.Request) (ctrl.Result, error) {
 	var isDeleteEvent bool
+	var hasPodCompleted bool
+
 	var pod *v1.Pod
 
 	if request.DeletedObject != nil {
@@ -67,6 +69,11 @@ func (r *PodReconciler) Reconcile(request custom.Request) (ctrl.Result, error) {
 		// convert to pod object
 		pod = obj.(*v1.Pod)
 	}
+
+	// If Pod is Completed, the networking for the Pod should be removed
+	// given the container will not be restarted again
+	hasPodCompleted = pod.Status.Phase == v1.PodSucceeded ||
+		pod.Status.Phase == v1.PodFailed
 
 	logger := r.Log.WithValues("UID", pod.UID, "pod", request.NamespacedName,
 		"node", pod.Spec.NodeName)
@@ -95,7 +102,7 @@ func (r *PodReconciler) Reconcile(request custom.Request) (ctrl.Result, error) {
 
 		var err error
 		var result ctrl.Result
-		if isDeleteEvent {
+		if isDeleteEvent || hasPodCompleted {
 			result, err = resourceHandler.HandleDelete(pod)
 		} else {
 			result, err = resourceHandler.HandleCreate(int(totalCount), pod)
@@ -104,7 +111,8 @@ func (r *PodReconciler) Reconcile(request custom.Request) (ctrl.Result, error) {
 			return result, err
 		}
 		logger.V(1).Info("handled resource without error",
-			"resource", resourceName, "is delete event", isDeleteEvent)
+			"resource", resourceName, "is delete event", isDeleteEvent,
+			"has completed pod", hasPodCompleted)
 	}
 
 	return ctrl.Result{}, nil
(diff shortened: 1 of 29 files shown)


controllers/core/pod_controller.go
```diff
@@ -104,7 +111,8 @@ func (r *PodReconciler) Reconcile(request custom.Request) (ctrl.Result, error) {
 			return result, err
 		}
 		logger.V(1).Info("handled resource without error",
-			"resource", resourceName, "is delete event", isDeleteEvent)
+			"resource", resourceName, "is delete event", isDeleteEvent,
+			"has completed pod", hasPodCompleted)
```
M00nF1sh, line 115: nit, `has pod completed`

pkg/provider/branch/provider.go
```diff
@@ -404,7 +404,10 @@ func (b *branchENIProvider) DeleteBranchUsedByPods(nodeName string, UID string)
 
 	err := trunkENI.PushBranchENIsToCoolDownQueue(UID)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to delte branch eni used by the pod %s", UID)
```
M00nF1sh, line 407: seems this is the only place we call PushBranchENIsToCoolDownQueue, can we make it no return value instead?
it seems better to me that not handle the error here.

test/framework/manifest/container.go
```diff
@@ -70,12 +71,20 @@ func (w *Container) Args(arg []string) *Container {
 	return w
 }
 
+func (w *Container) AddContainerPort(containerPort int) *Container {
```
M00nF1sh, line 74: feels better to accept a containerPort struct :D

test/framework/resource/k8s/deployment/manager.go
```diff
@@ -40,6 +41,39 @@ type defaultManager struct {
 	k8sClient client.Client
 }
 
+func (m *defaultManager) ScaleDeployment(ctx context.Context, name string, namespace string, replicas int) error {
+	var deployment *appsv1.Deployment
```
M00nF1sh, line 45: will this ever work? since deployment == nil

test/framework/resource/k8s/jobs/manager.go
```diff
@@ -39,6 +43,44 @@ type defaultManager struct {
...
 
+func (m *defaultManager) CreateJobAndWaitForJobToRun(ctx context.Context,
+	jobs *batchV1.Job) (*batchV1.Job, error) {
+	err := m.k8sClient.Create(ctx, jobs)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait till the cache is refreshed
+	time.Sleep(utils.PollIntervalShort)
```
M00nF1sh, line 54: we can use PollUntil instead of PollImmediateUntil, to avoid the sleep here :D

test/framework/resource/k8s/jobs/manager.go
```diff
@@ -39,6 +43,44 @@ type defaultManager struct {
...
+			return false, fmt.Errorf("some jobs has either"+
+				" succeeded or failed :%v", observedJob.Status)
+		} else if observedJob.Status.Active > 0 {
+			var count int
+			for _, pod := range podList.Items {
+				if pod.Status.Phase == v1.PodRunning && strings.Contains(pod.Name, jobs.Name) {
+					count++
+				}
+			}
+			return count == int(*jobs.Spec.Parallelism), nil
```
M00nF1sh, line 78: will apiserver automatically fills this Parallelism? wondering whether it's always non-nil

abhipth: I tried creating a Job without specifying parallelism and I could see in the spec that the parallelism was auto populated.
```
spec:
  backoffLimit: 4
  completions: 1
  parallelism: 1
```

test/framework/resource/k8s/jobs/manager.go
abhipth: Nice! That's much cleaner. Replaced to PollUntil everywhere.

test/framework/resource/k8s/deployment/manager.go
abhipth: looks like this is unused code, good catch!

Issue #17, opened 2021-02-04
arm64 support


cudders:
Hi,

We're investigating using arm64 node groups in EKS and have a lot of windows container workloads.
We'd need an arm64 build, same for vpc-admission-webhook I guess, is this on the roadmap?

Many thanks,
Chris

mikestef9:
We are planning on moving this controller to the Kubernetes control plane for EKS clusters, so it's not a component you will need to manually install there. Adding Arm support is also something we we would like to do though, so will leave this issue open to track this feature request.

cudders:
Perfect, thanks Mike.

iusergii:
@mikestef9 is there any issue we can track for moving controller to control plane?

abhipth:
@iusergii we are tracking it using this issue - https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/4

abhipth:
Windows Support is now available on EKS Control Plane. It is available on EKS Platform versions mentioned in [v1.1.0 release](https://github.com/aws/amazon-vpc-resource-controller-k8s/releases/tag/v1.1.0)

PR #50, opened 2021-08-27
Enable windows support in controller by default

abhipth:
*Issue #, if available:* #4, #17 

*Description of changes:* 
Enabling Support for Windows IPv4 Address Management in the new controller by default.

Fix the following issues by introducing Re-Sync API in Warm Pool.
- If User manually deletes the resources, the Pool should re-sync instead of failing continuously.
- If the Async Resource creation times out, the Pool should re-sync with upstream resources.
- Fix Issue where Pod with `terminationGracePeriodSeconds` is zero and Pod is terminated immediately after creation, the Pod Annotation on delete event will not contain the Resource and the resource would be leaked. 
- Fix issue where Pod with same namespace/name is annotated resource of the previous Pod.

Automation Tests
- Stress Test
- Integration Tests
- Add Integration tests to CI 

In next PR, we will expose a flag to enable this feature via `ConfigMap`


orsenthil:
This https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/50#discussion_r709679715 was the major comment. I reviewing rest of the code in pkg, and will be done tomorrow. 

Thank you.

webhooks/core/pod_webhook.go
```diff
@@ -89,6 +82,7 @@ func (i *PodResourceInjector) Handle(ctx context.Context, req admission.Request)
 		annotationMap = make(map[string]string)
 	}
 
+	// TODO: Stop processing events if ENABLE_POD_ENI is set to true + No trunk Node
```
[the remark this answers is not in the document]
abhipth: I agree, this is one part of the code that I eagerly wanted to simplify! Will work on it

pkg/provider/ip/provider.go
```diff
@@ -196,6 +200,24 @@ func (p *ipv4Provider) CreatePrivateIPv4AndUpdatePool(job *worker.WarmPoolJob) {
 	p.updatePoolAndReconcileIfRequired(instanceResource.resourcePool, job, didSucceed)
 }
 
+func (p *ipv4Provider) ReSyncPool(job *worker.WarmPoolJob) {
+	providerAndPool, found := p.instanceProviderAndPool[job.NodeName]
+	if !found {
+		p.log.Error(fmt.Errorf("instnace provider not found"), "node is not initialized",
```
orsenthil, line 206: _instance_

scripts/test/test-with-eksctl.sh
```diff
@@ -129,26 +130,32 @@ function set_pod_eni_flag_on_ipamd() {
 }
 
 function run_inegration_test() {
```
orsenthil, line 132: I think, this was meant as `run_integration_test`

test/framework/utils/utils.go
```diff
@@ -25,3 +25,11 @@ func NamespacedName(obj metav1.Object) types.NamespacedName {
 		Name:      obj.GetName(),
 	}
 }
+
+func CopyMap(original map[string]string) map[string]string {
```
orsenthil, line 29: TODO for later - These `utils` _usually_ could be shared across the project instead of one in tests and src.
In this case, I notice these aren't test specific too.

orsenthil (commented):
@abhipth  - I am done with my review of this.

- Except for one refactor suggestion, everything looks good to me.
- Exercising this by integration and manual testing might provide more feedback on this. A little hard to determine the effects by code review for this change.

Thank you!

pkg/handler/on_demand.go
```diff
@@ -48,11 +48,6 @@ func (h *onDemandResourceHandler) HandleCreate(requestCount int, pod *v1.Pod) (c
 
 // HandleDelete reclaims the on demand resource by passing the Delete Job to the respective Worker
 func (h *onDemandResourceHandler) HandleDelete(pod *v1.Pod) (ctrl.Result, error) {
-	if _, ok := pod.Annotations[h.resourceName]; !ok {
```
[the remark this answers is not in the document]
abhipth: Yeah, this is done for cases when termination grace period is set to 0 and Pod is immediately created and deleted. In that case I observed the delete event didn't have the annotation. This will query the local DS to see if Pod had been allocated a resource.

abhipth:
Would appreciate review of the file [[Link](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/5cd17167d343ea22fb0f825dd51c60dadca22986/webhooks/core/pod_webhook.go)]  from original component owners.

cc @wangamel (for fargate) @haouc (for sgp)

webhooks/core/pod_webhook.go
```diff
@@ -22,145 +22,207 @@ import (
...
+
+	return response
+}
+
+// HandleWindowsPod mutates the Windows Pod by injecting a secondary IPv4 Address
+// Limit to the Pod when the Windows IPAM feature is enabled via ConfigMap
+func (i *PodMutationWebHook) HandleWindowsPod(req admission.Request, pod *corev1.Pod,
+	log logr.Logger) (response admission.Response) {
+
+	// TODO: Don't Process event if feature is disabled
```
orsenthil, line 142: Will this function be called if the feature is disabled at the moment? Is it safe to keep the processing open?

orsenthil (approved):
LGTM. 

I have the question with `HandleWindowsPod` when the feature is disabled, please comment on that. Thank you.

webhooks/core/pod_webhook.go
abhipth: We have kept it allow all for now because of integration test. @sushrk is working on a follow up PR that will add the relevant checks here.

test/framework/utils/utils.go
abhipth: Moved it to utils.

pkg/pool/pool.go
```diff
@@ -85,6 +91,69 @@ func NewResourcePool(log logr.Logger, poolConfig *config.WarmPoolConfig, usedRes
...
+// ReSync syncs state of upstream with the local pool. If local resources have additional
+// resource which doesn't reflect in upstream list then these resources are removed. If the
+// upstream has additional resources which are not present locally, these resources are added
+// to the warm pool. During ReSync all Create/Delete operations on the Pool should be halted
+// but Assign/Free on the Pool can be allowed.
+func (p *pool) ReSync(upstreamResource []string) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	// This is possible if two Re-Syn were requested at same time
```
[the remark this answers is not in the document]
abhipth: I don't think it's possible, but still wanted to add a safeguard for now.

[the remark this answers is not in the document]
orsenthil: > but still wanted to add a safeguard for now.

Once you feel confident on other parts of the code, you could remove this. Otherwise, it can cause confusion to the reader further down the line.

PR #52, opened 2021-09-01
Process Pod event immediately after the Node is initialized instead of waiting for next Resync Period

abhipth:
**Issue #, if available:**
Currently on Controller startup, the sequence of event happens,

- Pod Controller starts, we receive all Pods events.
- Node are not yet initialized, so all Pod events are ignored.
- All Nodes are Initialized.
- Pod Controller re-syncs after 30 minutes and the Previously ignored Pod events are processed.

 This adds a 30 minutes (resync duration) delay to all missed Pod events during controller restarts.

**Description of changes:**

The proposed fix is to re-queue pod event and add after 1 seconds, till the Node is initialized so the Pod event is not ignored for 30 minutes.


controllers/core/node_controller.go
```diff
@@ -45,12 +45,13 @@ type NodeReconciler struct {
 func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
```
orsenthil, line 45: A comment about the situations in which UpdateNode, and AddNode is done, in the reconcile method might be helpful here. It seems like an important behavior change in reconcile.

controllers/core/node_controller.go
```diff
@@ -60,13 +61,16 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	err := r.Manager.AddOrUpdateNode(node)
-	if err != nil {
-		logger.Error(err, "failed to add node")
-		return ctrl.Result{}, err
+	_, found := r.Manager.GetNode(req.Name)
```
orsenthil, line 64: Do any tests exercise these changes in `node_controller` ?

orsenthil:
Hi @abhipth  - I am done with my first pass of the review. The code looks good and test coverage with mock is also good.

I have two other comments.

- Integration tests. This code is bringing a functionality, that might be best covered by integration tests.  Do you have any thoughts on this?

- I realized that this is against release-1.1 branch. I noticed the divergence with `master` increasing with this PR. Do you have thoughts or ideas on how and when bridge the gap. I hope these changes will go into master as well.

pkg/node/manager/manager.go
```diff
@@ -0,0 +1,358 @@
...
+		log.Info("node was being managed earlier, will be added as un-managed node now")
+		// Change the node in cache, but for de initializing all resource providers
+		// pass the async job the older cached value instead
+		m.dataStore[nodeName] = node.NewUnManagedNode()
+		op = Delete
+	} else if !cachedNode.IsManaged() {
+		log.V(1).Info("node not managed, no operation required")
+		return nil
+	} else {
+		err = m.updateSubnetIfUsingENIConfig(cachedNode, k8sNode)
```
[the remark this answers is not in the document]
abhipth: I added a comment about this, this only relates to a Managed Node. This doesn't fit in here as it's resource specific while Node Manager is generic. We may have to fix this in future.

controllers/core/node_controller.go
abhipth: This was missing, I added Unit tests for this.

orsenthil (commented):
I have only one critical / important question, commented here - https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/52/commits/13abdf07baf2cee5b22ca53489bf885c9b63c100#r701266281

Rest looks good to me.

pkg/node/manager/manager.go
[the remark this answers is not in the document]
abhipth: Good catch! This should be for node update + when node is toggled to managed. Will fix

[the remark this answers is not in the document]
orsenthil: The pattern of `if,elseif, else` and made be a bit curious.  Breaking into smaller conditionals or switch could be easier to read.

Thank you.

scripts/test/test-with-eksctl.sh
```diff
@@ -268,6 +268,10 @@ verify_controller_has_lease
 # the node for ENI Trunking/Branching
 set_pod_eni_flag_on_ipamd "true"
 
+# Allow for IPAMD to label the node after startup
+# TODO: Handle this in the Test Suite in more concrete manner
```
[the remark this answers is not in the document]
abhipth: Yes, this is hacky for now. IPAMD on startup sets the Label on the Nodes `vpc.amazonaws.com/has-trunk-attached` this allows for the controller to start managing the nodes. 

In our new Ginko test suite, we have one test that expects all the Node to have this label otherwise fails. IPAMD, takes some time to add this label and if that test is first to run then the suite fails. Will fix this properly soon.

pkg/node/manager/manager.go
```diff
@@ -205,6 +218,20 @@ func (m *manager) UpdateNode(nodeName string) error {
 	return nil
 }
 
+func (m *manager) GetNodeUpdateStatus(k8sNode *v1.Node, cachedNode node.Node) NodeUpdateStatus {
```
orsenthil, line 221: _minor_ - Can be refactored when visiting this again. 

a) Since all of these are have return statement, else and else is not required.
b) Also, cache check could have happen first, and then isSelectedForManagement.  (For the reader, that seems a bit easier to reason about).

orsenthil (approved):
LGTM. Thanks for additional fix and inclusion of the tests.

pkg/node/manager/manager.go
abhipth: Have taken up an AI to fix this in next versions.

PR #44, opened 2021-08-12
remove workarounds from custom pod controller and refactor code

abhipth:
*Description of changes:*

- Removing workaround from the memory optimized Pod controller to pass Delete Event with request.
- Renaming file and structs, interfaces names to more descriptive names.
- Refactor code so it's easier to add new resource like IPv4 Address on top of Pod ENI.

**Testing for the Change**
- Ran Integration test in loop (x5 times) - Passed
- Ensure memory doesn't increase as the Number of Pods increases - Passed
- Ensure existing Pod's networking is not removed once the controller restarts - Passed

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #56, opened 2021-09-09
Fix regression in custom controller

abhipth:
*Description of changes:*

With the #44 a regression was introduced where `Delete` and `Update` events were being processed by Pod reconciler at same time. 

**Why were the events being processed at same time?**
The Delete Request had the Pod Object, whereas the Add/Update request had Namespace/Name. Since these two requests won't hash to same value. The `workqueue` will consider these as different objects and send their events in parallel. 

**How was regression detected?**
It was detected while writing the stress test for enabling Windows. The test observed some leaked IP Resources because the Delete event was processed before the Update event.

**How do we fix it?**
Reverting back to the old way of doing things, where we don't modify the request but store the delete object in cache. And delete it once event is processed. Previously we were doing it using two different channels for Add/Update Event and Delete Events making code a little hard to read. Now switched to using a single channel.

### DIFF
--- a/controllers/core/pod_controller.go
+++ b/controllers/core/pod_controller.go
@@ -14,12 +14,17 @@
 package controllers
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/controllers/custom"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s/pod"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node/manager"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/resource"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
 	v1 "k8s.io/api/core/v1"
@@ -39,7 +44,12 @@ type PodReconciler struct {
 	// Manager manages all the nodes on the cluster
 	NodeManager manager.Manager
 	// DataStore is the cache with memory optimized Pod Objects
-	DataStore       cache.Indexer
+	DataStore cache.Indexer
+	// DeletedObjDataStore stores the objects deleted from etcd, reconciler
+	// must remove the object after accessing it
+	DeletedObjDataStore cache.Indexer
+	// DataStoreSynced when set to true indicates the data store is synced
+	// successfully
 	DataStoreSynced *bool
 }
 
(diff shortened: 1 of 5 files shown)


controllers/core/pod_controller.go
```diff
@@ -152,17 +174,39 @@ func getAggregateResources(pod *v1.Pod) map[string]int64 {
 func (r *PodReconciler) SetupWithManager(manager ctrl.Manager,
 	clientSet *kubernetes.Clientset, pageLimit int, syncPeriod time.Duration) error {
 
-	return custom.NewControllerManagedBy(manager).
+	// event chanel is used by the custom controller to send Pod events after optimizing
```
orsenthil, line 177: `event channel`

controllers/core/pod_controller.go
```diff
@@ -152,17 +174,39 @@ func getAggregateResources(pod *v1.Pod) map[string]int64 {
...
-			ResyncPeriod:            syncPeriod,
+			PageLimit:    pageLimit,
+			ResyncPeriod: syncPeriod,
+		}).Complete(sourceEventChan)
+	if err != nil {
+		return err
+	}
+
+	// Start the high level controller, low level controller will notify this controller
+	// using a chanel
```
orsenthil, line 201: `channel`

controllers/custom/custom_controller.go
```diff
@@ -79,79 +68,39 @@ type Options struct {
...
+	c.controller.Run(stop)
+
 	return nil
 }
 
-// WaitForCacheSync tills the cache has synced, this must be done under
-// mutex lock to prevent other controllers from starting at same time
-func (c *CustomController) WaitForCacheSync(controller cache.Controller) {
-	for !controller.HasSynced() && controller.LastSyncResourceVersion() == "" {
+// MarkDataStoreSynced sets the dataStoreSynced variable to true when controller has
```
orsenthil, line 96: I hope there is a test for this. This seems like a very tricky function here catering to few situations like nil or waiting for sync.

orsenthil (commented):
I have a question on `MarkDataStoreSynced` - rest of the code looks good to me.

abhipth:
I was able to reproduce the issue with these changes as well which leads mt to this regression was already present.

Here's how to reproduce the issue. Create deployment with Pods having `TerminationGracePeriodSeconds` set to 0. Scale up the deployment and immediately scale it down.

For some of the Pod event the following behavior is seen.

- Receive Add event. The Pod is annotated with resource.
- Receive Update event without any annotation.
- Receive Delete event without any annotation.

All of these events happen within 200ms. I couldn't find any upstream issues on why Delete event doesn't contain the updated annotation. Also looked at new version release of client-go and didn't see any related issues there.

Will handle this by querying the annotation from the local DS instead of relying on annotation.

orsenthil:
> Will handle this by querying the annotation from the local DS instead of relying on annotation.

SGTM.

PR #57, opened 2021-09-09
Update project dependencies

sushrk:
*Issue #, if available:*
N/A
*Description of changes:*
Updating project dependencies- go version, controller-runtime, amazon-vpc-cni. 


internal/aws-sdk-go/go.mod
```diff
@@ -1,10 +1,5 @@
 module github.com/aws/aws-sdk-go
 
-require (
```
abhipth, line 3: Why are we changing the internal package's go.mod? Was it not working without the change?

main.go
```diff
@@ -177,17 +178,19 @@ func main() {
...
+		Scheme:                     scheme,
+		MetricsBindAddress:         metricsAddr,
+		Port:                       9443,
+		LeaderElection:             enableLeaderElection,
+		LeaseDuration:              &leaseDuration,
+		RenewDeadline:              &renewDeadline,
+		RetryPeriod:                &retryPeriod,
+		LeaderElectionID:           config.LeaderElectionKey,
+		LeaderElectionNamespace:    config.LeaderElectionNamespace,
+		LeaderElectionResourceLock: resourcelock.ConfigMapsResourceLock,
```
abhipth, line 191: What's the advantage/downside of using the default leader election locking?

controllers/custom/custom_controller.go
```diff
@@ -108,7 +109,7 @@ type Request struct {
 }
 
 // Starts the low level controller
-func (c *CustomController) Start(stop <-chan struct{}) error {
+func (c *CustomController) Start(ctx context.Context) error {
```
abhipth, line 112: I think this is broken from the start, context.TODO() and context.Background() both return an empty context. We need to start using `ctrl.SetupSignalHandler()` for all the contexts.

go.mod
```diff
@@ -1,24 +1,24 @@
...
-go 1.13
+go 1.16
 
 require (
-	github.com/aws/amazon-vpc-cni-k8s v1.7.3
-	github.com/aws/aws-sdk-go v1.33.14
-	github.com/go-logr/logr v0.1.0
-	github.com/go-logr/zapr v0.1.0
+	github.com/aws/amazon-vpc-cni-k8s v1.9.0
+	github.com/aws/aws-sdk-go v1.40.6
```
abhipth, line 7: Do we need to switch this, given we are replacing this with internal?

go.mod
```diff
@@ -1,24 +1,24 @@
...
-	k8s.io/client-go v0.17.2
-	sigs.k8s.io/controller-runtime v0.5.0
+	github.com/prometheus/client_golang v1.11.0
+	github.com/stretchr/testify v1.7.0
+	go.uber.org/zap v1.18.1
+	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
+	k8s.io/client-go v0.21.3
+	sigs.k8s.io/controller-runtime v0.9.5
```
abhipth, line 21: nice!

pkg/k8s/pod/client_wrapper.go
```diff
@@ -186,7 +186,8 @@ func (p *podClientAPIWrapper) GetPod(namespace string, name string) (*v1.Pod, er
 // GetPodFromAPIServer returns the pod details by querying the API Server directly
 func (p *podClientAPIWrapper) GetPodFromAPIServer(namespace string, name string) (*v1.Pod, error) {
 	getPodFromAPIServeCallCount.Inc()
-	pod, err := p.coreV1.Pods(namespace).Get(name, metav1.GetOptions{
+	ctx := context.TODO()
```
abhipth, line 189: same, we need to use non empty context

webhooks/core/annotation_validation_webhook.go
```diff
@@ -46,9 +46,9 @@ func (a *AnnotationValidator) Handle(_ context.Context, req admission.Request) a
 		"request", req)
 
 	switch req.Operation {
-	case v1beta1.Create:
+	case admissionv1.Create:
```
abhipth, line 49: nice! along with the code we also need to change in the manifest files from v1beta1 to v1.

abhipth (changes requested):
Great work!

Could you also address the follwoing
- Use non empty context.
- Update the API versions for  `admissionregistration.k8s.io/v1beta1` in manifest as well.
- Update `apiextensions.k8s.io/v1beta1` as well.
 
See: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

controllers/core/node_controller_test.go
```diff
@@ -14,11 +14,12 @@
 package controllers
 
 import (
+	"context"
 	"testing"
 
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
+	mock_condition "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
```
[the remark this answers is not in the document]
sushrk: Yes.

internal/aws-sdk-go/go.mod
sushrk: This was auto-updated.
After removing internal/aws-sdk-go, this will be deleted. Will update when the other PR is merged.

go.mod
sushrk: Will be updated after merging the PR for removing internal/aws-sdk-go.

main.go
sushrk: < controller-runtime v0.7 had the value set to "configmaps", so was setting to the same. But verified that the new default "configmapsleases" also works.
Will update. 
Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L156

abhipth:
New PR was opened #60

PR #59, opened 2021-09-14
Remove aws-sdk-go from internal and use latest with ENI Trunking APIs 

abhipth:
*Description of changes:*
ENI Trunking API at the time of launch were not available in aws-sdk-go. Now aws-sdk-go has the API's available in SDK so removing the internal folder with custom aws-sdk-go. 

Note: Even though API are public, the feature is still private, which means the controller can still not run on Worker nodes.


README.md
```diff
@@ -10,7 +10,7 @@ This is a new open source project and we are actively working on enhancing the p
 
 ## Usage
 
-The ENI Trunking APIs are not yet publicly accessible. Attempting to run the controller on your worker node for enabling [Security Group for Pod](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) or use of aws-sdk-go from the internal directory for managing Trunk and Branch Network Interface will result in failure of the API calls. 
+The ENI Trunking APIs are not yet publicly accessible. Attempting to run the controller on your worker node for enabling [Security Group for Pod](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for managing Trunk and Branch Network Interface will result in failure of the API calls.
```
orsenthil, line 13: Should this README be updated?

From PR description, I read that The ENI Trunking APIs *are* publicly accessible with the update. 

With the caveat that, "Even though API are public, the feature is still private, which means the controller can still not run on Worker nodes."

abhipth: @orsenthil that makes sense, reworking to this, let me know if this looks good

ENI Trunking is a private feature even though the APIs are publicly accessible using AWS SDK. Hence, attempting to run the controller on your worker node for enabling [Security Group for Pod](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for managing Trunk and Branch Network Interface will result in failure of the API calls.

orsenthil: @abhipth - The update looks good.

In terms of README.md itself, you could consider removing the header section called `Usage` because README explains about this repo/controller in the first paragraph, and what is present in the usage section is a caveat.

orsenthil: Or moving the `Usage` section to the top, as appropriate.

abhipth: Moved Usage to top, removed some stale information from README as well.

orsenthil (approved):
LGTM.

PR #63, opened 2021-09-20
block IP address resource annotation from unauthorized users

abhipth:
*Description of changes:* Prevent Unauthorized users from Adding IPv4 Resource Annotation. Since the controller manages this resource, only the SA used by the controller is allowed to make the modifications.

The integration tests for this will be added once the ConfigMap feature for enabling Windows IPAM is added. Added a TODO for now.


M00nF1sh (approved):
/lgtm

PR #65, opened 2021-09-21
ConfigMap to enable/disable Windows IPAM

sushrk:
*Description of changes:*

1. Added controller to watch for ConfigMap changes
2. Handle Windows IPAM updates on ConfigMap changes
3. Added test cases

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -62,3 +62,22 @@ rules:
   - get
   - list
   - watch
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: controller-role
+  namespace: kube-system
+rules:
+- apiGroups:
+  - ""
+  resourceNames:
+  - amazon-vpc-cni
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -32,16 +32,13 @@ import (
 // ConfigMapReconciler reconciles a ConfigMap object
 type ConfigMapReconciler struct {
 	client.Client
-	Log         logr.Logger
-	Scheme      *runtime.Scheme
-	NodeManager manager.Manager
-	K8sAPI      k8s.K8sWrapper
-	Condition   condition.Conditions
+	Log                   logr.Logger
+	Scheme                *runtime.Scheme
+	NodeManager           manager.Manager
+	K8sAPI                k8s.K8sWrapper
+	Condition             condition.Conditions
+	curWinIPAMEnabledCond bool
 }
-
-var (
-	curWinIPAMEnabledCond = false
-)
 
 //+kubebuilder:rbac:groups=core,resources=configmaps,namespace=kube-system,resourceNames=amazon-vpc-cni,verbs=get;list;watch
 
@@ -70,9 +67,9 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Check if the flag value has changed
 	newWinIPAMEnabledCond := r.Condition.IsWindowsIPAMEnabled()
 
-	if curWinIPAMEnabledCond != newWinIPAMEnabledCond {
-		curWinIPAMEnabledCond = newWinIPAMEnabledCond
-		logger.Info("updated configmap", config.EnableWindowsIPAMKey, curWinIPAMEnabledCond)
+	if r.curWinIPAMEnabledCond != newWinIPAMEnabledCond {
+		r.curWinIPAMEnabledCond = newWinIPAMEnabledCond
+		logger.Info("updated configmap", config.EnableWindowsIPAMKey, r.curWinIPAMEnabledCond)
 
 		// Flag is updated, update all nodes
 		err := r.UpdateNodesOnConfigMapChanges()
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node/manager"
@@ -35,7 +36,12 @@ type ConfigMapReconciler struct {
 	Scheme      *runtime.Scheme
 	NodeManager manager.Manager
 	K8sAPI      k8s.K8sWrapper
+	Condition   condition.Conditions
 }
+
+var (
+	curWinIPAMEnabledCond = false
+)
 
 //+kubebuilder:rbac:groups=core,resources=configmaps,namespace=kube-system,resourceNames=amazon-vpc-cni,verbs=get;list;watch
 
@@ -46,7 +52,7 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	logger := r.Log.WithValues("configmap", req.NamespacedName)
 
 	// only update nodes on amazon-vpc-cni updates, return here for other updates
-	if req.Name != config.VpcCniConfigMapName {
+	if req.Name != config.VpcCniConfigMapName || req.Namespace != config.VpcCNIConfigMapNamespace {
 		return ctrl.Result{}, nil
 	}
 	configmap := &corev1.ConfigMap{}
@@ -61,18 +67,20 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	if val, ok := configmap.Data[config.EnableWindowsIPAMKey]; ok {
-		logger.Info("updated configmap", config.EnableWindowsIPAMKey, val)
-	}
+	// Check if the flag value has changed
+	newWinIPAMEnabledCond := r.Condition.IsWindowsIPAMEnabled()
 
-	// TODO: Only call UpdateNodesOnConfigMapChanges on enable-windows-ipam flag changes
+	if curWinIPAMEnabledCond != newWinIPAMEnabledCond {
+		curWinIPAMEnabledCond = newWinIPAMEnabledCond
+		logger.Info("updated configmap", config.EnableWindowsIPAMKey, curWinIPAMEnabledCond)
 
-	// Configmap is created/updated/deleted, update nodes
-	err := r.UpdateNodesOnConfigMapChanges()
-	if err != nil {
-		// Error in updating nodes
-		logger.Error(err, "Failed to update nodes on configmap changes")
-		return ctrl.Result{}, err
+		// Flag is updated, update all nodes
+		err := r.UpdateNodesOnConfigMapChanges()
+		if err != nil {
+			// Error in updating nodes
+			logger.Error(err, "Failed to update nodes on configmap changes")
+			return ctrl.Result{}, err
+		}
 	}
 
 	return ctrl.Result{}, nil
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -1,0 +1,107 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node/manager"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ConfigMapReconciler reconciles a ConfigMap object
+type ConfigMapReconciler struct {
+	client.Client
+	Log         logr.Logger
+	Scheme      *runtime.Scheme
+	NodeManager manager.Manager
+	K8sAPI      k8s.K8sWrapper
+}
+
+//+kubebuilder:rbac:groups=core,resources=configmaps,namespace=kube-system,resourceNames=amazon-vpc-cni,verbs=get;list;watch
+
+// Reconcile handles configmap create/update/delete events by invoking NodeManager
+// to update the status of the nodes as per the enable-windows-ipam flag value.
+
+func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("configmap", req.NamespacedName)
+
+	// only update nodes on amazon-vpc-cni updates, return here for other updates
+	if req.Name != config.VpcCniConfigMapName {
+		return ctrl.Result{}, nil
+	}
+	configmap := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, req.NamespacedName, configmap); err != nil {
+		if errors.IsNotFound(err) {
+			// If the configMap is deleted, de-register all the nodes
+			logger.Info("amazon-vpc-cni configMap is deleted")
+		} else {
+			// Error reading the object
+			logger.Error(err, "Failed to get configMap")
+			return ctrl.Result{}, err
+		}
+	}
+
+	if val, ok := configmap.Data[config.EnableWindowsIPAMKey]; ok {
+		logger.Info("updated configmap", config.EnableWindowsIPAMKey, val)
+	}
+
+	// TODO: Only call UpdateNodesOnConfigMapChanges on enable-windows-ipam flag changes
+
+	// Configmap is created/updated/deleted, update nodes
+	err := r.UpdateNodesOnConfigMapChanges()
+	if err != nil {
+		// Error in updating nodes
+		logger.Error(err, "Failed to update nodes on configmap changes")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}).
+		Complete(r)
+}
+
+func (r *ConfigMapReconciler) UpdateNodesOnConfigMapChanges() error {
+	nodeList, err := r.K8sAPI.ListNodes()
+	if err != nil {
+		return err
+	}
+	var errList []error
+	for _, node := range nodeList.Items {
+		_, found := r.NodeManager.GetNode(node.Name)
+		if found {
+			err = r.NodeManager.UpdateNode(node.Name)
+			if err != nil {
+				errList = append(errList, err)
+			}
+		}
+	}
+	if len(errList) > 0 {
+		return fmt.Errorf("failed to update one or more nodes %v", errList)
+	}
+	return nil
+}
--- a/controllers/core/configmap_controller_test.go
+++ b/controllers/core/configmap_controller_test.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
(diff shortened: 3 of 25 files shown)


pkg/node/manager/manager.go
```diff
@@ -386,3 +390,22 @@ func canAttachTrunk(node *v1.Node) bool {
 	_, ok := node.Labels[config.HasTrunkAttachedLabel]
 	return ok
 }
+
+// Update nodes on configmap changes
+func (m *manager) UpdateNodesOnConfigMapChanges() error {
+	nodeList, err := m.wrapper.K8sAPI.ListNodes()
```
abhipth, line 396: This API doesn't really fit for `NodeManager`. If you look at the other APIs they are all generic but this seems resource specific.

abhipth: It would be good to keep this in ConfigMap reconciler.

controllers/core/suite_test.go
```diff
@@ -64,6 +64,9 @@ var _ = BeforeSuite(func(done Done) {
 	err = corev1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = corev1.AddToScheme(scheme.Scheme)
```
abhipth, line 67: this seems duplicate of above two line of code

main.go
```diff
@@ -179,6 +182,17 @@ func main() {
 	renewDeadline := time.Second * time.Duration(leaderLeaseRenewDeadline)
 	retryPeriod := time.Second * time.Duration(leaderLeaseRetryPeriod)
 
+	// filter cache to subscribe to events from specific configmaps
+	newCache := cache.BuilderWithOptions(cache.Options{
+		SelectorsByObject: cache.SelectorsByObject{
+			&corev1.ConfigMap{}: {
+				Field: fields.AndSelectors(fields.SelectorFromSet(fields.Set{"metadata.name": "amazon-vpc-cni"}),
```
abhipth, line 189: need to go through this again tomorrow.

pkg/condition/conditions.go
```diff
@@ -14,17 +14,25 @@
...
 
 type condition struct {
 	hasDataStoreSynced *bool
 	log                logr.Logger
+	K8sAPI             k8s.K8sWrapper
 }
 
 const CheckDataStoreSyncedInterval = time.Second * 10
 
+var enableWindowsIPAMKey = "enable-windows-ipam"
```
abhipth, line 34: I think it's being reused at other places too, can we declare this as constant in config.go

pkg/condition/conditions.go
```diff
@@ -37,10 +45,34 @@ type Conditions interface {
 	IsPodSGPEnabled() bool
 }
 
-func NewControllerConditions(dataStoreSyncFlag *bool, log logr.Logger) Conditions {
+var (
+	conditionWindowsIPAMEnabled = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
```
abhipth, line 50: Why do we need a Summary metrics here? This represents a binary output 0/1 which should not require summary observations. 
Have you considered using https://prometheus.io/docs/concepts/metric_types/#gauge

pkg/condition/conditions.go
```diff
@@ -51,10 +83,27 @@ func (c *condition) WaitTillPodDataStoreSynced() {
...
+	if len(vpcCniConfigMap.Data) == 0 {
+		return false
+	}
+	if val, ok := vpcCniConfigMap.Data[enableWindowsIPAMKey]; ok {
+		enableWinIPAM, err := strconv.ParseBool(val)
+		if err != nil {
+			return false
+		}
+		if enableWinIPAM {
+			c.log.Info("Return IsWindowsIPAMEnabled = True")
```
abhipth, line 101: this is going to generate a lot of logging.

test/framework/manifest/configmap.go
```diff
@@ -0,0 +1,53 @@
...
+func (c *ConfigMapBuilder) Name(name string) *ConfigMapBuilder {
+	c.name = name
+	return c
+}
+func (c *ConfigMapBuilder) Namespace(namespace string) *ConfigMapBuilder {
+	c.namespace = namespace
+	return c
+}
+func (c *ConfigMapBuilder) Data(data map[string]string) *ConfigMapBuilder {
+	return c
```
abhipth, line 52: we are not setting the data here

abhipth (changes requested):
I have done a high level review right now. There are a couple of changes requested. Please let me know your thoughts on that.

I will go through the changes again tomorrow in details.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,83 @@
...
+		if errors.IsNotFound(err) {
+			logger.Info(req.Name, " ConfigMap is deleted")
+		} else {
+			// Error reading the object
+			logger.Error(err, "Failed to get ConfigMap ", req.Name)
+			return ctrl.Result{}, err
+		}
+	}
+
+	logger.Info("Enable-Windows-IPAM=", configmap.Data["enable-windows-ipam"])
```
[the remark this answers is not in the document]
sushrk: Yes, nice catch. 
Fixed!

pkg/condition/conditions.go
sushrk: Noted. 
Removing the log here, changes to the flag will be logged in the Reconcile function.

test/framework/manifest/configmap.go
sushrk: Fixed.

main.go
abhipth: Wondering how does this work?

```
// AndSelectors creates a selector that is the logical AND of all the given selectors
```

So does this mean we cache, ConfigMap with Name `amazon-vpc-cni` AND name `config.LeaderElectionKey` AND namesapce `kube-system`? This would result in nothing being watched I believe. Has this been tested?

abhipth: What we are looking for is OR or IN. But I don't think that's possible. [See](https://github.com/kubernetes/kubernetes/issues/32946)

pkg/condition/conditions.go
```diff
@@ -51,10 +83,27 @@ func (c *condition) WaitTillPodDataStoreSynced() {
 	}
 }
 
-// TODO: Add implementation later
 func (c *condition) IsWindowsIPAMEnabled() bool {
-	// TODO: Switch using ConfigMap
-	return true
+	// Return false if configmap not present/any errors
+	vpcCniConfigMap, err := c.K8sAPI.GetConfigMap("amazon-vpc-cni")
```
abhipth, line 88: We can read this from a constant.

pkg/node/manager/manager.go
```diff
@@ -338,7 +342,7 @@ func (m *manager) isSelectedForManagement(v1node *v1.Node) bool {
 		return false
 	}
 
-	return isWindowsNode(v1node) || canAttachTrunk(v1node)
+	return m.conditions.IsWindowsIPAMEnabled() || canAttachTrunk(v1node)
```
abhipth, line 345: shouldn't this be `isWindowsNode(v1node) && m.conditions.IsWindowsIPAMEnabled()`?

test/framework/resource/k8s/configmap/wrapper.go
```diff
@@ -0,0 +1,49 @@
...
+func UpdateConfigMapAndWait(k8sClient client.Client, ctx context.Context, configmap *v1.ConfigMap, data map[string]string) {
+	By("updating the configmap")
+	updatedConfigmap := &v1.ConfigMap{}
+	err := k8sClient.Get(ctx, client.ObjectKey{
+		Namespace: configmap.Namespace,
+		Name:      configmap.Name,
+	}, updatedConfigmap)
+	Expect(err).NotTo(HaveOccurred())
+
+	updatedConfigmap.Data = data
```
abhipth, line 46: We need to make a DeepCopy before modifying the object.

test/integration/windows/windows_test.go
```diff
@@ -48,22 +49,32 @@ var _ = Describe("Windows Integration Test", func() {
...
+		data                 map[string]string
+		enableWindowsIPAMKey string
 	)
 
 	BeforeEach(func() {
 		namespace = "windows-test"
 		jobParallelism = 1
 		podLabelKey = "role"
 		podLabelVal = "integration-test"
+		enableWindowsIPAMKey = "enable-windows-ipam"
```
abhipth, line 64: this can be read from constants.

test/integration/windows/windows_test.go
```diff
@@ -74,6 +85,76 @@ var _ = Describe("Windows Integration Test", func() {
...
 		frameWork.NSManager.DeleteAndWaitTillNamespaceDeleted(ctx, namespace)
+		configMapWrapper.DeleteConfigMap(frameWork.K8sClient, ctx, configmap)
+	})
+
+	Describe("windows IPAM tests", func() {
+		var deployment *appsV1.Deployment
+
+		BeforeEach(func() {
+
+			deployment = manifest.NewWindowsDeploymentBuilder().
```
abhipth, line 96: We already have these sort of tests, can't we reuse those instead of a new one?

abhipth: We could test with a single Pod instead of deployment as we are just testing the feature flag.

- flag=false - flag=true, Pod create successfully.
- flag=false, Pod fail to create.
- No ConfigMap, Pod fail to create
- ConfigMap with No Data field, Pod fail to crate
- ConfigMap with Data Field but wrong value, Pod fail to create.

We also need to verify the Pod doesn't have the ResourceLimit injected when the flag is not true.

test/framework/verify/pod.go
```diff
@@ -146,10 +146,14 @@ func (v *PodVerification) ExpectPodHaveDesiredPhase(namespace string,
...
-func (v *PodVerification) WindowsPodHaveExpectedIPv4Address(pod *v1.Pod) {
+func (v *PodVerification) WindowsPodHaveIPv4Address(pod *v1.Pod, expected bool) {
 	By("matching the IPv4 from annotation to the pod IP")
 	ipAddWithCidr, found := pod.Annotations["vpc.amazonaws.com/PrivateIPv4Address"]
-	Expect(found).To(BeTrue())
-	// Remove the CIDR and compare pod IP with the IP annotated from VPC Controller
-	Expect(pod.Status.PodIP).To(Equal(strings.Split(ipAddWithCidr, "/")[0]))
+	Expect(found).To(Equal(expected))
+
+	if expected {
```
abhipth, line 154: This should no longer be required once we switch to testing Pod instead of Deployment fail to create when the feature is disabled.

config/rbac/role.yaml
```diff
@@ -23,6 +23,21 @@ rules:
...
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
```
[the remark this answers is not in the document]
sushrk: removed.

pkg/node/manager/manager.go
sushrk: Good point, I have moved this to the ConfigMap controller.

webhooks/core/pod_webhook_test.go
```diff
@@ -45,7 +46,8 @@ var (
 )
 
 type Mock struct {
-	SGPMock *mock_utils.MockSecurityGroupForPodsAPI
+	SGPMock        *mock_utils.MockSecurityGroupForPodsAPI
+	ConditionsMock *mock_condition.MockConditions
```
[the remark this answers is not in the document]
sushrk: Yes, will add in the next commit.

test/integration/windows/windows_test.go
sushrk: Added the new test cases, please have a look.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+// ConfigMapReconciler reconciles a ConfigMap object
+type ConfigMapReconciler struct {
+	client.Client
+	Log         logr.Logger
+	Scheme      *runtime.Scheme
+	NodeManager manager.Manager
+	K8sAPI      k8s.K8sWrapper
+}
+
+//+kubebuilder:rbac:groups=core,resources=configmaps,resourceNames=amazon-vpc-cni,verbs=get;list;watch
```
abhipth, line 40: We need to further restict it with `kube-system` namespace?

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+// Reconcile handles configmap create/update/delete events by invoking NodeManager
+// to update the status of the nodes as per the enable-windows-ipam flag value.
+
+func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("configmap", req.NamespacedName)
+
+	// only update nodes on amazon-vpc-cni updates, return here for other updates
+	if req.Name != config.VpcCniConfigMapName {
```
abhipth, line 49: One way this condition can be exploited is by creating/updating the same name ConfigMap but in different namespace?
maybe having the additional namespace Guard will make this condition more concrete.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+		return ctrl.Result{}, nil
+	}
+	configmap := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, req.NamespacedName, configmap); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("amazon-vpc-cni configMap is deleted")
```
abhipth, line 55: nit: we can add one comment stating, in case map is deleted we need to de register all nodes. This can come later though.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+		} else {
+			// Error reading the object
+			logger.Error(err, "Failed to get configMap")
+			return ctrl.Result{}, err
+		}
+	}
+
+	if val, ok := configmap.Data[config.EnableWindowsIPAMKey]; ok {
+		logger.Info("ConfigMap updated, enable-windows-ipam=", "Value", val)
```
abhipth, line 64: `logger.Info("configmap updated", "enable-windows-ipam", val)`

Please replace `enable-windows-ipam` everywhere with a constant.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}).
+		Complete(r)
+}
+
+func (r *ConfigMapReconciler) UpdateNodesOnConfigMapChanges(nodeManager manager.Manager) error {
```
abhipth, line 85: You can skip passing `nodeManager` as argument because you can use `r.NodeManager` inside the function body.

controllers/core/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	if val, ok := configmap.Data[config.EnableWindowsIPAMKey]; ok {
+		logger.Info("ConfigMap updated, enable-windows-ipam=", "Value", val)
+	}
+
+	// Configmap is created/updated/deleted, update nodes
+	err := r.UpdateNodesOnConfigMapChanges(r.NodeManager)
```
abhipth, line 68: We should only call this function when the `enable_windows_ipam` value changes. We will keep on updating node in future when we add additional values that are not relevant for Windows IPAM.

I am fine with
- Fixing it now (if it's a quick fix)
- Adding a TODO and fixing it in the follow up release.

pkg/condition/conditions.go
```diff
@@ -37,10 +44,35 @@ type Conditions interface {
 	IsPodSGPEnabled() bool
 }
 
-func NewControllerConditions(dataStoreSyncFlag *bool, log logr.Logger) Conditions {
+var (
+	conditionWindowsIPAMEnabled = prometheus.NewGaugeVec(
```
abhipth, line 48: we don't need a vector here since this is not multi dimensional. `prometheus.NewGauge` should be fine.

pkg/node/manager/manager.go
```diff
@@ -172,6 +175,7 @@ func (m *manager) UpdateNode(nodeName string) error {
 	cachedNode, found := m.dataStore[nodeName]
 	if !found {
 		m.Log.Info("the node doesn't exist in cache anymore, it might have been deleted")
+		return nil
```
abhipth, line 178: Good catch!

pkg/node/manager/manager.go
abhipth: @sushrk you may want to address this as well. Thanks!

test/integration/windows/windows_test.go
```diff
@@ -76,6 +83,74 @@ var _ = Describe("Windows Integration Test", func() {
...
 
+	Describe("configMap enable-windows-ipam tests", func() {
+		var testPod *v1.Pod
+		var createdPod *v1.Pod
+		BeforeEach(func() {
+			testPod, err = manifest.NewWindowsPodBuilder().Container(testerContainer).Build()
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Context("when configmap created", func() {
```
abhipth, line 94: nice test structure!!

test/integration/windows/windows_test.go
```diff
@@ -76,6 +83,74 @@ var _ = Describe("Windows Integration Test", func() {
...
+			})
+			JustAfterEach(func() {
+				// restore the configmap after each test
+				data = map[string]string{config.EnableWindowsIPAMKey: "true"}
+				configMapWrapper.UpdateConfigMap(frameWork.K8sClient, ctx, configMap, data)
+			})
+
+			Context("when enable-windows-ipam is True", func() {
+				It("pod should be running and have resourceLimit injected", func() {
+					createdPod, err = frameWork.PodManager.CreateAndWaitTillPodIsCompleted(ctx, testPod)
```
abhipth, line 108: I hope the container is not a server, if it is then this condition will never return. Maybe you wanted to use CreateAndWaitTillPodIsRunning here?

abhipth (commented):
Thanks! the new changes look good overall. Have few minor comments.

main.go
sushrk: I've changed this to watch only for the amazon-vpc-cni configmap. 
Since we are no longer watching the leader election configmap, using the fieldSelector to match name and namespace of a single configmap should be good.

controllers/core/configmap_controller.go
sushrk: You're right. I will add this.

main.go
```diff
@@ -179,6 +182,17 @@ func main() {
 	renewDeadline := time.Second * time.Duration(leaderLeaseRenewDeadline)
 	retryPeriod := time.Second * time.Duration(leaderLeaseRetryPeriod)
 
+	// filter cache to subscribe to events from specific configmaps
+	newCache := cache.BuilderWithOptions(cache.Options{
+		SelectorsByObject: cache.SelectorsByObject{
+			&corev1.ConfigMap{}: {Field: fields.Set{
+				"metadata.name":      config.VpcCniConfigMapName,
+				"metadata.namespace": config.KubeSystemNamespace,
```
M00nF1sh, line 190: 1. why don't using cache.Options.Namespace?
2. let's use separate variable for LeaderElectionNamespace and maybe `VpcCNIConfigMapNamespace` instead of `KubeSystemNamespace` since technically they might be different in the future.

sushrk: cache.Options.Namespace will restrict the cache for watching all other resources. 
SelectorsByObject will restrict it per GVK as we only want to filter cache for the ConfigMap. 

I've changed the namespace variables.

pkg/condition/conditions.go
sushrk: Changed it to use gauge.

controllers/core/configmap_controller_test.go
```diff
@@ -0,0 +1,109 @@
...
+			Client:      client,
+			Log:         zap.New(),
+			NodeManager: mockNodeManager,
+		},
+		MockNode:       mockNode,
+		MockConditions: mockConditions,
+	}
+}
+
+func Test_Reconcile_UpdateNode(t *testing.T) {
```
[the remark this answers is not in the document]
sushrk: Added!

pkg/node/manager/manager.go
sushrk: Fixed.

pkg/condition/conditions.go
```diff
@@ -51,10 +82,22 @@ func (c *condition) WaitTillPodDataStoreSynced() {
...
 }
 
-// TODO: Add implementation later
 func (c *condition) IsWindowsIPAMEnabled() bool {
-	// TODO: Switch using ConfigMap
-	return true
+	// Return false if configmap not present/any errors
+	vpcCniConfigMap, err := c.K8sAPI.GetConfigMap(config.VpcCniConfigMapName)
+
+	if err == nil && len(vpcCniConfigMap.Data) > 0 {
```
abhipth, line 89: shouldn't this be `vpcCniConfigMap.Data != nil`? We will need UTs to test all of these conditions

pkg/k8s/wrapper.go
```diff
@@ -146,3 +148,17 @@ func (k *k8sWrapper) AdvertiseCapacityIfNotSet(nodeName string, resourceName str
...
+	configMap := &v1.ConfigMap{}
+	err := k.cacheClient.Get(context.Background(), types.NamespacedName{
+		Name: configMapName,
+	}, configMap)
+	return configMap, err
+}
+
+func (k *k8sWrapper) ListNodes() (*v1.NodeList, error) {
+	nodeList := &v1.NodeList{}
+	err := k.cacheClient.List(context.Background(), nodeList)
```
abhipth, line 162: We should use the non empty context here.

test/framework/verify/pod.go
```diff
@@ -146,10 +146,20 @@ func (v *PodVerification) ExpectPodHaveDesiredPhase(namespace string,
...
 	By("matching the IPv4 from annotation to the pod IP")
 	ipAddWithCidr, found := pod.Annotations["vpc.amazonaws.com/PrivateIPv4Address"]
 	Expect(found).To(BeTrue())
 	// Remove the CIDR and compare pod IP with the IP annotated from VPC Controller
 	Expect(pod.Status.PodIP).To(Equal(strings.Split(ipAddWithCidr, "/")[0]))
+
+}
+
+func (v *PodVerification) WindowsPodHaveResourceLimits(pod *v1.Pod, expected bool) {
+	_, found := pod.Spec.Containers[0].Resources.Limits[config.ResourceNameIPAddress]
```
abhipth, line 159: as discussed offline integration tests are failing here, need to check if `Spec.Containers[0].Resources.Limits != nil`

test/integration/windows/windows_suite_test.go
```diff
@@ -51,6 +54,10 @@ var _ = BeforeSuite(func() {
 	ctx = context.Background()
 	verify = verifier.NewPodVerification(frameWork, ctx)
 
+	By("creating the configmap with enable-windows-ipam set to true")
+	configMap = manifest.NewConfigMapBuilder().Build()
+	configMapWrapper.CreateConfigMap(frameWork.K8sClient, ctx, configMap)
```
abhipth, line 59: If the ConfigMap is already there will this fail? I think we can do 
- Get
  - If present, enable the feature
  - Create otherwise

test/integration/windows/windows_suite_test.go
```diff
@@ -72,6 +79,9 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	By("deleting the configmap")
+	configMapWrapper.DeleteConfigMap(frameWork.K8sClient, ctx, configMap)
```
abhipth, line 83: can we revert back to the older state if ConfigMap already existed (as in future it might contain CNI related variables)

abhipth (dismissed):
Changes look good! Thanks for the effort you put in. 
It would be good to have UTs for IsWindowsIPAMEnabled() but that can come as follow up PR too.

test/integration/windows/windows_suite_test.go
sushrk: Will resolve this in a follow up PR.

sushrk: Resolved in the recent commit to fix integration test errors.

test/integration/windows/windows_suite_test.go
sushrk: Resolved in the last commit.

abhipth (approved):

PR #68, opened 2021-09-27
Block Windows IPAM till Old Controller Deployment is deleted

abhipth:
*Description of changes:*
Disable the new controller if the old Controller deployment is still active or becomes active afterwards. This will help prevent two controllers running together at the same which could lead to disruption of running Pods. 
 
### DIFF
--- a/pkg/k8s/wrapper.go
+++ b/pkg/k8s/wrapper.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/prometheus/client_golang/prometheus"
 
+	appV1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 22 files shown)


M00nF1sh (commented):
This works for me, however I feel we did too much to sync with old windows VPC resource controller change. (since i imaging in 99% case, once user disabled the old windows VPC resource controller, they won't reinstall it again)

The best user experience i'd expect is
1. have a mutating webhook for our vpc resource controller's configmap.
2. If the configMap's value of enable windows support is changed to true, we do a get(without cache) to check for  old windows vpc resource controller's existence, and if it exits, we reject the configmap change with clear error message(e.g. you need to uninstall windows vpc resource controller first)
3. let the vpc resource controller solely rely on the configmap's value.

This way 
* it have clear customer experience(they get instant feedback when they mutates configs), and can be extended to other config values.
* it reduces the rbac permission/memory consumption that used to list deployments.

what do you think

pkg/k8s/wrapper.go
```diff
@@ -94,6 +96,15 @@ func NewK8sWrapper(client client.Client, coreV1 corev1.CoreV1Interface, ctx cont
 	return &k8sWrapper{cacheClient: client, eventRecorder: recorder, context: ctx}
 }
 
+func (k *k8sWrapper) GetDeployment(namespace string, name string) (*appV1.Deployment, error) {
+	deployment := &appV1.Deployment{}
+	err := k.cacheClient.Get(k.context, types.NamespacedName{
```
M00nF1sh, line 101: it's a bad practice to store context objects. let's refactor this in future versions given it's already like this in other apis.

M00nF1sh (approved):
accepting this as a temporary solution.
we should evaluate to create this configmap with certain labels from the controller itself if not present.

/lgtm

PR #72, opened 2021-10-04
change log level to debug level

abhipth:
*Description of changes:* The following log line is printed on controller restarts for all Pods when the nodes are being initialized making the logs very noisy hence changing to debug level.


M00nF1sh (approved):
/lgtm

PR #84, opened 2021-11-22
modify go proxy link when testing in China Regions

abhipth:
*Issue #, if available:*
The Canary script fails for China Region because the default GOPROXY is banned in China. Switching to https://github.com/goproxy/goproxy.cn for testing in China Region.


orsenthil (approved):
LGTM.

Thank you!

PR #85, opened 2021-12-03
increase timeouts/retries for canary tests

abhipth:
*Issue #, if available:*
Was seeing the following mode of failures in the Canary tests over multiple runs.

1. Rate Limiting Exceeded
   ```
    An error occurred (Throttling) when calling the AttachRolePolicy operation (reached max retries: 2): Rate exceeded
    ```
    **Cause:** If tests for all K8s version are triggered at same time we may have this issue.

    **Resolution:**  Increase the max retries and added jitter in the script that invokes the canary script.

2. Windows Pod timeout because of docker image pull
   ```
   Windows Integration Test
   /repository/vpc-resource-controller/test/integration/windows/windows_test.go:33
   configMap enable-windows-ipam tests
   /repository/vpc-resource-controller/test/integration/windows/windows_test.go:87
   when configmap created
   /repository/vpc-resource-controller/test/integration/windows/windows_test.go:97
   [CANARY] when enable-windows-ipam is True
   /repository/vpc-resource-controller/test/integration/windows/windows_test.go:113
   pod should be running and have resourceLimit injected [It]
   /repository/vpc-resource-controller/test/integration/windows/windows_test.go:114
  
   Unexpected error:
   <*errors.errorString | 0xc000365890>: {
   s: \"timed out waiting for the condition\",
   }
   timed out waiting for the condition
   occurred
   ```
    **Cause:** Windows Pod take more time to pull container image hence have higher pod startup latency.
    From audit logs saw, Create Event - 2021-11-29T08:37:59.586-08:00 and Update/Delete Event - 2021-11-29T08:40:10.398-08:00 where the Container was not started. This test fails only when it's executed the first time when the docker images are not pulled.

    **Resolution:**  Increase the wait duration for the Pod to become ready. 

3. Service test failing intermittently 
    ```
     Unexpected error:
     <*errors.errorString | 0xc000547ff0>: {
     s: \"failed to execute job :{[] 2021-11-18 16:42:47 +0000 UTC <nil> 30 0 1 }\",
     }
     failed to execute job :{[] 2021-11-18 16:42:47 +0000 UTC <nil> 30 0 1 }
     occurred
    ```
   **Possible Cause:**: Suspecting the service doesn't become ready.
    **Resolution:**: Increasing the timeout further.

*Description of changes:*

### DIFF
--- a/scripts/test/run-canary-test.sh
+++ b/scripts/test/run-canary-test.sh
@@ -116,6 +116,9 @@ function set_env_aws_node() {
 }
 
 function run_canary_tests() {
+  # For each component, we want to cover the most important test cases. We also don't want to take more than 30 minutes
+  # per repository as these tests are run sequentially along with tests from other repositories
+  # Currently the overall execution time is ~50 minutes and we will reduce it in future
   (cd $INTEGRATION_TEST_DIR/perpodsg && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 15m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
   (cd $INTEGRATION_TEST_DIR/windows && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 27m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
   (cd $INTEGRATION_TEST_DIR/webhook && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 5m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
--- a/scripts/test/run-canary-test.sh
+++ b/scripts/test/run-canary-test.sh
@@ -46,9 +46,11 @@ function load_cluster_details() {
   echo "VPC ID: $VPC_ID, Service Role ARN: $SERVICE_ROLE_ARN, Role Name: $ROLE_NAME"
 }
 
+# This operation fails with rate limit exceeded when test is running for multiple K8s
+# version at same time, hence we increase the exponential retries on the aws call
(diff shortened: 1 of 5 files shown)


scripts/test/run-canary-test.sh
```diff
@@ -115,7 +117,7 @@ function set_env_aws_node() {
 
 function run_canary_tests() {
   (cd $INTEGRATION_TEST_DIR/perpodsg && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 15m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
-  (cd $INTEGRATION_TEST_DIR/windows && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 25m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
+  (cd $INTEGRATION_TEST_DIR/windows && CGO_ENABLED=0 ginkgo --focus="CANARY" -v -timeout 27m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID)
```
achevuru, line 120: Curious...How did we arrive at these numbers? I do see that we already have hard coded/specific timeout values for these tests but a comment on how we converged at these numbers will help. Even better if they are configurable but that can be a separate enhancement.

test/framework/utils/poll.go
```diff
@@ -22,4 +22,7 @@ const (
 	// ResourceCreationTimeout is the number of seconds till the controller waits
 	// for the resource creation to complete
 	ResourceCreationTimeout = 120 * time.Second
+	// Windows Container Images are much larger in size and pulling them the first
```
achevuru, line 25: Nice. Thanks for adding the reasoning behind the value chosen for timeout.

scripts/test/run-canary-test.sh
kishorj: If 25m timeout did not work, 27m would likely not work for the same reason. Please explain why 27m is a better choice here.

kishorj (dismissed):

kishorj (approved):

PR #62, opened 2021-09-18
Add Introspect API for debugging resources across all managed nodes

abhipth:
*Issue #, if available:* #34 

*Description of changes:*
Add introspect API for debugging issues with all Resources across all managed Nodes. This API can later be exposed for customers if required.

Example output.

**Introspect a Node**
```
curl 192.168.6.143:22775/node/ip-192-168-28-99.us-west-2.compute.internal
{
	"vpc.amazonaws.com/PrivateIPv4Address": {
		"UsedResources": {
			"019e0387-df0a-4ec1-a449-76e73a3219c6": "192.168.10.187/19",
			"5a1008ce-7d5e-45fa-8d8c-16b8c9b69b24": "192.168.26.221/19"
		},
		"WarmResources": [
			"192.168.23.206/19",
			"192.168.9.91/19",
			"192.168.30.13/19"
		],
		"CoolingResources": null
	},
	"vpc.amazonaws.com/pod-eni": {}
```

```
curl 192.168.6.143:22775/node/ip-192-168-74-131.us-west-2.compute.internal
{
	"vpc.amazonaws.com/PrivateIPv4Address": {},
	"vpc.amazonaws.com/pod-eni": {
		"TrunkENIID": "eni-0b40ea0edb1313415",
		"InstanceID": "i-03475ffa4ba099884",
		"PodToBranchENI": {
			"7b3fa2e4-e073-41b9-99fd-2f9ed655d0a9": [
				{
					"eniId": "eni-0b5c8ff07fe5e6f7b",
					"ifAddress": "0e:a1:a3:43:70:6b",
					"privateIp": "192.168.68.74",
					"vlanId": 2,
					"subnetCidr": "192.168.64.0/19"
				}
			],
			"81fe2fc0-f805-4e01-876b-a18b44d7472c": [
				{
					"eniId": "eni-0efc320df38a31be0",
					"ifAddress": "0e:09:7b:3b:f3:e5",
					"privateIp": "192.168.75.199",
					"vlanId": 3,
					"subnetCidr": "192.168.64.0/19"
				}
			],
			"92709dc6-796b-4654-9695-9199ecd54ec3": [
				{
					"eniId": "eni-0bb3d659f87398dd2",
					"ifAddress": "0e:53:35:d2:d6:91",
					"privateIp": "192.168.78.19",
					"vlanId": 1,
					"subnetCidr": "192.168.64.0/19"
				}
			]
		},
		"DeleteQueue": null
	}
}
```

**Introspect All Resource**
```
[...]

PR #90, opened 2022-01-07
Refactoring Describe ENI API calls to paginated version

haouc:
*Issue #, if available:*
Unpaginated API call DescribeNetworkInterface could put throttling on EC2. We should use DescribeNetworkInterfacePages if not specifying interface ID as parameter.

*Description of changes:*
Refactoring the API call to paginated version.

*Tests:*

1. e2e tests focus on security group for pods with the setting 4 parallelism/3 nodes as stress (Passed 
```
perpodsg % CGO_ENABLED=0 GOOS=$OS ginkgo -v -timeout 40m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id $VPC_ID --ginkgo.skip=LOCAL
```
```
Ran 18 of 22 Specs in 1553.904 seconds
SUCCESS! -- 18 Passed | 0 Failed | 0 Pending | 4 Skipped
PASS

Ginkgo ran 1 suite in 25m59.650309764s
Test Suite Passed
```
2. performance/soak tests with 100 nodes and 0 - 1000 pods (SGPP) every 15 minutes for 10 hours (No leaked ENI)

### DIFF
--- a/main.go
+++ b/main.go
@@ -123,7 +123,7 @@ func main() {
 	flag.StringVar(&introspectBindAddr, "introspect-bind-addr", ":22775",
 		"Port for serving the introspection API")
 	flag.IntVar(&describeENIPageSize, "describe-eni-page-size", 100,
-		"The page size limiting the number of every page returned from paginated EC2 describe ENI API calls")
+		"The page size limiting the number of items included in the output of EC2 DescribeNetworkInterface per page")
 
 	flag.Parse()
 
--- a/main.go
+++ b/main.go
@@ -87,6 +87,7 @@ func main() {
 	var logLevel string
 	var clusterName string
 	var listPageLimit int
+	var describeENIPageSize int
 	var leaderLeaseDurationSeconds int
 	var leaderLeaseRenewDeadline int
 	var leaderLeaseRetryPeriod int
@@ -121,6 +122,8 @@ func main() {
 	flag.StringVar(&outputPath, "log-file", "stderr", "The path to redirect controller logs")
 	flag.StringVar(&introspectBindAddr, "introspect-bind-addr", ":22775",
 		"Port for serving the introspection API")
+	flag.IntVar(&describeENIPageSize, "describe-eni-page-size", 100,
+		"The page size limiting the number of every page returned from paginated EC2 describe ENI API calls")
 
 	flag.Parse()
 
@@ -240,7 +243,8 @@ func main() {
 	if err != nil {
 		setupLog.Error(err, "unable to create ec2 wrapper")
 	}
-	ec2APIHelper := ec2API.NewEC2APIHelper(ec2Wrapper, clusterName)
+	setupLog.V(1).Info("setting up page size for describe interfaces API calls", "max page size", describeENIPageSize)
+	ec2APIHelper := ec2API.NewEC2APIHelper(ec2Wrapper, clusterName, describeENIPageSize)
 
 	sgpAPI := utils.NewSecurityGroupForPodsAPI(
 		mgr.GetClient(),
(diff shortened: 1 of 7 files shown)


orsenthil (approved):
LGTM, @haouc . Thank you.

pkg/aws/ec2/api/eni_cleanup_test.go
```diff
@@ -85,11 +85,21 @@ func TestENICleaner_cleanUpAvailableENIs(t *testing.T) {
 
 	gomock.InOrder(
 		// Return network interface 1 and 2 in first cycle
-		mockWrapper.EXPECT().DescribeNetworkInterfaces(mockDescribeNetworkInterfaceIp).
-			Return(mockDescribeInterfaceOpWith1And2, nil),
+		mockWrapper.EXPECT().DescribeNetworkInterfacesPages(gomock.Any(), gomock.Any()).
```
abhipth, line 88: Can we use the expected sample input here instead of any input?

abhipth (commented):
Just want to confirm, if we have verified the following with the changes as there are no integration tests for the same. 

If we set the max page size to let's say 20 and
- If we have more than 20 leaked ENIs are they cleaned in the ENI cleanup cycle?
- If we have more than 20 Branch ENIs do they show up in the response of the [Introspect API](https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/62) - ?

pkg/aws/ec2/api/eni_cleanup_test.go
haouc: I think we should be fine to use Any() in this case because we invoke DoAndReturn() which declares the action and will execute the mocked function and the args will be checked. Please let me know if you think otherwise. Thanks.

main.go
```diff
@@ -121,6 +122,8 @@ func main() {
 	flag.StringVar(&outputPath, "log-file", "stderr", "The path to redirect controller logs")
 	flag.StringVar(&introspectBindAddr, "introspect-bind-addr", ":22775",
 		"Port for serving the introspection API")
+	flag.IntVar(&describeENIPageSize, "describe-eni-page-size", 100,
+		"The page size limiting the number of every page returned from paginated EC2 describe ENI API calls")
```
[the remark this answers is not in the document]
haouc: Yeah, makes sense. Will update.

haouc:
> Just want to confirm, if we have verified the following with the changes as there are no integration tests for the same.
> 
> If we set the max page size to let's say 20 and
> 
>     * If we have more than 20 leaked ENIs are they cleaned in the ENI cleanup cycle?
> 
>     * If we have more than 20 Branch ENIs do they show up in the response of the [Introspect API](https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/62) - ?
Created 30 leaked ENI were cleaned up through two cycles by the new paginated API calls. Branch ENIs has been confirmed shown up in introspection API's response.
Example log for clean up (scanning interval was overridden to 10 minutes for testing) (all ENIs were confirmed removed by checking console list)
```
{"level":"debug","timestamp":"2022-01-14T19:15:20.008Z","logger":"eni cleaner","msg":"Paginated Describe ENI call to return pages","current page size":1}
{"level":"debug","timestamp":"2022-01-14T19:15:20.139Z","logger":"eni cleaner","msg":"Paginated Describe ENI call to return pages","current page size":2}
...
{"level":"debug","timestamp":"2022-01-14T19:25:36.519Z","logger":"eni cleaner","msg":"Paginated Describe ENI call to return pages","current page size":2}
{"level":"debug","timestamp":"2022-01-14T19:25:37.401Z","logger":"eni cleaner","msg":"Paginated Describe ENI call to return pages","current page size":2}
```

abhipth (approved):

PR #102, opened 2022-04-13
do limited retries when checking addon status

abhipth:
*Issue #, if available:* Canary tests are stuck in infitinite loop, if vpc-cni add on doesn't become ready. With this change we do limited retries.

*Description of changes:*

### DIFF
--- a/scripts/test/run-canary-test.sh
+++ b/scripts/test/run-canary-test.sh
@@ -64,25 +64,40 @@ function detach_controller_policy_cluster_role() {
 
 function wait_for_addon_status() {
   local expected_status=$1
-
+  local retry_attempt=0
   if [ "$expected_status" =  "DELETED" ]; then
-    while $(aws eks describe-addon $ENDPOINT_FLAG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME --region $REGION); do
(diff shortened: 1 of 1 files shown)


jayanthvn (approved):

scripts/test/run-canary-test.sh
```diff
@@ -64,25 +64,40 @@ function detach_controller_policy_cluster_role() {
 
 function wait_for_addon_status() {
   local expected_status=$1
-
+  local retry_attempt=0
   if [ "$expected_status" =  "DELETED" ]; then
-    while $(aws eks describe-addon $ENDPOINT_FLAG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME --region $REGION); do
+    while $(aws eks describe-addon $ENDPOINT_FLAG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME --region "$REGION"); do
+      if [ $retry_attempt -ge 30 ]; then
```
achevuru, line 70: I think `5` retries spaced out by `30s` each will give us the same result and we will end up reducing no.of `describe-addon` calls. Sure, in the worst case we might wait an additional `25s` before we move on but that might be a minor penalty. Might be worth considering. 

But, did we see this behavior (i.e.,) VPC CNI stuck in `Terminating` state when we debugged a stuck canary job?

Issue #99, opened 2022-03-08
Reconcile Security Groups for Pods


ellistarn:
**Describe the Bug**:

When using Karpenter and SGP, pods are stuck in ContainerCreating for up to 30 minutes. https://github.com/aws/karpenter/issues/1252

**Observed Behavior**:
Security groups are only attached for pods that are bound after a Node is `Ready`. For pods that define `pod.spec.nodeName` or are bound by optimistic schedulers (e.g. Karpenter), the pod will remain stuck in Container Creating

**Expected Behavior**:
The VPC Resource Controller should trigger a reconciliation loop for all pods on a node when the node becomes ready. 

**How to reproduce it (as minimally and precisely as possible)**:
- https://karpenter.sh/v0.6.4/getting-started/
- Use the security group for pods feature
- Observe ContainerCreating

**Additional Context**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- OS (Linux/Windows):

haouc:
@ellistarn seems this is no longer an issue regarding the change in https://github.com/aws/karpenter/issues/1252 ? I am closing this issue for now. If this still need a fix in this controller, please feel free to reopen. Thanks

ellistarn:
We ended up rearchitecting in a way that stopped this from happening. Still probably a useful resiliency feature for SGPP, but that's a judgement call for you folks.

PR #103, opened 2022-04-15
Reconcile pods on node events

dewjam:
*Issue #, if available:*
#99 
[Karpenter Issue](https://github.com/aws/karpenter/issues/1252)

*Description of changes:*
This change gives the Custom Pod Controller the capability to watch for events from other resource types.  In this case, we add a Watch for v1.Node events, and then reconcile all pods on the given node as a result.  The pod spec is retrieved from the "optimized" cache.  

*This change is largely influenced by how the default Controller implementation handles Watches.*

### DIFF
--- a/controllers/core/pod_controller.go
+++ b/controllers/core/pod_controller.go
@@ -24,9 +24,14 @@ import (
 
 	"github.com/go-logr/logr"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;update;patch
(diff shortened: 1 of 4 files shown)


controllers/core/pod_controller.go
```diff
@@ -165,5 +170,17 @@ func (r *PodReconciler) SetupWithManager(ctx context.Context, manager ctrl.Manag
 			PageLimit:               pageLimit,
 			ResyncPeriod:            syncPeriod,
 			MaxConcurrentReconciles: 2,
-		}).Complete(r)
+		}).Watches(&source.Kind{Type: &v1.Node{}},
```
dewjam, line 173: Right now, any node event will trigger pods to be reconciled.  We can use a predicate to limit events so they are only generated on label changes, but it may be desirable to reconcile pods on all node events.

controllers/custom/custom_controller.go
```diff
@@ -221,8 +239,15 @@ func (c *CustomController) processNextWorkItem() bool {
 
 func (c *CustomController) reconcileHandler(obj interface{}) bool {
 	var req Request
+	var normalRequest reconcile.Request
 	var ok bool
-	if req, ok = obj.(Request); !ok {
+
+	// Requests generated through a watch return reconcile.Request
+	// instead of custom.Request, so we must convert it to custom.Request
+	if normalRequest, ok = obj.(reconcile.Request); ok {
```
dewjam, line 247: I'm not thrilled with this approach, but its required unless a custom Event Handler is created which returns a custom.Request instead.

haouc:
@dewjam due to the confirmed workaround [here](https://github.com/aws/karpenter/issues/1252), is that ok to close this PR for now? Thanks.

PR #105, opened 2022-05-20
add integration test script

sushrk:
*Description of changes:*
Add integration test script to run all integration test with default vpc-rc.

Test run(redacted for brevity):

---
Running VPC Resource Controller integration test with the following variables
KUBE CONFIG: /root/.kube/config
CLUSTER_NAME: networking-prow-2716796472
REGION: us-west-2
Running Suite: Per Pod Security Group Suite - /repository/vpc-resource-controller/test/integration/perpodsg
..

Ginkgo ran 1 suite in 24m42.957284916s
Test Suite Passed
Running Suite: Windows Integration Test Suite - /repository/vpc-resource-controller/test/integration/windows
..

Ginkgo ran 1 suite in 1h12m3.612501721s
Test Suite Passed
Running Suite: Validating WebHook Test Suite - /repository/vpc-resource-controller/test/integration/webhook
..

Ginkgo ran 1 suite in 2m23.756179517s
Test Suite Passed

---


jayanthvn (approved):
lgtm

scripts/test/lib/cluster.sh
```diff
@@ -0,0 +1,36 @@
...
+  AWS_MAX_ATTEMPTS=10 aws iam attach-role-policy \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSVPCResourceController \
+    --role-name "$ROLE_NAME" > /dev/null
+}
+
+function detach_controller_policy_cluster_role() {
+  echo "Detaching the IAM Policy from Cluster Service Role"
+  aws iam detach-role-policy \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSVPCResourceController \
+    --role-name $ROLE_NAME > /dev/null
```
cgchinmay, line 26: 1. What happens if there is a failure in Deletion, does the script terminate ? 
2. Wondering if we should preserve the output of these commands instead > /dev/null, would help us debug if there are any errors.

sushrk: 1. Yeah, the script would exit as we have "set -e"
2. I think the error would still be printed out, did a small test by trying to detach the policy from a role which doesn't have the policy:
`An error occurred (NoSuchEntity) when calling the DetachRolePolicy operation: Policy arn:aws:iam::aws:policy/AmazonEKSVPCResourceController was not found.`

scripts/test/install-vpc-cni.sh
```diff
@@ -5,7 +5,7 @@
 set -eo pipefail
 
 SCRIPTS_DIR=$(cd "$(dirname "$0")" || exit 1; pwd)
-DEFAULT_VPC_CNI_VERSION="v1.7.10"
+DEFAULT_VPC_CNI_VERSION="1.11"
```
jayanthvn, line 8: Did you mean 1.11.0? we can use the latest 1.11.3.

sushrk: I wanted to pin the VPC CNI version to v1.11.0 for testing since that is the recommended version.

sushrk: I wanted to pin the VPC CNI version to v1.11.0 for testing since that is the recommended version.

sushrk: I've resolved this. We will use the latest patch version available for the release for testing as well.

jayanthvn (approved):
lgtm. I hope any failures we do the appropriate cleanup?

sushrk:
> lgtm. I hope any failures we do the appropriate cleanup?

That's right. The prow tester/sweeper image will cleanup resources in case of failures in the integration test script. For Github runner issues, test-with-eksctl.sh script will handle deletion of the IAM resources & cluster deletion.

PR #108, opened 2022-06-02
use go install to fix integration test error

sushrk:
*Description of changes:*

Since upgrading to Go 1.17, we need to switch to 'go install' to install controller-gen. 
From go docs:
```Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead.```

This is to resolve the following error while running integration test cases:
```
make: controller-gen: Command not found
/bin/sh: /home/github-runner/go/bin/controller-gen: No such file or directory
make: *** [generate] Error 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ ifeq (, $(findstring v0.6.2,$(shell controller-gen --version)))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2 ;\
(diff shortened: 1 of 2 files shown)


jayanthvn:
Integration test error seems to be different now?

sushrk:
> > Integration test error seems to be different now?
> 
> Recent test failures are due to Windows test timeout. Another issue is ginkgo is not honoring skip and is running all test cases in the suite which might be causing the timeout: https://github.com/aws/amazon-vpc-resource-controller-k8s/runs/6700749272?check_suite_focus=true#step:9:444

Timeout behavior issues could be due to github runners using ginkgo v1. Runners are updated to ginkgo v2 now. Re-triggering the test.

jayanthvn:
Thank you!

haouc (dismissed):
lgtm

haouc (approved):
lgtm

PR #113, opened 2022-06-22
merge test-prow to master

sushrk:
*Description of changes:*
Merge test-prow branch to master.
Tests are stable now and we can run tests directly against master branch.

### DIFF
--- a/scripts/test/lib/cluster.sh
+++ b/scripts/test/lib/cluster.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+LIB_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+source "$LIB_DIR"/k8s.sh
+
 function load_cluster_details() {
   CLUSTER_INFO=$(aws eks describe-cluster --name $CLUSTER_NAME --region $REGION $ENDPOINT_FLAG)
   VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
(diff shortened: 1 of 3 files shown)


scripts/test/lib/cluster.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+source k8s.sh
```
cgchinmay, line 3: I think good to add the entire path, like [here,](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/b0e626d0795aa6f991c1ae93cae462a077536245/scripts/test/install-vpc-cni.sh#L10) moving this file would break this.

sushrk: Yes, good catch! Will update, thanks.

scripts/test/lib/cluster.sh
```diff
@@ -32,5 +36,5 @@ function set_env_aws_node() {
 
   echo "Setting environment variable $KEY to $VAL on aws-node"
   kubectl set env daemonset aws-node -n kube-system $KEY=$VAL
-  kubectl rollout status ds -n kube-system aws-node
+  check_ds_rollout "aws-node" "kube-system"
```
orsenthil, line 39: Looks like this takes care of the timeout.

orsenthil (approved):
LGTM.

scripts/test/lib/cluster.sh
sushrk: Yes, check_ds_rollout() has a default timeout of 2m, which should be good: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/scripts/test/lib/k8s.sh#L36

cgchinmay (approved):

PR #114, opened 2022-06-24
Adding regression e2e tests based on controller emitted metrics

haouc:
*Issue #, if available:*

*Description of changes:*
We don't have a regression e2e to monitor the metrics changes when releasing new version of the controller. In this change, we are adding
1, new LOCAL controller e2e test suite on metrics
- Describe ENI API call volume change 
- Mem allocation change
- Goroutine change

2, using promethues to retrieve and filter metrics and thus we can add more check on emitted metrics later

*Test*
Tests were done in local environment, including ginkgo test in a three nodes cluster and executing the test script to scale up to 100 nodes. All tests were passed.


test/integration/metrics/metrics_test_thresholds.go
```diff
@@ -0,0 +1,15 @@
+package metrics
+
+type Thresholds struct {
+	MemAllocated       float64
+	GoRoutines         float64
+	DescribeENIApiCall float64
+}
+
+func NewThresholds() *Thresholds {
```
sushrk, line 9: Can we add a comment on how we determine the thresholds and how to update in future when we have more data and metrics?

orsenthil:
Hi @haouc , the code changes look good.  I am not sure why we need 'go routine' metrics.

test/integration/metrics/metrics_test_thresholds.go
haouc: The comment is added. Thanks.

haouc:
We are testing Describe ENI API call count and MEM allocated gauge to check the regression of the controller updates. We also include metrics check for go routines to make sure we have some visibility on routines management change.

orsenthil (approved):
LGTM

PR #115, opened 2022-07-01
add endpoint flag for beta test

sushrk:
*Description of changes:*
1. Add missing endpoint flag to fix beta prow job failures.
2. Fix regression test error- Use```eksctl``` to scale nodegroup. The test cluster creates unmanaged nodegroups and ```aws eks``` does not know about unmanaged nodegroups. 
```
An error occurred (ResourceNotFoundException) when calling the DescribeNodegroup operation: No node group found for name: None.
```

### DIFF
--- a/scripts/test/test-with-eksctl.sh
+++ b/scripts/test/test-with-eksctl.sh
@@ -158,16 +158,15 @@ function run_integration_test() {
   --aws-region=$AWS_REGION \
   --aws-vpc-id=$VPC_ID) || TEST_FAILED=true
 
+  # SGP + Metrics for Regression Test
   if [[ $additional_gingko_params == "--focus=LOCAL" ]]; then
-    # SGP + Metrics for Regression Test
-    ng=$(aws eks list-nodegroups --cluster-name $CLUSTER_NAME --query 'nodegroups[0]' --output text)
-    asg_name=$(aws eks describe-nodegroup --nodegroup-name $ng --cluster-name $CLUSTER_NAME --query 'nodegroup.resources.autoScalingGroups[?amiType=='AL2_x86_64'].name' --output text)
-    asg_config=$(aws eks describe-nodegroup --nodegroup-name $ng --cluster-name $CLUSTER_NAME --query 'nodegroup.scalingConfig')
-    max_size=$(echo $asg_config | jq '.maxSize')
-    min_size=$(echo $asg_config | jq '.minSize')
-    desired_size=$(echo $asg_config | jq '.desiredSize')
     # We need a larger group of nodes to test the regression of controllers
-    aws autoscaling update-auto-scaling-group --auto-scaling-group-name $asg_name --max-size=150 --desired-capacity=100 --min-size=0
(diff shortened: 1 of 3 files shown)


scripts/test/run-integration-tests.sh
```diff
@@ -34,6 +34,10 @@ KUBE CONFIG: $KUBE_CONFIG_PATH
 CLUSTER_NAME: $CLUSTER_NAME
 REGION: $REGION"
 
+if [[ -n "${ENDPOINT}" ]]; then
```
cgchinmay, line 37: I think we can move this logic inside cluster.sh. Since that script is used by run-canary-tests.sh as well so would be good to have common logic in 1 place.

sushrk: Makes sense, will do that.

scripts/test/test-with-eksctl.sh
```diff
@@ -158,16 +158,15 @@ function run_integration_test() {
...
   if [[ $additional_gingko_params == "--focus=LOCAL" ]]; then
-    # SGP + Metrics for Regression Test
-    ng=$(aws eks list-nodegroups --cluster-name $CLUSTER_NAME --query 'nodegroups[0]' --output text)
-    asg_name=$(aws eks describe-nodegroup --nodegroup-name $ng --cluster-name $CLUSTER_NAME --query 'nodegroup.resources.autoScalingGroups[?amiType=='AL2_x86_64'].name' --output text)
-    asg_config=$(aws eks describe-nodegroup --nodegroup-name $ng --cluster-name $CLUSTER_NAME --query 'nodegroup.scalingConfig')
-    max_size=$(echo $asg_config | jq '.maxSize')
-    min_size=$(echo $asg_config | jq '.minSize')
-    desired_size=$(echo $asg_config | jq '.desiredSize')
     # We need a larger group of nodes to test the regression of controllers
-    aws autoscaling update-auto-scaling-group --auto-scaling-group-name $asg_name --max-size=150 --desired-capacity=100 --min-size=0
```
cgchinmay, line 170: This does fix the issue but it also means that we cannot use it against beta cluster as eksctl isn't supported. A suggestion lets skip this test if its beta and follow it up with another PR to make it work against beta clusters

cgchinmay:
Could you also manually trigger Manual Integration test workflow against this PR and verify if it works

scripts/test/test-with-eksctl.sh
```diff
@@ -158,16 +158,15 @@ function run_integration_test() {
...
+    ng=$(eksctl get nodegroup --cluster=$CLUSTER_NAME --output=json | jq '.[0]')
```
cgchinmay, line 164: This does fix the issue but it also means that we cannot use it against beta cluster as eksctl isn't supported. A suggestion - lets skip this test for now if an endpoint specified which means its beta and follow it up with another PR to make it work against beta clusters

sushrk:
> Could you also manually trigger Manual Integration test workflow against this PR and verify if it works

The integration test did run against the latest changes in this PR.. verified that it has succeeded.

cgchinmay:
lgtm

orsenthil (approved):
LGTM.

PR #130, opened 2022-09-28
refine github workflows

M00nF1sh:
*Description of changes:*
refine github workflows, mainly:
1. migrate off self-host workers with Github workers & OIDC instead
2. separate integration test after push and for pull request(which requires manual workflow dispatch)
3. merge build & unit test into a single workerflow
4. explicitly define permissions for workflow files
5. use latest dependencies such as checkout@v3, golang/1.19, eks 1.23
6. fix linting issues for 1.19, use bash to run shell scripts


jayanthvn (approved):
lgtm

Issue #128, opened 2022-09-21
The controller shouldn't requeue not found pods


haouc:
**What would you like to be enhanced**:
The controller shouldn't requeue pods which are not found. Not found will remain as not-found till cache updates and requeue them.
The similar improvement in controller runtime can be referenced at https://github.com/kubernetes-sigs/controller-runtime/issues/377

```
{"level":"error","timestamp":"2022-09-13T18:23:20.282Z","logger":"vpc.amazonaws.com/pod-eni-worker","msg":"re-queuing job","job":{"Operation":"Create","UID":"","PodName":"hello-world-7c4b75bf8c-tzzp8","PodNamespace":"default","RequestCount":1,"NodeName":""},"retry count":1,"error":"failed to find pod default/hello-world-7c4b75bf8c-tzzp8","stacktrace":"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).processNextItem\n\t/workspace/pkg/worker/worker.go:154\ngithub.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).runWorker\n\t/workspace/pkg/worker/worker.go:132"}
{"level":"error","timestamp":"2022-09-13T18:24:19.143Z","logger":"vpc.amazonaws.com/pod-eni-worker","msg":"re-queuing job","job":{"Operation":"Create","UID":"","PodName":"hello-world-7c4b75bf8c-tzzp8","PodNamespace":"default","RequestCount":1,"NodeName":""},"retry count":3,"error":"failed to find pod default/hello-world-7c4b75bf8c-tzzp8","stacktrace":"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).processNextItem\n\t/workspace/pkg/worker/worker.go:154\ngithub.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).runWorker\n\t/workspace/pkg/worker/worker.go:132"}
[...]

PR #133, opened 2022-10-05
avoid requeuing not found pods

haouc:
*Issue #, if available:*
#128 
*Description of changes:*
If the pod is not found, the call returns k8s API errors NotFound. If the job doesn't find the pod, worker will forget the item. If the pod has update, cache client will fetch the event and create another job for queuing.

Test result
```
{"level":"error","timestamp":"2022-10-05T00:52:26.642Z","logger":"vpc.amazonaws.com/pod-eni-worker","msg":"won't requeue a not found errored job","job":{"Operation":"Create","UID":"","PodName":"hello-world-5d97946fc7-8sgvz","PodNamespace":"default","RequestCount":1,"NodeName":""},"job":{"Operation":"Create","UID":"","PodName":"hello-world-5d97946fc7-8sgvz","PodNamespace":"default","RequestCount":1,"NodeName":""},"error":" \"default/hello-world-5d97946fc7-8sgvz\" not found","stacktrace":"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).processNextItem\n\t/workspace/pkg/worker/worker.go:164\ngithub.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker.(*worker).runWorker\n\t/workspace/pkg/worker/worker.go:141"}
```

### DIFF
--- a/pkg/k8s/pod/client_wrapper.go
+++ b/pkg/k8s/pod/client_wrapper.go
@@ -19,7 +19,9 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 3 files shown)


jdn5126 (approved):
Change LGTM, but just so I understand, if the job was not found, we previously requeued the job and it just spun indefinitely?

haouc:
> Change LGTM, but just so I understand, if the job was not found, we previously requeued the job and it just spun indefinitely?

yes,,in some cases, the pod was queued without DELETE event. Since our old logic only checks if the node object is cached and not check if the node is deleted and removed from cache, we think the node is not ready yet and indefinitely requeue the pod to wait the node being ready.

orsenthil (commented):
LGTM (post merge).

Issue #134, opened 2022-10-10
Observed authentication behavior inconsistent with documentation


JacobHenner:
**Describe the Bug**:

Starting on Friday, we noticed that pods in per-pod-securitygroups enabled namespaces were failing to start. The pod events showed that vpc-resource-controller was failing to annotate the affected pods due to PodSecurityPolicy (PSP). Oddly, there was no known change to cluster RBAC or PSP during this time.

It appears as if over the weekend the identity vpc-resource-controller uses to authenticate to our clusters changed, and that the identity used is no longer the service account added to the documentation in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/89, visible [here](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/docs/troubleshooting.md#psp-blocking-controller-annotations). 

Before the issue started the apiserver audit logs show the vpc-resource-controller authenticating to the cluster as ServiceAccount `eks-vpc-resource-controller`. After the issue started, audit logs show vpc-resource-controller authenticating as User `eks:vpc-resource-controller`. Our PSPs allow the former but not the latter. I've confirmed that ServiceAccount `eks-vpc-resource-controller` still exists within our clusters, and has remained undisturbed between before/after.

**Expected Behavior**:

vpc-resource-controller would continue to authenticate as ServiceAccount `eks-vpc-resource-controller`, or the linked documentation should include User `eks:vpc-resource-controller`.

**How to reproduce it (as minimally and precisely as possible)**:

Unknown, as this appears to be a change on the AWS-managed-control-plane side, and vpc-resource-controller logs are not exposed in cloudwatch.

This issue started Friday morning EDT in one cluster, and then progressed to at least two other clusters over the weekend. All clusters exist in us-east-1.

**Additional Context**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.21
- CNI Version: `602401143452.dkr.ecr.us-east-1.amazonaws.com/amazon-k8s-cni:v1.11.3`
- OS (Linux/Windows): Linux

See also: Support case 10949972441

haouc:
@JacobHenner thanks for sharing and reporting the issue. Yes, EKS just released the User based authentication for this controller to be consistent with other EKS services and avoid confusion to users. I will sync with our doc team to update the doc and call out this change specifically.

PR #135, opened 2022-10-10
update troubleshooting for using PSP with User authentication

haouc:
*Issue #, if available:*
#134 
*Description of changes:*
EKS recently switched authentication from service account to User for VPC resource controller. This could be a breaking change. We need call this out in our doc. A PR was also created to update [amazon-eks-user-guide](https://github.com/awsdocs/amazon-eks-user-guide/pull/612)

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -230,12 +230,15 @@ If the Pod is still stuck in `ContainerCreating` you can,
 ## List of Common Issues  
 
 ### PSP Blocking Controller Annotations
-If you have a PSP that blocks annotation to Pod, you will have to allow annotation from the following Service Account
+If you have a PSP that blocks annotation to Pod, you will have to allow annotation from the following User `eks:vpc-resource-controller`
 ```
 subjects:
   - kind: Group
     apiGroup: rbac.authorization.k8s.io
     name: system:authenticated
+  - apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -230,12 +230,15 @@ If the Pod is still stuck in `ContainerCreating` you can,
...
 ### PSP Blocking Controller Annotations
-If you have a PSP that blocks annotation to Pod, you will have to allow annotation from the following Service Account
+If you have a PSP that blocks annotation to Pod, you will have to allow annotation from the following User `eks:vpc-resource-controller`
 ```
 subjects:
   - kind: Group
     apiGroup: rbac.authorization.k8s.io
     name: system:authenticated
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
```
jdn5126, line 240: Shouldn't the order be switched here? i.e.
```
- kind: User
  apiGroup: rbac.authorization.k8s.io
  name: eks:vpc-resource-controller
```

haouc: The order should also work. @jdn5126. 
I can change it to be better aligned with documented format.
https://kubernetes.io/docs/reference/access-authn-authz/rbac/

haouc (commented):
Change the order to be aligned with documented way.

jdn5126 (approved):

Issue #124, opened 2022-08-01
Pod IP address assignment slow/stuck


fernandesnikhil:
**Describe the Bug**:
Pods using the Security Groups for pods feature remain stuck in the `ContainerCreating` status waiting for an IP address to be assigned. We notice the pods do not have an IP address assigned and the most recent event on the pod is like the following:
```
Warning  FailedCreatePodSandBox  6s (x175 over 38m)  kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "928809399f9e9384f0a1f97c7dace19e9ee3e33596428695fa6dda205a987bd0": plugin type="aws-cni" name="aws-cni" failed (add): add cmd: failed to assign an IP address to container
```

**Observed Behavior**:
Pods using the Security Groups for pods feature are stuck waiting in excess of 10 mintues for IP address assignment.

**Expected Behavior**:
Pods using the Security Groups for pods feature are assigned IP addresses quickly (not sure what's the EKS SLA here, I assume in a minute or two?)

**How to reproduce it (as minimally and precisely as possible)**:
Unknown

**Additional Context**:
We have had several instances of this happening on a single EKS cluster. In the worst case it we have noted pods waiting over 100 minutes for an IP address. We never see these pods receive an IP address. We keep deleting these stuck pods until we find the newly created pods have an IP address assigned within a few minutes after the pod has been created. These incidents where the VPC resource controller appears to be unable to assign pod IPs for pods using the Security Groups for pods feature seem to happen sporadically and eventually recovers on it's own.

**Environment**:
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.6", GitCommit:"f59f5c2fda36e4036b49ec027e556a15456108f0", GitTreeState:"archive", BuildDate:"1980-01-01T00:00:00Z", GoVersion:"go1.16.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.13-eks-84b4fe6", GitCommit:"e1318dce57b3e319a2e3fecf343677d1c4d4aa75", GitTreeState:"clean", BuildDate:"2022-06-09T18:22:07Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

- CNI Version: `v1.11.0`
[...]

Issue #104, opened 2022-05-11
Remove the Low QPS threshold for SGPP


orsenthil:
**What would you like to be enhanced**:

Currently, SGPP assignment by design , a low QPS call done VPC Resource Controller when it tries to associate Trunk ENI to Branch.
 
https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/config/loader.go#L29-L30

That is because SGP by costly operation, and this will be lower qps set than any other EC calls.

However, we should consider removing that low QPS call and let the max pod creation threshold be handled by kubernetes.

**Why is the change needed and what use case will it solve**:

There was  customer report that when SGPP was enabled, their deployment scaling took a long time, and they wanted to avoid it

bwhaley:
Folks, any updates on when this will be prioritized? SGFP is effectively unusable in our environment (which is not even that big in terms of AWS customer scale) due to this low limit.

PR #136, opened 2022-10-16
tuning workers count and QPS limiter

haouc:
*Issue #, if available:*
#104 #124 
*Description of changes:*
We should increase the workers and QPS limit on client and user calling to EC2 API to help minimize the delay/latency when large cluster/node groups churning and restarting the vpc resource controller for various reasons (such as control plane updates, control plane scaling, the controller update).

Tests have been done with clusters scale of 100 nodes, 500 nodes, and 1000 nodes. Pods using the security group for pods feature have been tested with scale from 500 up to 20k.

### DIFF
--- a/pkg/utils/httpClient.go
+++ b/pkg/utils/httpClient.go
@@ -23,6 +23,7 @@ import (
 // NewRateLimitedClient returns a new HTTP client with rate limiter.
 func NewRateLimitedClient(qps int, burst int) (*http.Client, error) {
 	if qps == 0 {
+		fmt.Printf("Creating a default rate limited http client with QPS = %d, Burst = %d\n", qps, burst)
 		return http.DefaultClient, nil
 	}
 	if burst < 1 {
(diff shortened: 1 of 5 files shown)


jdn5126:
Changes lgtm, will give others a chance to review before approving.

jdn5126 (approved):

pkg/utils/httpClient.go
```diff
@@ -23,6 +23,7 @@ import (
 // NewRateLimitedClient returns a new HTTP client with rate limiter.
 func NewRateLimitedClient(qps int, burst int) (*http.Client, error) {
 	if qps == 0 {
+		fmt.Printf("Creating a default rate limited http client with QPS = %d, Burst = %d\n", qps, burst)
```
kishorj, line 26: This should be logged instead of printf

haouc: Thanks for pointing this out. This is no longer needed since we are setting the QPS limit instead of leaving them as 0. I am pushing a new commit removing this.

kishorj (approved):

PR #139, opened 2022-10-18
move to lease based resource lock

jyotimahapatra:
*Issue #, if available:*

*Description of changes:*
Kubernetes contains a MultiLock based lock in `coordination/v1` api. 
We will take the change through multiple iterations. In the first iteration, we're moving to multi lock, that allows safe migration. In a future version, we will move multilock into lease lock and complete the migration.

Guidance: https://github.com/kubernetes/kubernetes/issues/107454

### DIFF
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -30,7 +30,7 @@ rules:
   verbs:
   - create
 - apiGroups:
-    - ""
+    - coordination.k8s.io
   resources:
     - leases
   verbs:
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -30,13 +30,17 @@ rules:
   verbs:
   - create
 - apiGroups:
-    - coordination.k8s.io
+  - coordination.k8s.io
   resources:
-    - leases
+  - leases
   verbs:
-    - get
-    - list
-    - watch
-    - create
-    - update
-    - patch
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resourceNames:
+  - cp-vpc-resource-controller
+  resources:
+  - leases
+  verbs:
+  - get
+  - update
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -29,3 +29,14 @@ rules:
   - events
   verbs:
   - create
+- apiGroups:
+    - ""
+  resources:
+    - leases
+  verbs:
+    - get
+    - list
+    - watch
+    - create
+    - update
+    - patch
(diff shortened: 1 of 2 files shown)


jdn5126 (approved):

haouc:
I think we need to test this change in control plane although I don't think this will make regression.

main.go
```diff
@@ -216,7 +216,7 @@ func main() {
 		RetryPeriod:                &retryPeriod,
 		LeaderElectionID:           config.LeaderElectionKey,
 		LeaderElectionNamespace:    config.LeaderElectionNamespace,
-		LeaderElectionResourceLock: resourcelock.ConfigMapsResourceLock,
```
kishorj, line 219: we'd also need rbac permissions for lease resources. 
config/rbac/leader_election_role.yaml and the installation manifests

config/rbac/leader_election_role.yaml
```diff
@@ -29,3 +29,14 @@ rules:
   - events
   verbs:
   - create
+- apiGroups:
+    - ""
```
kishorj, line 33: apigroup coordination.k8s.io for leases

main.go
orsenthil: @kishorj, I have updated. Please take a look.

config/rbac/leader_election_role.yaml
```diff
@@ -29,3 +29,14 @@ rules:
...
+  resources:
+    - leases
+  verbs:
```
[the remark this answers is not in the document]
orsenthil: Since we are replacing configmaps with leases, I was using the same verbs that configmaps already has.


```
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
```

I wasn't sure what is required for leader election. Is list, and watch required? I wasn't certain and so used the ones that we already used for configmap.

main.go
haouc: I have tested this change and it looks fine to me. Can you do the following changes
```
- apiGroups:
  - coordination.k8s.io
  resources:
  - leases
  verbs:
  - create
- apiGroups:
  - coordination.k8s.io
  resourceNames:
  - cp-vpc-resource-controller
  resources:
  - leases
  verbs:
  - get
  - update
```
in RBAC

and adding the annotation in main
```
// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=create
// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,namespace=kube-system,resourceNames=cp-vpc-resource-controller,verbs=get;update
```

Actually the annotation will auto add the permissions into config/rbac/role.yaml when building. Thanks.

haouc (approved):
lgtm, Thanks.

PR #141, opened 2022-10-18
Revert dependencies

haouc:
*Issue #, if available:*

*Description of changes:*
Due to some reported performance issues after we upgraded dependencies with v1.1.3, we would like to revert those updates for the next release. The future releases will add go and dependencies upgrades with more tests and more conservatively on their versions.


orsenthil (approved):
LGTM. Looks like we like to move a controller-runtime (0.9.x) instead of latest. These changes have been tested internally and we will upgrade after cutting a RC Release.

PR #153, opened 2022-10-28
increase concurrent routines for pod controller

haouc:
*Issue #, if available:*

*Description of changes:*
We increase concurrent routines in pod reconciling. This will help minimize pods creation latency, especially during high rate pods churn and restart of the controller.


sushrk (approved):
Thanks, lgtm!

PR #156, opened 2022-11-07
skip addon installation when addon info is not available

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Adding changes to handle canary test script failure when addon information is not available. In this case, we will skip installing the EKS-managed addon, and tests will be run against the default self-managed addon.


orsenthil (approved):
LGTM.

PR #157, opened 2022-11-11
create event controller, its tests, and RBAC for sgp feature

haouc:
*Issue #, if available:*

*Description of changes:*
We want to change how the controller knows to enable nodes with ENI trunking. The new strategy is VPC CNI broadcasting events and the controller consumes the events and then initialize network resources.
The workflow should be like 

1. user turns on the feature of Security Group for Pods
2. VPC CNI broadcasts event asking for trunk interface
3. the controller watches events and label nodes if events are valid
4. the controller initializes resources
5. VPC CNI finishes network setup on node

#### Tests
Since this change focus on how node gets initialized, we tested scale node to 1000 with the feature turned on.
Test Environment settings -> EKS v1.22, VPC CNI v1.11.4 or new version with the regarding changes on top of v1.11.4

Latest released controller v1.1.4
Scaling started at `Fri Nov 11 20:45:42 UTC 2022`
All nodes were Ready at `Fri Nov 11 20:47:53 UTC 2022`
All nodes were initialized with RNI trunking at `Fri Nov 11 20:58:08 UTC 2022`
Used time ~ 13 minutes

This change on top of v1.1.4
Scaling started at `Fri Nov 11 18:54:18 UTC 2022`
All nodes were Ready at `Fri Nov 11 18:57:32 UTC 2022`
All nodes were initialized with RNI trunking at `Fri Nov 11 19:07:15 UTC 2022`
Used time ~ 13 minutes

### DIFF
--- a/controllers/core/event_controller.go
+++ b/controllers/core/event_controller.go
@@ -64,10 +64,26 @@ var (
 		[]string{"operation"},
 	)
 
+	eventControllerSGPEventsFailureCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "failed_security_group_pod_events_reconciled_by_event_controller",
+			Help: "The number of SGP events that were reconcile by the controller but failed on label node",
+		},
+		[]string{"operation"},
+	)
+
 	eventControllerCNEventsCount = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "custom_networking_events_reconciled_by_event_controller",
 			Help: "The number of custom networking events that were reconcile by the controller",
+		},
+		[]string{"operation"},
+	)
+
+	eventControllerCNEventsFailureCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "failed_custom_networking_events_reconciled_by_event_controller",
+			Help: "The number of custom networking events that were reconcile by the controller but failed on label node",
 		},
 		[]string{"operation"},
 	)
@@ -99,6 +115,8 @@ func PrometheusRegister() {
 		metrics.Registry.MustRegister(eventControllerEventFailureCount)
 		metrics.Registry.MustRegister(eventControllerCNEventsCount)
 		metrics.Registry.MustRegister(eventReconcileOperationLatency)
+		metrics.Registry.MustRegister(eventControllerSGPEventsFailureCount)
+		metrics.Registry.MustRegister(eventControllerCNEventsFailureCount)
 		prometheusRegistered = true
 	}
 }
@@ -138,14 +156,24 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 				return ctrl.Result{}, nil
 			} else {
 				if r.isEventToManageNode(event) && r.isValidEventForSGP(event) {
-					r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, "true")
+					if err = r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, "true"); err != nil {
+						// by returning an error, we request that our controller will get Reconcile() called again
+						// we use the GetNode() as a safeguard to avoid repeating reconciling on deleted nodes
+						eventControllerSGPEventsFailureCount.WithLabelValues("trunk_label_failed").Inc()
+						return ctrl.Result{}, err
+					}
 					eventControllerSGPEventsCount.WithLabelValues("trunk_labelled").Inc()
 					r.Log.Info("Lable node with trunk label as true", "Node", node.Name, "Label", config.HasTrunkAttachedLabel)
 				}
 				if r.isValidEventForCustomNetworking(event) {
 					configKey, configName := parseEventMsg(event.Message)
 					if configKey == config.CustomNetworkingLabel {
-						r.K8sAPI.AddLabelToManageNode(node, config.CustomNetworkingLabel, configName)
+						if err = r.K8sAPI.AddLabelToManageNode(node, config.CustomNetworkingLabel, configName); err != nil {
+							// by returning an error, we request that our controller will get Reconcile() called again
+							// we use the GetNode() as a safeguard to avoid repeating reconciling on deleted nodes
+							eventControllerCNEventsFailureCount.WithLabelValues("eniconfig_label_failed").Inc()
+							return ctrl.Result{}, err
+						}
 						eventControllerCNEventsCount.WithLabelValues("eniconfig_labelled").Inc()
 						r.Log.Info("Label node with eniconfig label with configured name", "Node", node.Name, "Label", config.CustomNetworkingLabel)
 					}
--- a/controllers/core/event_controller.go
+++ b/controllers/core/event_controller.go
@@ -30,16 +30,27 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
-	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
 )
 
 const (
-	MaxEventConcurrentReconciles = 4
-	EventCreatedPastMinutes      = 2
-	EventFilterKey               = "reportingComponent"
+	MaxEventConcurrentReconciles  = 4
+	EventCreatedPastMinutes       = 2
+	EventFilterKey                = "reportingComponent"
+	LoggerName                    = "event"
+	TotalEventsCountMetrics       = "reconcile"
+	EventFailureCountMetrics      = "failing_list"
+	SGPEventsFailureCountMetrics  = "trunk_label_failed"
+	SGPEventsCountMetrics         = "trunk_labelled"
+	CNEventsFailureCountMetrics   = "eniconfig_label_failed"
+	CNEventsCountMetrics          = "eniconfig_labelled"
+	ReconcileEventsLatencyMetrics = "reconcile_events"
+	LabelOperation                = "operation"
+	LabelTime                     = "time"
+	EventRegardingKind            = "Node"
 )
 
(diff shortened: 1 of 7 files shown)


controllers/core/event_controller.go
```diff
@@ -0,0 +1,207 @@
...
+		if r.isValidEventForSGP(event) || r.isValidEventForCustomNetworking(event) {
+			nodeName := string(event.InvolvedObject.UID)
+
+			if node, err := r.K8sAPI.GetNode(nodeName); err != nil {
+				// if not found the node, we don't requeue and just wait VPC CNI to send another request
+				r.Log.V(1).Info("Event reconciler didn't find the node and wait for next request for the node", "Node", node.Name)
+				return ctrl.Result{}, nil
+			} else {
+				if r.isEventToManageNode(event) && r.isValidEventForSGP(event) {
+					r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, "true")
```
jayanthvn, line 141: Need to handle error scenario here?

controllers/core/event_controller.go
```diff
@@ -0,0 +1,207 @@
...
+					eventControllerSGPEventsCount.WithLabelValues("trunk_labelled").Inc()
+					r.Log.Info("Lable node with trunk label as true", "Node", node.Name, "Label", config.HasTrunkAttachedLabel)
+				}
+				if r.isValidEventForCustomNetworking(event) {
+					configKey, configName := parseEventMsg(event.Message)
+					if configKey == config.CustomNetworkingLabel {
+						r.K8sAPI.AddLabelToManageNode(node, config.CustomNetworkingLabel, configName)
```
jayanthvn, line 148: Same comment as above. Error handling is missing.

controllers/core/event_controller.go
```diff
@@ -0,0 +1,207 @@
...
+		metrics.Registry.MustRegister(eventControllerTotalEventsCount)
+		metrics.Registry.MustRegister(eventControllerSGPEventsCount)
+		metrics.Registry.MustRegister(eventControllerEventFailureCount)
+		metrics.Registry.MustRegister(eventControllerCNEventsCount)
+		metrics.Registry.MustRegister(eventReconcileOperationLatency)
+		prometheusRegistered = true
+	}
+}
+
+func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
haouc: Since the node will [stay as managed](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/4cfca50bbfff4410386d1565f5c78195ac34272b/pkg/node/manager/manager.go#L225-L237), we will only check if their subnet has changed.
https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/node/manager/manager.go#L202-L210

controllers/core/event_controller.go
haouc: Added error handling and unit test for error. Thanks.

controllers/core/event_controller.go
haouc: Added error handling and unit test for error. Thanks.

controllers/core/event_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		metrics.Registry.MustRegister(eventControllerCNEventsFailureCount)
+		prometheusRegistered = true
+	}
+}
+
+func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	start := time.Now()
+	var err error
+
+	logger := r.Log.WithValues("event", req.Name)
```
jayanthvn, line 128: nit: maybe we can add const for things like "event", "reconcile"..similar to this - `EventFilterKey = "reportingComponent"`

controllers/core/event_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	events, err := r.K8sAPI.ListEvents(ops)
+	if err != nil {
+		logger.Error(err, "List events failed")
+		eventControllerEventFailureCount.WithLabelValues("failing_list").Inc()
+		return ctrl.Result{}, err
+	}
+
+	for _, event := range events.Items {
+		// only check and process events were created in the past EventCreatedPastMinutes minutes
+		if !event.GetCreationTimestamp().After(time.Now().Add(-time.Minute * EventCreatedPastMinutes)) {
```
jayanthvn, line 147: We seemed to be picking events in past 2 mins, Is this number based on scale testing? Also if event is continuously generated by the same node don't we need to skip those?

controllers/core/event_controller.go
```diff
@@ -0,0 +1,235 @@
...
+			if node, err := r.K8sAPI.GetNode(nodeName); err != nil {
+				// if not found the node, we don't requeue and just wait VPC CNI to send another request
+				r.Log.V(1).Info("Event reconciler didn't find the node and wait for next request for the node", "Node", node.Name)
+				return ctrl.Result{}, nil
+			} else {
+				if r.isEventToManageNode(event) && r.isValidEventForSGP(event) {
+					if err = r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, "true"); err != nil {
+						// by returning an error, we request that our controller will get Reconcile() called again
+						// we use the GetNode() as a safeguard to avoid repeating reconciling on deleted nodes
+						eventControllerSGPEventsFailureCount.WithLabelValues("trunk_label_failed").Inc()
```
jayanthvn, line 162: nit: Same here "trunk_label_failed", "trunk_labelled", we can move it to a const

controllers/core/event_controller.go
```diff
@@ -0,0 +1,235 @@
...
+						return ctrl.Result{}, err
+					}
+					eventControllerSGPEventsCount.WithLabelValues("trunk_labelled").Inc()
+					r.Log.Info("Lable node with trunk label as true", "Node", node.Name, "Label", config.HasTrunkAttachedLabel)
+				}
+				if r.isValidEventForCustomNetworking(event) {
+					configKey, configName := parseEventMsg(event.Message)
+					if configKey == config.CustomNetworkingLabel {
+						if err = r.K8sAPI.AddLabelToManageNode(node, config.CustomNetworkingLabel, configName); err != nil {
```
jayanthvn, line 171: If the trunk label succeeded but not the custom networking label, based on your comment reconciler should pick the event again and we will again try to add the trunk label in line 158 isn't it? wondering if client-go allows relabelling?

controllers/core/event_controller.go
haouc: Yes, the test was scaled to 1000 nodes. We should be safe for even larger clusters since VPC CNI will keep sending event till trunk is attached. Actually the node added by RC will no longer be processed since this code
```
func (k *k8sWrapper) AddLabelToManageNode(node *v1.Node, labelKey string, labelValue string) error {
	if node.Labels[labelKey] == labelValue {
		return nil
	} else {
		newNode := node.DeepCopy()
		newNode.Labels[labelKey] = labelValue
		return k.cacheClient.Status().Patch(k.context, newNode, client.MergeFrom(node))
	}
}
```

controllers/core/event_controller.go
haouc: @jayanthvn same as the above comment, the `AddLabelToManagedNode` will not relabel the node object if the label matches requested. Requeued requests should be safe if one of them fails in the previous attempt.

controllers/core/event_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	corev1 "k8s.io/api/core/v1"
```
jayanthvn, line 33: Should we move to v1 events? I think the corev1 events will be deprecated - https://github.com/kubernetes-sigs/controller-runtime/issues/1462

pkg/k8s/wrapper.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	appV1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
```
jayanthvn, line 25: Nit: coreapiv1 or apicorev1 ? Similarly for the previous - apiappv1...

controllers/core/event_controller.go
haouc: Makes sense. I upgraded to k8s.io events v1. Thanks

controllers/core/event_controller.go
```diff
@@ -0,0 +1,250 @@
...
+					labelled, err := r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, config.BooleanTrue)
+					if err != nil {
+						// by returning an error, we request that our controller will get Reconcile() called again
+						// we use the GetNode() as a safeguard to avoid repeating reconciling on deleted nodes
+						eventControllerSGPEventsFailureCount.WithLabelValues(SGPEventsFailureCountMetrics).Inc()
+						return ctrl.Result{}, err
+					}
+					if labelled {
+						eventControllerSGPEventsCount.WithLabelValues(SGPEventsCountMetrics).Inc()
+						r.Log.Info("Lable node with trunk label as true", "Node", node.Name, "Label", config.HasTrunkAttachedLabel)
```
jayanthvn, line 177: nit : Lable -> Label

jayanthvn (approved):
Just a minor nit, rest lgtm.

controllers/core/event_controller.go
haouc: Thanks, fixed in a commit

jayanthvn (approved):

PR #160, opened 2022-12-13
adding cache for event controller querying node instances

haouc:
make change to use const for cache config

add prometheus metrics on cache ops

*Issue #, if available:*
N/A
*Description of changes:*
For larger clusters, events emitted can be very large volume. In order to maintain the event controller performance under large load situation, we should avoid calling k8s and the controller's APIs for already processed nodes although the clients are cached. 
I am adding a [BigCache](https://github.com/allegro/bigcache) which is a TTL cache with configuration as following

- sharding: 32
- max size: 2 MB
- TTL in minutes: 30

The change has been tested with Security Group for Pods with/without custom networking enabled.
By monitoring the added prometheus metrics around the cache ops, we can tell the difference from a 500 nodes cluster scaling its nodes from 0 to 500.
```
# TYPE event_reconcile_node_cache_operations counter
event_reconcile_node_cache_operations{cache_hit="",cache_miss="NODE_NOT_CACHED",cache_set_error=""} 1296
event_reconcile_node_cache_operations{cache_hit="NODE_CACHED",cache_miss="",cache_set_error=""} 1.0047677e+07
# HELP event_reconcile_node_cache_size Event controller node cache size
# TYPE event_reconcile_node_cache_size gauge
event_reconcile_node_cache_size 500
```

### DIFF
--- a/controllers/core/event_controller.go
+++ b/controllers/core/event_controller.go
@@ -31,6 +31,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	eventsv1 "k8s.io/api/events/v1"
+
+	bigcache "github.com/allegro/bigcache/v3"
 )
 
 const (
@@ -47,6 +49,12 @@ const (
 	ReconcileEventsLatencyMetrics = "reconcile_events"
 	LabelOperation                = "operation"
 	LabelTime                     = "time"
+	LabelCacheHit                 = "cache_hit"
+	LabelCacheMiss                = "cache_miss"
+	LabelCacheSetErr              = "cache_set_error"
+	CacheHitMetricsValue          = "NODE_CACHED"
+	CacheMissMetricsValue         = "NODE_NOT_CACHED"
+	CacheErrMetricsValue          = "NODE_CACHE_ERROR"
 	EventRegardingKind            = "Node"
 )
 
@@ -56,7 +64,15 @@ type EventReconciler struct {
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 	K8sAPI k8s.K8sWrapper
+	cache  *bigcache.BigCache
 }
+
+type Feature int
+
+const (
+	EnableSGP Feature = iota
+	EnableCN
+)
 
 var (
 	eventControllerTotalEventsCount = prometheus.NewCounterVec(
@@ -115,8 +131,32 @@ var (
 		[]string{LabelTime},
 	)
 
+	eventControllerEventNodeCacheSize = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "event_reconcile_node_cache_size",
+			Help: "Event controller node cache size",
+		},
+	)
+
+	eventControllerEventNodeCacheOpsCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "event_reconcile_node_cache_operations",
+			Help: "The number of ops on retrieving node from cache",
+		},
+		[]string{LabelCacheHit, LabelCacheMiss, LabelCacheSetErr},
+	)
+
 	prometheusRegistered = false
 )
+
+func NewEventReconciler(log logr.Logger, scheme *runtime.Scheme, k8sAPI k8s.K8sWrapper, nodeCache *bigcache.BigCache) *EventReconciler {
+	return &EventReconciler{
+		Log:    log,
+		Scheme: scheme,
+		K8sAPI: k8sAPI,
+		cache:  nodeCache,
+	}
+}
 
 func PrometheusRegister() {
 	if !prometheusRegistered {
@@ -127,6 +167,8 @@ func PrometheusRegister() {
 		metrics.Registry.MustRegister(eventReconcileOperationLatency)
 		metrics.Registry.MustRegister(eventControllerSGPEventsFailureCount)
 		metrics.Registry.MustRegister(eventControllerCNEventsFailureCount)
+		metrics.Registry.MustRegister(eventControllerEventNodeCacheSize)
+		metrics.Registry.MustRegister(eventControllerEventNodeCacheOpsCount)
 		prometheusRegistered = true
 	}
 }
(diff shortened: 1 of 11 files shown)


controllers/core/event_controller.go
```diff
@@ -47,7 +49,20 @@ const (
...
+	CacheMissMetricsValue         = "NODE_NOT_CACHED"
+	CacheErrMetricsValue          = "NODE_CACHE_ERROR"
 	EventRegardingKind            = "Node"
+	SGPPrefix                     = "SGP"
+	CNPrefix                      = "CN"
+
+	// customized configurations for BigCache
+	InstancesCacheTTL     = 30 * time.Minute // scaling < 1k nodes should be under 20 minutes
+	InstancesCacheShards  = 32               // must be power of 2
+	InstancesCacheMaxSize = 2                // in MB
```
jdn5126, line 65: With your 500 node test, how big was cache? Just curious is 2MB is enough

controllers/core/event_controller.go
```diff
@@ -47,7 +49,20 @@ const (
...
+	CacheHitMetricsValue          = "NODE_CACHED"
+	CacheMissMetricsValue         = "NODE_NOT_CACHED"
+	CacheErrMetricsValue          = "NODE_CACHE_ERROR"
 	EventRegardingKind            = "Node"
+	SGPPrefix                     = "SGP"
+	CNPrefix                      = "CN"
+
+	// customized configurations for BigCache
+	InstancesCacheTTL     = 30 * time.Minute // scaling < 1k nodes should be under 20 minutes
+	InstancesCacheShards  = 32               // must be power of 2
```
jdn5126, line 64: Looks like the key is `(SGPPrefix | CNPrefix) + hostId)`. For 1k nodes and 5 security groups, the space < 2^13, so I think 32 shards may be too many. 16 is probably sufficient and more efficient in the average case.

controllers/core/event_controller_test.go
```diff
@@ -181,18 +226,27 @@ func TestEventReconciler_Reconcile_SGPEvent(t *testing.T) {
 	for _, e := range events {
 		mock.MockK8sAPI.EXPECT().ListEvents(ops).Return(e.eventList, nil)
 
+		_, cacheErr := mock.Reconciler.cache.Get(e.testNodeName)
```
jdn5126, line 229: If I am reading this correctly, it looks like we are always expecting cache hit in this test case. Is that correct? Can we make sure to cover this code path with cache hit, cache miss, and cache expiry (simulate entry expiring before call)?

controllers/core/event_controller.go
haouc: Cache size was for 500 and roughly calculate 2MB should be good for up to 20k instance IDs with custom networking enabled and 40k instance IDs without custom networking. 20k entries plus 30 minutes expiry setting, I think 2MB should be good for cases I can think of. If cache gets full and thus causes cache miss, missing records should be ok for controller to function properly. But I am ok to increase it to 4MB just for in case there are very high churning rate on nodes. I mean the size should be enough for "static" group of nodes, but bigger might be safer for very "dynamic" group of nodes.

controllers/core/event_controller.go
haouc: Yeah, I was thinking the same thing. Sharding 32 might bring too much overhead. I would wait for other members' suggestions as well.

controllers/core/event_controller_test.go
haouc: Sure, Jeff. I can add test cases for all cases. Expiry can be tested with a setting with short time.

controllers/core/event_controller.go
```diff
@@ -47,7 +49,20 @@ const (
...
+	LabelCacheSetErr              = "cache_set_error"
+	CacheHitMetricsValue          = "NODE_CACHED"
+	CacheMissMetricsValue         = "NODE_NOT_CACHED"
+	CacheErrMetricsValue          = "NODE_CACHE_ERROR"
 	EventRegardingKind            = "Node"
+	SGPPrefix                     = "SGP"
+	CNPrefix                      = "CN"
+
+	// customized configurations for BigCache
+	InstancesCacheTTL     = 30 * time.Minute // scaling < 1k nodes should be under 20 minutes
```
jayanthvn, line 63: As discussed offline, lets check how to delete the cache entry when the node is getting deleted (maybe node controller can do it). Since for 30 mins we will have stale data and node churn is a very common scenario.

haouc: Th new updated commit has the change. Thanks.

jdn5126 (approved):

controllers/core/node_controller.go
```diff
@@ -67,7 +69,9 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	if err := r.Client.Get(ctx, req.NamespacedName, node); err != nil {
 		if errors.IsNotFound(err) {
-			_, found := r.Manager.GetNode(req.Name)
+			cachedNode, found := r.Manager.GetNode(req.Name)
+			// delete the not found node instance id from node event cache for housekeeping
+			r.deleteNodeFromNodeEventCache(cachedNode.GetNodeInstanceID())
```
jayanthvn, line 74: Don't we need error checks here? I assume we are getting from the node manager but what if node manager returns `nil` ?

controllers/core/event_controller.go
```diff
@@ -157,44 +199,82 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 				r.Log.V(1).Info("Event reconciler didn't find the node and wait for next request for the node", "Node", node.Name)
 				return ctrl.Result{}, nil
 			} else {
-				if r.isEventToManageNode(event) && r.isValidEventForSGP(event) {
+				eventControllerEventNodeCacheOpsCount.WithLabelValues([]string{"", CacheMissMetricsValue, ""}...).Inc()
+				eventControllerEventNodeCacheSize.Set(float64(r.cache.Len()))
+				if r.isEventToManageNode(event) && eventForSGP {
 					// make the label value to false that indicates the node is ok to be proceeded by providers
 					// provider will decide if the node can be attached with trunk interface
 					labelled, err := r.K8sAPI.AddLabelToManageNode(node, config.HasTrunkAttachedLabel, config.BooleanFalse)
```
jayanthvn, line 238: Nit: Maybe later we can move this logic into a small utility function to label the node and increment the counter.

haouc: Makes sense. I can follow up with this in another PR. Thanks.

controllers/core/event_controller.go
```diff
@@ -157,44 +199,82 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+			// we use the GetNode() as a safeguard to avoid repeating reconciling on deleted nodes
 			if node, err := r.K8sAPI.GetNode(nodeName); err != nil {
 				// if not found the node, we don't requeue and just wait VPC CNI to send another request
 				r.Log.V(1).Info("Event reconciler didn't find the node and wait for next request for the node", "Node", node.Name)
 				return ctrl.Result{}, nil
 			} else {
-				if r.isEventToManageNode(event) && r.isValidEventForSGP(event) {
+				eventControllerEventNodeCacheOpsCount.WithLabelValues([]string{"", CacheMissMetricsValue, ""}...).Inc()
+				eventControllerEventNodeCacheSize.Set(float64(r.cache.Len()))
+				if r.isEventToManageNode(event) && eventForSGP {
```
jayanthvn, line 235: This `isEventToManageNode` naming has to be changed since it is bit confusing to the reader and also seems to be redundant. This can be fixed in follow up PR.

controllers/core/node_controller.go
haouc: This is deleting the instance id from event cache key set. If that ops fails, we log the failure. But I think we can live with it and don't have to retry or bubble up the failure. The cache can evict old entries in the worst case.

controllers/core/event_controller.go
haouc: Yeah. This could be redundant. I can do a follow up PR to revisit this with tests. Thanks.

jayanthvn (approved):
lgtm

PR #170, opened 2023-01-29
fix a bug when cache for two features

haouc:
*Issue #, if available:*

*Description of changes:*
This bug was found during tests enabling custom networking. The cache entry's value is incorrectly initialized and updated. Instead of updating the feature value purely based on coming event, we need to update the value based on coming event and the already cached value for another feature flag. 

For example
If SGP value has been cached, in cache --> "instance-id": {1, 0}
when the custom networking event comes, the update should be 
- new entry value {0, 0}
- update the new entry based on CN feature {0, 1}
- update the new entry based on existing cache entry {1, 0} to {1, 1}
- save the entry {1, 1} into cache

This PR also adds a prometheus metrics for cache add, and fix a typo in previous code.

### DIFF
--- a/controllers/core/event_controller.go
+++ b/controllers/core/event_controller.go
@@ -51,9 +51,11 @@ const (
 	LabelTime                     = "time"
 	LabelCacheHit                 = "cache_hit"
 	LabelCacheMiss                = "cache_miss"
+	LabelCacheAdd                 = "cache_add"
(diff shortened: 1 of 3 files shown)


jdn5126 (commented):
Is the case motivating this change covered in `event_controller_test.go`? I think it would be good to add if not, so we prevent regression

jdn5126 (approved):

PR #174, opened 2023-02-08
use GET API to retrieve event

haouc:
*Issue #, if available:*

*Description of changes:*
LIST API causes large MEM and CPU usage. GET API use less MEM and much better CPU utilization.

Changes

1. changing from LIST to GET in event controller
2, update base image to latest
3, a bug when calling removing instance id from cache but the node is unmanaged

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -59,6 +59,7 @@ type Node interface {
 	IsManaged() bool
 
 	GetNodeInstanceID() string
+	HasInstance() bool
 }
 
 // NewManagedNode returns node managed by the controller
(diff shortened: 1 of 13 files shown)


haouc:
By switching to GET API, we are accepting the tradeoffs

1. If events are missed by the new controller, the controller will miss the moment to add the node to initialization process
2. The missing will add latency to the process but the VPC CNI will resend the event in 30 s since the node will not be labelled and attached with trunk interface.
3. Since the time of events delivery is not guaranteed, the latency will be hard to estimate and vary from environment to environment.
4. This will be eventually consistent system. We expect any missed nodes will be added successfully eventually.
5. GET will not provide filter comparing to LIST, the controller will have to filter from client side.

Pros to use GET API

1. Avoid larger memory allocation to maintain events list 
2. Avoid computing intensive of iterating LIST events whenever an event is watched

pkg/node/node.go
```diff
@@ -210,3 +219,10 @@ func (n *node) GetNodeInstanceID() string {
 
 	return n.instance.InstanceID()
 }
+
+func (n *node) HasInstance() bool {
+	n.lock.RLock()
+	defer n.lock.RUnlock()
```
jdn5126, line 225: Do you need a lock here just to read the value?

jdn5126 (approved):
Went through the PR and the changes make sense and LGTM. Are the integration tests the main mechanism for regression coverage?

PR #175, opened 2023-02-10
Integration test script enhancements to run on deveks

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
Integration test script enhancements to run tests on deveks cluster


scripts/test/lib/cluster.sh
```diff
@@ -18,6 +19,19 @@ function load_cluster_details() {
   echo "VPC ID: $VPC_ID, Service Role ARN: $SERVICE_ROLE_ARN, Role Name: $ROLE_NAME"
 }
 
+function load_deveks_cluster_details() {
+
+  PROVIDER_ID=$(kubectl get nodes -ojson | jq -r '.items[0].spec.providerID')
+  INSTANCE_ID=${PROVIDER_ID##*/}
+  VPC_ID=$(aws ec2 describe-instances --instance-ids ${INSTANCE_ID} --no-cli-pager | jq -r '.Reservations[].Instances[].VpcId')
```
haouc, line 26: You can get VPC ID directly from `aws eks describe-cluster --name $CLUSTER_NAME --endpoint $ENDPOINT --query 'cluster.resourcesVpcConfig.vpcId'`

scripts/test/lib/cluster.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
 LIB_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+ENDPOINT_FLAG=""
 
 source "$LIB_DIR"/k8s.sh
 
-if [[ -n "${ENDPOINT}" ]]; then
+if [[ ! -z "${ENDPOINT}" ]]; then
```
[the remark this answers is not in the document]
jaydeokar: Nevemind, this should not be needed, although it does the same work. I've reverted this

scripts/test/lib/cluster.sh
jaydeokar: This is cluster created with deveks, which doesn't get registered on this api. That's why the new function for deveks tests

haouc: Makes sense. Thanks.

haouc (approved):
lgtm

PR #177, opened 2023-02-14
Add more filters to bring down memory usage

haouc:
initialize unmanaged node as well

update base image to latest

*Issue #, if available:*

*Description of changes:*
In the PR, I am proposing three changes

1. using more specific filters/conditions in manager resource cache for events, event controller setup with manager. This can help avoid cache/watch too many events which can cause high memory/cpu utilization in some case.
2. initializing unmanaged nodes as well to avoid dereference nil pointer when calling those nodes' fields
3. updating base image to gain latest patches

Tests have been done for 

1. Integration tests for webhook, SGP feature, and Windows
2. Scaling tests have been done with 3k to 10k pods in a cluster having 400 - 1k nodes
3. Stress tests have been done with flooding events into the cluster every second
4. API call volume have been verified

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN GIT_VERSION=$(git describe --tags --always) && \
         CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
         -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" -a -o controller main.go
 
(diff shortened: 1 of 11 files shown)


jdn5126 (approved):
There seems to be some overlapping changes with #174, so I assume #174 will merge first and then this will be rebased? Either way, the content and coverage LGTM

haouc:
Thanks @jdn5126 , they are actually same except this one is using LIST api and the other one using GET api. The other one was created due to LIST had high memory usage. This one adds more filters / conditions into manager, cache and list option and significantly lower down the memory utilization in large scale clusters in terms of resources (nodes, pods, and even events). I am preferring this one since GET api may have potential latency although GET has lower CPU/MEM usage. I will keep the other one for in case that we may have to switch to GET for any reason in the future.

Issue #172, opened 2023-02-02
Update to newer k8s versions (and golang as well!)


dims:
currently we are using 1.21 k8s in our vendored dependencies:
https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/go.mod#L23-L24

upstream does not guarantee more than a skew of 2, so could we get this update do say 1.25?

while we are at it, please update to newer 1.19.x golang as well please.

thanks,
Dims

haouc:
@dims Thank you for creating this request. We are working on upgrading and testing the dependencies. I will provide updates here.

haouc:
We have upgraded dependencies in the PR #179 . Closing this issue.

PR #179, opened 2023-02-21
upgrade go, k8s, and controller-runtime

haouc:
*Issue #, if available:*
#172 
*Description of changes:*
We want to upgrade dependencies versions.

Tests:
Integration tests passed locally


go.mod
```diff
@@ -1,27 +1,82 @@
 module github.com/aws/amazon-vpc-resource-controller-k8s
 
-go 1.16
+go 1.20
 
 require (
 	github.com/allegro/bigcache/v3 v3.1.0
 	github.com/aws/amazon-vpc-cni-k8s v1.9.0
```
jdn5126, line 7: nit: while we are here, we may as well update VPC CNI dependency to latest version

go.mod
```diff
@@ -1,27 +1,82 @@
...
 	github.com/aws/aws-sdk-go v1.40.43
```
jdn5126, line 8: nit: this can be updated to `v1.43.29`

go.mod
```diff
@@ -1,27 +1,82 @@
...
-	github.com/go-logr/logr v0.4.0
-	github.com/go-logr/zapr v0.4.0
+	github.com/go-logr/logr v1.2.3
+	github.com/go-logr/zapr v1.2.3
 	github.com/golang/mock v1.4.4
```
jdn5126, line 11: nit: this can be `v1.6.0`

go.mod
```diff
@@ -1,27 +1,82 @@
...
+	go.uber.org/zap v1.24.0
+	golang.org/x/time v0.3.0
 	gomodules.xyz/jsonpatch/v2 v2.2.0
-	k8s.io/api v0.21.3
-	k8s.io/apimachinery v0.21.3
-	k8s.io/client-go v0.21.3
-	sigs.k8s.io/controller-runtime v0.9.5
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
+	k8s.io/client-go v0.26.1
```
jdn5126, line 25: nit: I think we should stay at `v0.25.5` here, same as CNI, until EKS support 1.26

haouc: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/go.mod has upgraded to v0.26 as well as controller-runtime. I understand they are matching the upstream release. I am trying to make this upgraded a little ahead since this controller upgrade need time. Please share the possible issues if using 0.26. Thanks.

jdn5126: Ah ok, that sounds reasonable to me then. My only concern was EKS coverage, but I guess we are already always running kops and deveks against latest upstream, so this should be fine

haouc: Yeah, I think we have k8s backwards compatibility promise to support old versions. Upgrading to 0.26 should be working fine with older EKS versions.

go.mod
jdn5126, line 7: Can you also update this to v1.12.2? I can approve after that

haouc: @jdn5126 vpc cni version was updated. thanks

jdn5126 (approved):

Issue #163, opened 2023-01-09
The controller should handle corner case that node's label map could be nil


haouc:
**What would you like to be enhanced**:
I want the controller handling some corner cases when nodes have nil as labels map. Currently the controller has no check on this. Since VPC CNI is transferring the labeling job to this controller, the controller should handle this case by
- either, retry
- or, add an empty map and then label the node

**Why is the change needed and what use case will it solve**:
Otherwise, a panic could be expected in the corner cases.

PR #180, opened 2023-03-02
refactor to use self managed retrywatcher

haouc:
*Issue #, if available:*

*Description of changes:*
Due to high level event APIs off-load pressures onto API sever and ETCD server, we need refactor the event controller to use low level watcher from client-go. RetryWatcher is an utility that can help restart the watcher when it stops for various reasons. Using this watcher, we can significantly reduce the resource usage and loads on servers side.

Tests have been done
- Functional integration tests
- 48 hours performance tests
- 48 hours dual labeling tests for both SGP and custom networking labels

Resource usage monitored
- 48 hours MEM/CPU usage monitoring
- 48 hours goroutines leak monitoring

### DIFF
--- a/pkg/k8s/wrapper.go
+++ b/pkg/k8s/wrapper.go
@@ -23,7 +23,6 @@ import (
 
 	appv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
-	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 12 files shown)


pkg/k8s/wrapper.go
```diff
@@ -207,16 +208,16 @@ func (k *k8sWrapper) AddLabelToManageNode(node *v1.Node, labelKey string, labelV
 		return false, nil
 	} else {
 		newNode := node.DeepCopy()
+		// in case if the node object has nil label map
```
haouc, line 211: This should prevent the issue #163 from happening

Issue #185, opened 2023-03-13
p4d.24xlarge does not have trunking enabled


leroylim20:
**Describe the Bug**:
From limits.go, p4d.2xlarge is trunking compatible

`"p4d.24xlarge":      {Interface: 60, IPv4PerInterface: 50, IsTrunkingCompatible: true, BranchInterface: 62},`

However, the node is initialised as `vpc.amazonaws.com/has-trunk-attached=false`

kubectl describe node ip-192-168-74-242.ap-northeast-1.compute.internal

```
Name:               ip-192-168-74-242.ap-northeast-1.compute.internal
Roles:              <none>
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=p4d.24xlarge
                    beta.kubernetes.io/os=linux
                    eks.amazonaws.com/capacityType=ON_DEMAND
                    eks.amazonaws.com/nodegroup=p4d-ng-test
                    eks.amazonaws.com/nodegroup-image=ami-03f2933a939357653
                    failure-domain.beta.kubernetes.io/region=ap-northeast-1
                    failure-domain.beta.kubernetes.io/zone=ap-northeast-1a
                    k8s.io/cloud-provider-aws=6fd6a5af17ce9e8e886d91d064ee43e9
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=ip-192-168-74-242.ap-northeast-1.compute.internal
                    kubernetes.io/os=linux
                    node.kubernetes.io/instance-type=p4d.24xlarge
                    topology.kubernetes.io/region=ap-northeast-1
                    topology.kubernetes.io/zone=ap-northeast-1a
                    vpc.amazonaws.com/has-trunk-attached=false
Annotations:        alpha.kubernetes.io/provided-node-ip: 192.168.74.242
                    node.alpha.kubernetes.io/ttl: 0
                    volumes.kubernetes.io/controller-managed-attach-detach: true
CreationTimestamp:  Mon, 13 Mar 2023 13:03:50 +0800
Taints:             <none>
Unschedulable:      false
Lease:
  HolderIdentity:  ip-192-168-74-242.ap-northeast-1.compute.internal
  AcquireTime:     <unset>
  RenewTime:       Mon, 13 Mar 2023 13:08:04 +0800
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
[...]

PR #186, opened 2023-03-14
limit file includes networkcards for controller to handle multicards

haouc:
*Issue #, if available:*
#185 
*Description of changes:*
Currently the controller can only support single card since VPC CNI can only support single network card. This file update provides Network cards information to the controller for provisioning interface correctly.


jdn5126 (approved):
Does an equivalent VPC CNI change need to be made for this to be taken advantage of?

orsenthil (approved):
LGTM. Please ensure that the file in vpc-cni repo is the same one too and diff is 0.

haouc:
> Does an equivalent VPC CNI change need to be made for this to be taken advantage of?

@jdn5126 @orsenthil I don't think VPC CNI need to use the same file as this controller uses because CNI maintains its own list of supported types for falling back if the call to EC2 doesn't work.

haouc:
This change is mainly for adding support for multi cards cases. Currently this controller doesn't know if the instance has multiple network cards and how many interfaces per card. This causes controller incorrectly assign interfaces to their available index. After adding these cards metadata in this file, controller will be able to use these info to correctly get index based on card interfaces range. We will need another PR to change the logic how search for index based on this change.

PR #190, opened 2023-03-20
check default card interfaces for index

haouc:
*Issue #, if available:*
#185 
*Description of changes:*
Instead of using instance max interfaces for index, we should use default network card's interfaces count, especially for multi-card instance types. Currently this controller and VPC CNI only support single network card.

### DIFF
--- a/pkg/aws/ec2/instance.go
+++ b/pkg/aws/ec2/instance.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 )
 
 // ec2Instance stores all the information that can be shared across the providers for an instance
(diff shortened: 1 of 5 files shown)


pkg/aws/ec2/instance.go
```diff
@@ -117,7 +118,23 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
...
-	i.deviceIndexes = make([]bool, limits.Interface)
+	defaultCardIdx := limits.DefaultNetworkCardIndex
+	var defaultNetworkCardLimit int64
+	for _, card := range limits.NetworkCards {
+		if card.NetworkCardIndex == int64(defaultCardIdx) {
+			defaultNetworkCardLimit = card.MaximumNetworkInterfaces
+			break
+		}
+	}
+	if defaultNetworkCardLimit == 0 {
```
sushrk, line 129: Curious, do we expect this value to be 0 ever?

sushrk (approved):
LGTM, thanks!

pkg/aws/ec2/instance.go
haouc: Shouldn't be. This check is for in case the limit file doesn't have cards loaded for the instance type. In the case, the `defaultNetworkCardLimit` will be default int value 0. In this case, we need to return an error.

PR #194, opened 2023-03-24
clean up node label introduced from event solution

haouc:
*Issue #, if available:*

*Description of changes:*
This is for clean up `not-supported` label for not supported ec2 instance type since CNI will also reset as well and not support this label.

### DIFF
--- a/pkg/node/manager/manager_test.go
+++ b/pkg/node/manager/manager_test.go
@@ -427,7 +427,7 @@ func Test_performAsyncOperation(t *testing.T) {
 	}
 
 	job.op = Init
-	mock.MockK8sAPI.EXPECT().GetNode(nodeName).Return(v1Node, nil)
+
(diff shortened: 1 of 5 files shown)


pkg/node/manager/manager_test.go
```diff
@@ -427,7 +427,8 @@ func Test_performAsyncOperation(t *testing.T) {
 	}
 
 	job.op = Init
-	mock.MockK8sAPI.EXPECT().GetNode(nodeName).Return(v1Node, nil)
+	// TODO: since diabled ops updating node label to true, we need disable call to GetNode for now
+	mock.MockK8sAPI.EXPECT().GetNode(nodeName).Return(v1Node, nil).Times(0)
```
sushrk, line 431: This should be removed as per comment above.

sushrk (commented):
General feedback: Should we move the changes to a new branch and clean up the master branch?

sushrk (approved):
lgtm, thanks!

Issue #199, opened 2023-04-04
Newer controller-gen doesn't correctly recognize markers. Need upgrade.


haouc:
**What would you like to be enhanced**:
Looks like the controller is still using 0.6.x controller-tools to manage markers. Controller-gen has a breaking change from 0.7 and the latest version is 0.11.x. This project need to upgrade markers to use newer controller-tools.
**Why is the change needed and what use case will it solve**:
This project is still using an old controller-tool version. It should be upgraded to work with newer controller-tools version.

sushrk:
Partly addressed in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/369

go/v4 is the default version now, https://book.kubebuilder.io/plugins/go-v4-plugin

Keeping this issue open till be move to v4.

PR #201, opened 2023-04-12
add healthz deep subpathes for all controllers

haouc:
*Issue #, if available:*

*Description of changes:*
We want to add deep health check probes into individual controllers. 

The current healthz `ping` is simple and basically just checking if the VPC RC manager goroutine (main routine) is not failing. Reported customers’ cases told us the main check may not be sufficient in some cases. What we want
- if one goroutine blocks resources, the healthz need to restart the entire controller
- the healthz probes should cover all sub controllers
- the heatlhz probe should have its own timeout instead of rely on kubelet timeout
- the health status monitored by kubelet should be aggregated (one failure should fails all)
- simple ping or complicate probing real data should depend on how critical and vulnerable the targeted controller may be.

Tests have been done
```
[+]health-annotation-validating-webhook ok
[+]health-branch-provider ok
[+]health-cm-controller ok
[+]health-custom-pod-controller ok
[+]health-deploy-controller ok
[+]health-interface-cleaner ok
[+]health-introspect-controller ok
[+]health-ipv4-provider ok
[+]health-node-controller ok
[+]health-node-manager ok
[+]health-node-validating-webhook ok
[+]health-pod-controller ok
[+]health-pod-mutating-webhook ok
[+]health-resource-manager ok
[+]health-root-manager-ping ok
healthz check passed
```

### DIFF
--- a/controllers/apps/deployment_controller.go
+++ b/controllers/apps/deployment_controller.go
@@ -16,15 +16,17 @@ package apps
 import (
 	"context"
 
-	"github.com/aws/amazon-vpc-resource-controller-k8s/controllers/core"
+	controllers "github.com/aws/amazon-vpc-resource-controller-k8s/controllers/core"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	rcHealthz "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/healthz"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node/manager"
 
 	"github.com/go-logr/logr"
 	appV1 "k8s.io/api/apps/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 )
 
 type DeploymentReconciler struct {
@@ -63,7 +65,13 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	return ctrl.Result{}, nil
 }
 
-func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager, healthzHandler *rcHealthz.HealthzHandler) error {
+	// add health check on subpath for deployment controller
+	// TODO: this is a simple controller and unlikely hit blocking issue but we can revisit this after subpaths are released for a while
+	healthzHandler.AddControllersHealthCheckers(
+		map[string]healthz.Checker{"health-deploy-controller": rcHealthz.SimplePing("deployment controller", r.Log)},
+	)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&appV1.Deployment{}).
 		Complete(r)
--- a/main.go
+++ b/main.go
@@ -29,6 +29,7 @@ import (
 	ec2API "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	rcHealthz "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/healthz"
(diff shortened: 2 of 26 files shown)


main.go
```diff
@@ -276,11 +275,15 @@ func main() {
...
 	supportedResources := []string{config.ResourceNamePodENI, config.ResourceNameIPAddress}
-	resourceManager, err := resource.NewResourceManager(ctx, supportedResources, apiWrapper)
+	resourceManager, err := resource.NewResourceManager(ctx, supportedResources, apiWrapper, ctrl.Log.WithName("managers").WithName("resource"))
 	if err != nil {
 		ctrl.Log.Error(err, "failed to init resources", "resources", supportedResources)
 		os.Exit(1)
 	}
+	// add health check on subpath for resource manager which includes providers as well
+	for key, value := range resourceManager.GetCheckers() {
+		healthHandler.AddControllerHealthChecker(key, value)
```
sushrk, line 285: Synced offline- should we pass the healthHandler to each controller to add health checks?

config/webhook/manifests.yaml
```diff
@@ -16,7 +16,6 @@ webhooks:
   failurePolicy: Ignore
   matchPolicy: Equivalent
   name: mpod.vpc.k8s.aws
-  timeoutSeconds: 5
```
sushrk, line 19: Synced offline- changes to this file should be reverted.

main.go
haouc: Thanks, I have changed to use parameters.

controllers/apps/deployment_controller.go
```diff
@@ -63,8 +66,10 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return ctrl.Result{}, nil
 }
 
-func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager) (healthz.Checker, error) {
+	r.Checker = rcHealthz.SimplePing("deployment controller", r.Log)
```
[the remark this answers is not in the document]
haouc: Currently, a simple ping should be good. I added a TODO annotation.

config/webhook/manifests.yaml
haouc: I reverted the file.

pkg/worker/worker.go
```diff
@@ -126,6 +126,12 @@ func (w *worker) SetWorkerFunc(workerFunc func(interface{}) (ctrl.Result, error)
 
 // SubmitJob adds the job to the rate limited queue
 func (w *worker) SubmitJob(job interface{}) {
+	// in theory, only health check endpoint should send a nil job to test periodically
+	if job == nil {
```
sushrk, line 130: nit: maybe good to check if it is the health check endpoint here.

sushrk (commented):
Few minor comments, please take a look.

pkg/worker/worker.go
haouc: I don't think there us an easy way to check that. We will have to do either of
- mock the parameter to reflecting caller identity. But this will be no difference from current `nil`
- use runtime package to check caller frames. I think this adds unnecessary complexity.

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,89 @@
...
+func setTimeOutFromFlag(timeout time.Duration) {
+	HealthzTimeout = timeout
+}
+
+func (hh *HealthzHandler) AddControllerHealthChecker(name string, controllerCheck healthz.Checker) bool {
+	if _, ok := hh.CheckersMap[name]; ok {
+		return false
+	} else {
+		hh.CheckersMap[name] = controllerCheck
+		return true
```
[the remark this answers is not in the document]
haouc: I think you are right. This function doesn't need a boolean return type.

pkg/healthz/healthz.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func NewHealthzHandler(timeout int) *HealthzHandler {
+	setTimeOutFromFlag(time.Duration(timeout))
+	return &HealthzHandler{
+		CheckersMap: make(map[string]healthz.Checker),
+	}
+}
+
+func setTimeOutFromFlag(timeout time.Duration) {
```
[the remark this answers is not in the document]
haouc: yes, it can.

haouc:
Seems controller-gen has some flakiness. I will take some look at this tool as working on #199

sushrk (approved):
LGTM, thank you.

PR #204, opened 2023-04-24
added a condition and updated config controller to check for prefix de…

jiechen0826:
…legation flag and added two new APIs to pool interface to set warm pool when prefix delegation flag is toggled.

*Issue #, if available:*

*Description of changes:*
- Added a new condition to check for prefix delegation flag in config map
- Updated config controller to reconcile whenever the flag is toggled
- Added two APIs to `pool` interface to set resource config for warm pool when the flag is toggled
- Updated secondary IPv4 address provider to detect when PD flag is enabled, call its warm pool to update capacity to 0

### DIFF
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -32,12 +32,13 @@ import (
 // ConfigMapReconciler reconciles a ConfigMap object
 type ConfigMapReconciler struct {
 	client.Client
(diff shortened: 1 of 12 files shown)


jiechen0826:
Hi @haouc and @jayanthvn, please take a look at the PR and let me know if it should be broken down into smaller ones.

PR #205, opened 2023-04-24
added a new condition to check for prefix delegation flag

jiechen0826:
*Issue #, if available:*

*Description of changes:*
Added a new condition to check for prefix delegation flag in config map `amazon-vpc-cni`, so that resource manager can pass it to resource provider, which will use the condition to initialize resource pools accordingly.

### DIFF
--- a/pkg/condition/conditions.go
+++ b/pkg/condition/conditions.go
@@ -39,6 +39,10 @@ type Conditions interface {
 	// IsWindowsIPAMEnabled to process events only when Windows IPAM is enabled
 	// by the user
 	IsWindowsIPAMEnabled() bool
+
+	// IsWindowsPrefixDelegationEnabled to process events only when Windows Prefix Delegation is enabled
+	IsWindowsPrefixDelegationEnabled() bool
+
 	// IsPodSGPEnabled to process events only when Security Group for Pods feature
 	// is enabled by the user
 	// IsPodSGPEnabled() bool We need to check if SGP is enabled via ConfigMap + Environment variables
@@ -61,6 +65,12 @@ var (
 			Help: "Binary value to indicate whether user has set enable-windows-ipam to true",
 		})
 
+	conditionWindowsPrefixDelegationEnabled = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "windows_prefix_delegation_enabled",
+			Help: "Binary value to indicate whether user has set enable-windows-prefix-delegation to true",
+		})
+
 	prometheusRegistered = false
 )
 
@@ -68,6 +78,7 @@ func prometheusRegister() {
 	if !prometheusRegistered {
 		metrics.Registry.MustRegister(
 			conditionWindowsIPAMEnabled,
+			conditionWindowsPrefixDelegationEnabled,
 		)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/condition/conditions.go
```diff
@@ -106,6 +118,35 @@ func (c *condition) IsWindowsIPAMEnabled() bool {
 	return false
 }
 
+func (c *condition) IsWindowsPrefixDelegationEnabled() bool {
```
haouc, line 121: Where will this function be called? I assume this should be called after the check on IsWindowsTurnedOn passes. If so, seems no reason to check if Windows is on again here.

pkg/condition/conditions.go
```diff
@@ -106,6 +118,35 @@ func (c *condition) IsWindowsIPAMEnabled() bool {
...
+	if c.IsOldVPCControllerDeploymentPresent() {
+		return false
+	}
+
+	// Return false if configmap not present/any errors
+	vpcCniConfigMap, err := c.K8sAPI.GetConfigMap(config.VpcCniConfigMapName, config.KubeSystemNamespace)
+
+	if err == nil && vpcCniConfigMap.Data != nil {
```
haouc, line 129: From design point of view, if users turn on Windows prefix, that should indicate the user is using Windows. I wonder if it is necessary to have two toggles for users:
- EnableWindowsIPAMKey
- EnableWindowsPrefixDelegationKey

Will that work same if `EnableWindowsPrefixDelegationKey` is true, bypass `EnableWindowsIPAMKey` and turn on Windows support?

jiechen0826: You are right. We can make `EnableWindowsPrefixDelegationKey` bypass `EnableWindowsIPAMKey` here. 

If users need to turn off PD but still keep Windows support, they will have to explicitly turn on `EnableWindowsIPAMKey` though.

jiechen0826: @haouc Just wanted to clarify: what happens if users set conflicting flags, e.g. PD=true, IPAM=false. In this case, which flag should override? If following the above logic of PD flag overriding IPAM flag, does it mean IPAM is still enabled even though it is explicitly set to false?

haouc: Yeah, my opinion is if PD is on Windows should be considered as on. This can be called out explicitly in doc that the best practice is turn on IPAM and / or turn on PD. If users only turn on PD, the controller will by default think the user turns on IPAM. 
If you believe requiring 1, turn on IPAM, 2 if wanted, turn on PD. If only turning on PD, this can be considered as a misconfiguration. This should also work to me. 
It is basically UX thing.

pkg/condition/conditions.go
jiechen0826: This will be called by resource providers to know how to initialize their pools and no call is made to `IsWindowsIPAMEnabled()` before this one. So adding here to make sure IPAM is indeed enabled.

pkg/condition/conditions.go
jiechen0826: After getting input from Product, we will keep these two flags as independent for now, and no override behavior here to avoid future confusion. PD flag will only be effective when IPAM is enabled.

haouc (approved):
lgtm

PR #210, opened 2023-04-29
add a new CRD to delegate vpc resource requests

haouc:
*Issue #, if available:*

*Description of changes:*
We want to add a new CRD to delegate communications between this controller and VPC CNI, or other operators from dataplane in the future.

Design considerations
```
* The CRD object will be one-to-one mapping to worker node
    * one static object per node and last through the node lifespan
    * the object maps to VPC resources and EKS managed featured on the mapped node
* Instead of using v1beta1 we use v1alpha1 for easier, more compatible and conventional development iterations
* Adding Spec and Status
    * Spec: providing a description of the characteristics we want this CRD to have
        * Enabled features are listed as feature flags
            * SecurityGroupForPods
            * CustomNetworking
            * ...
    * Status: (Currently is empty) describing the current state of the VPC resources managed through this CRD, supplied and updated by the EKS networking addons.
        * Future potential use cases for VPC resources, for instance, various types of ENIs
            * list of trunk interfaces
            * list of primary interface
            * list of secondary interfaces
            * list of security groups
            * list of IP addresses to support warm target for branch interfaces
* Using name CNINode/CNINodes which are aligned with Kubernetes well adopted API extensions, such as CSI, CRI, and etc.
* Using enum for Spec and Status
* Using feature flag similar concept for Spec items
* metadata ObjectReference will be provided during creation to handle cascade deletion open node deletion events
* Will add Finalizer to control the GC process (design in process)
    * The Finalizer will refer to resources listed in Status 
* VPC RC will be responsible to CREATE/PATCH/DELET the CRD objects
```

Tests:

Created an object of the CRD
```
% k get cnd
NAME              FEATURES
cninode-example   ["SecurityGroupsForPods","CustomNetworking"]
```


jdn5126 (commented):
Definition looks good to me. I'll give others a chance to review before approving

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+package v1beta1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// VPCResourceRequestSpec defines the desired state of VPCResourceRequest
+type VPCResourceRequestSpec struct {
+	RequestTypes  map[string]ResourceType `json:"requestTypes,omitempty"`
+	RequestTarget Target                  `json:"requestTarget,omitempty"`
```
ellistarn, line 23: Consider removing request, to avoid stuttering. e.g. VPCResourceRequest.RequestType vs VPCResourceRequest.Type
```suggestion
	Types  map[string]ResourceType `json:"types,omitempty"`
	Target Target                  `json:"target,omitempty"`
```

config/rbac/role.yaml
```diff
@@ -72,6 +72,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - vpcresources.k8s.aws
```
ellistarn, line 76: Where did we choose this group name?

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Target represents a request's target
+// Target could be pod, node or other Kinds
+type Target struct {
+	Name  string `json:"name,omitempty"`
+	Value string `json:"value,omitempty"`
+}
+
+// Type represents a request's type
+// ResourceType could features, such as SGP, custom networking, etc.
```
ellistarn, line 35: Is it possible to strongly type these, or potentially use Enums? What are all of the options here.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="Target",type=string,JSONPath=`.spec.requestTarget`,description="The target of the request CRD object"
+// +kubebuilder:printcolumn:name="Types",type=string,JSONPath=`.spec.requestTypes`,description="The type of the request CRD object"
+// +kubebuilder:printcolumn:name="Message",type=string,JSONPath=`.spec.message`,description="The message of the request CRD object"
+// +kubebuilder:resource:shortName=vrr
+
+// Custom Resource Definition for Request to VPC resource controller
+type VPCResourceRequest struct {
```
ellistarn, line 48: Thoughts on "VPCResource"? Request is implicit -- essentially all kubernetes apis are requesting that something happens.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// VPCResourceRequestSpec defines the desired state of VPCResourceRequest
+type VPCResourceRequestSpec struct {
+	RequestTypes  map[string]ResourceType `json:"requestTypes,omitempty"`
+	RequestTarget Target                  `json:"requestTarget,omitempty"`
+	Message       string                  `json:"message,omitempty"`
```
ellistarn, line 24: What is message used for?

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+// Type represents a request's type
+// ResourceType could features, such as SGP, custom networking, etc.
+type ResourceType struct {
+	Name  string `json:"name,omitempty"`
+	Value string `json:"value,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="Target",type=string,JSONPath=`.spec.requestTarget`,description="The target of the request CRD object"
+// +kubebuilder:printcolumn:name="Types",type=string,JSONPath=`.spec.requestTypes`,description="The type of the request CRD object"
```
ellistarn, line 43: Will this dump a bunch of json on kubectl get? How big can this get?

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: Makes sense, Ellis. Will update the naming.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: Currently, the target is just the host name. The types are features shared with the CRD object. They are not big yet. But yes, they could be big in the future.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: This was suggested to indicate something requested to VPC resource controller to manage. Probably we should be specific by using "VPCResourceController"?

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: We want to make this extensible for future cases, for example interfaces, IP addresses, and probably other networking resources. Making this strongly typed could add some limitations on the future use. @jayanthvn do you have any recommendations on this? Thanks

config/samples/vpcresourcerequest.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: VPCResourceRequest
+metadata:
+  name: ip-192-168-244-122.us-west-2.compute.internal
+  namespace: kube-system
+spec:
+  requestTarget: 
+    name: "node"
+    value: "ip-192-168-244-122.us-west-2.compute.internal"
+  requestTypes:
+    trunk-eni:
```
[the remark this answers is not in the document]
haouc: Will update this. Thanks for pointing this out.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
ellistarn: An enum should enable you to explicitly expand this over time. It also has an effect of making the interface self-documenting.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
ellistarn: How is it used today?

config/rbac/role.yaml
ellistarn: ack. Similar to karpenter.k8s.aws. Might be a little cleaner as vpc.k8s.aws, but that ship has sailed, I think.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: Currently the usage is very limited to signal if a trunk interface can be attached by CNI labels the node.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
haouc: Makes sense. We should do that.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
```diff
@@ -0,0 +1,66 @@
...
+// VPCResourceRequestSpec defines the desired state of VPCResourceRequest
+type VPCResourceRequestSpec struct {
+	RequestTypes  map[string]ResourceType `json:"requestTypes,omitempty"`
+	RequestTarget Target                  `json:"requestTarget,omitempty"`
+	Message       string                  `json:"message,omitempty"`
+}
+
+// Target represents a request's target
+// Target could be pod, node or other Kinds
+type Target struct {
```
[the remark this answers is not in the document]
haouc: Ack, makes better sense to use ObjectRef here.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
ellistarn: Can you elaborate a little bit on this? Are there some examples I can look at?

config/samples/vpcresourcerequest.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: vpcresources.k8s.aws/v1beta1
```
ellistarn, line 1: Consider starting the API as an alpha.
```suggestion
apiVersion: vpcresources.k8s.aws/v1alpha1
```

apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -0,0 +1,57 @@
...
+	CustomNetworking      FeatureName = "CustomNetworking"
+)
+
+// Important: Run "make" to regenerate code after modifying this file
+// CNINodeSpec defines the desired state of CNINode
+type CNINodeSpec struct {
+	Features []FeatureName `json:"features,omitempty"`
+}
+
+type VPCResource string
```
ellistarn, line 34:
```suggestion
```

apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="Features",type=string,JSONPath=`.spec.features`,description="The features delegated to VPC resource controller"
+// +kubebuilder:resource:shortName=cnd
```
ellistarn, line 38:
```suggestion
// +kubebuilder:resource:shortName=cni
```

apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+// +kubebuilder:object:root=true
+type CNINode struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              CNINodeSpec `json:"spec,omitempty"`
```
ellistarn, line 44: Consider including a status, even if it's empty for now.

ellistarn (approved):
Nice! A couple of minor optional comments.

ellistarn (approved):
Nice! A couple of minor optional comments.

ellistarn (dismissed):
Nice! A couple of minor optional comments.

apis/vpcresources/v1alpha1/cninode_types.go
haouc: Yeah, thought about `cni` as well, but it is definitely a conflict with well known `CNI`. Waiting for comments.

apis/vpcresources/v1alpha1/cninode_types.go
haouc: Make sense to force the pattern from day 1.

apis/vpcresources/v1alpha1/cninode_types.go
haouc: Missed this when remove the block of const. Will update.

apis/vpcresources/v1beta1/vpcresourcerequest_types.go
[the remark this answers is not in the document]
haouc: Based on offline discussion, the design has been updated with fields we only need.

apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -0,0 +1,61 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FeatureName is a type of feature name supported by EKS networking. It can be Security Group for Pods, custom networking, or others
```
jayanthvn, line 20: Nit: EKS networking -> AWS VPC CNI

apis/vpcresources/v1alpha1/cninode_types.go
jayanthvn: maybe `awsvpccni` or just `vpccni`? Btw, was there any other options instead of CNINode or CNINodes? Also this is specific to `aws-vpc-cni` only and not `amazon-vpc-cni-plugins` right? Is there any plans to extend it to windows node? One option maybe windows nodes can utilize this for enabling/disabling PD?

apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	SecurityGroupsForPods FeatureName = "SecurityGroupsForPods"
+	CustomNetworking      FeatureName = "CustomNetworking"
+)
+
+// Important: Run "make" to regenerate code after modifying this file
+// CNINodeSpec defines the desired state of CNINode
+type CNINodeSpec struct {
+	Features []FeatureName `json:"features,omitempty"`
```
M00nF1sh, line 31: shall we have list of features or  something like
```
spec:
 securityGroupsForPods:
    enabled: true
 customNetworking:
    enabled: false
```
the benefit i can see is we can optionally have more configurations for each feature when necessary.

M00nF1sh (approved):

apis/vpcresources/v1alpha1/cninode_types.go
haouc: yes, we discussed this as well. For this version as alpha1, we intend to onboard the two features and can reference them as feature flags then the boolean value is not a must. But I agree we can definitely add support on values when we onboard more features or combined configurations which need track their status.

haouc: Synced offline, we can accept the current design for now and keep the door open.

apis/vpcresources/v1alpha1/cninode_types.go
haouc: I am closing this PR for now. But we will keep the option open to other meaningful short name.

PR #212, opened 2023-05-01
Added new resource type and updated pod controller to handle prefix-deconstructed IPs

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Added new resource type for prefix-deconstructed IPs
- Updated resource manager to initialize this new resource type
- Updated pod controller reconciliation to use corresponding resource handler during pod creation and deletion events


jiechen0826:
Hi @haouc @jayanthvn, please take a look at my PRs. Thanks a lot!

jiechen0826:
> @jiechen0826 can you look at the unit test failures here as well?

Yea the reason the unit tests failed is I tried to make the PR small so it didn't contain all the necessary changes for it to compile and build. If I add the other files it might make the PR too big for review. After a few PRs into the feature branch, they will be passing the unit tests again.

controllers/core/pod_controller.go
```diff
@@ -189,3 +197,32 @@ func (r *PodReconciler) SetupWithManager(ctx context.Context, manager ctrl.Manag
...
+	if isDeletionEvent {
+		resourceID, present := pod.Annotations[resourceName]
+		if !present {
+			r.Log.Error(fmt.Errorf("resource ID was not found in annotation"),
+				"resource", resourceName, "is delete event", isDeletionEvent)
+			return resourceName
+		}
+		// Check prefix provider to see if it's a prefix deconstructed IP
+		providerMap := r.ResourceManager.GetResourceProviders()
+		prefixProvider := providerMap[config.ResourceNameIPAddressFromPrefix]
```
haouc, line 214: The pod controller is async reconciler. I am not sure if there is safe to read from the map without locking. But adding locks in reconciling need cautions. Can you check you other change in Manager and see if can read from there and protected by a lock?

jiechen0826: Actually I can't do it in resource manage since it doesn't have a lock there. It's always locked by the caller

jiechen0826: Pushed a new commit to add the lock for resource map in resource manager, and a new API to get resource provider by name.

pkg/resource/manager.go
```diff
@@ -103,14 +113,31 @@ func NewResourceManager(ctx context.Context, resourceNames []string, wrapper api
 }
 
 func (m *Manager) GetResourceProviders() map[string]provider.ResourceProvider {
+	m.lock.RLock()
```
haouc, line 116: I guess you don't need lock here, right? This function creates the map and the map will be maintained and shared with goroutines. Seems no reason to lock the map creation. Can you double check if this is called once when creating manager? Also this is a `Set` function instead of `Get`. Can you rename it? Thanks.

jiechen0826: I think map creation happens here: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/0c6381a1ef46b1315dbe3b88d453115a1623e88f/pkg/resource/manager.go#L107-L110

Here is making a copy of the original map and returning the copy, so I guess it still counts as a `Get`?

haouc: hmm, I see. May I ask the reason to make a copy there? I don't see an update/modification here? Since we have small PRs, I am not able to check what the function is for.

jiechen0826: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/0e53849b687af632c490bf3c1b3cf3a69d7abcaa/pkg/resource/manager.go#L125-L131

I'm not sure why the original author made a copy of the map here, but I'm guessing the original map was not returned in case it could be modified by the caller?

controllers/core/pod_controller.go
jiechen0826: Removing the newly added lock here as discussed offline.

pkg/resource/manager.go
jiechen0826: Realized this is not a copy of `resource` which contains both provider and handler, it's a new map from resource name to resource provider.

PR #213, opened 2023-05-01
Updated IP provider to support PD feature toggles

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Added support for toggling between two IP allocation modes. If toggling from secondary IP to PD, secondary IP provider will drain its pool. If toggling from PD to secondary IP mode, IP provider will set its pool to active state by passing in a valid default resource config. 
- Updated capacity calculation to take into account existing IPv4 prefixes.

### DIFF
--- a/pkg/provider/ip/provider.go
+++ b/pkg/provider/ip/provider.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2"
-	ec2api "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
@@ -266,7 +265,7 @@ func (p *ipv4Provider) CreatePrivateIPv4AndUpdatePool(job *worker.WarmPoolJob) {
 		return
 	}
 	didSucceed := true
-	ips, err := instanceResource.eniManager.CreateIPV4Resource(job.ResourceCount, ec2api.ResourceTypeIPv4Address, p.apiWrapper.EC2API, p.log)
+	ips, err := instanceResource.eniManager.CreateIPV4Resource(job.ResourceCount, config.ResourceTypeIPv4Address, p.apiWrapper.EC2API, p.log)
 	if err != nil {
 		p.log.Error(err, "failed to create all/some of the IPv4 addresses", "created ips", ips)
 		didSucceed = false
@@ -301,7 +300,7 @@ func (p *ipv4Provider) DeletePrivateIPv4AndUpdatePool(job *worker.WarmPoolJob) {
 		return
 	}
 	didSucceed := true
-	failedIPs, err := instanceResource.eniManager.DeleteIPV4Resource(job.Resources, ec2api.ResourceTypeIPv4Address, p.apiWrapper.EC2API, p.log)
+	failedIPs, err := instanceResource.eniManager.DeleteIPV4Resource(job.Resources, config.ResourceTypeIPv4Address, p.apiWrapper.EC2API, p.log)
 	if err != nil {
 		p.log.Error(err, "failed to delete all/some of the IPv4 addresses", "failed ips", failedIPs)
 		didSucceed = false
--- a/pkg/provider/ip/provider.go
+++ b/pkg/provider/ip/provider.go
@@ -19,7 +19,9 @@ import (
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2"
+	ec2api "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/ec2/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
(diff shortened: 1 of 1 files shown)


jiechen0826:
Hi @haouc @jayanthvn, please take a look at my PRs. Thanks a lot!

jdn5126:
@jiechen0826 I see a lot of unit test failures here. Did you run `make test` before submitting?

jiechen0826:
> @jiechen0826 I see a lot of unit test failures here. Did you run `make test` before submitting?

Yes I did. The unit tests are failing. The reason the unit tests failed is I tried to make the PR small so it didn't contain all the necessary changes for it to compile and build. If I add the other files which also contain changes, it might make the PR too big for review. After a few PRs into the feature branch, they will be passing the unit tests again.

pkg/provider/ip/provider.go
```diff
@@ -120,11 +154,52 @@ func (p *ipv4Provider) DeInitResource(instance ec2.EC2Instance) error {
 
 // UpdateResourceCapacity updates the resource capacity based on the type of instance
 func (p *ipv4Provider) UpdateResourceCapacity(instance ec2.EC2Instance) error {
+	resourceProviderAndPool, isPresent := p.getInstanceProviderAndPool(instance.Name())
+	if !isPresent {
+		p.log.Error(nil, "cannot find the instance provider and pool form the cache", "node-name", instance.Name())
+		return nil
+	}
+
+	resourceProviderAndPool.lock.Lock()
```
haouc, line 163: Does this lock guard `isPrevPDEnabled`?

pkg/provider/ip/provider.go
```diff
@@ -66,35 +76,59 @@ func (p *ipv4Provider) InitResource(instance ec2.EC2Instance) error {
...
+		presentIPSet[ip] = struct{}{}
+	}
+
 	pods, err := p.apiWrapper.PodAPI.GetRunningPodsOnNode(nodeName)
 	if err != nil {
 		return err
 	}
 
-	podToResourceMap := map[string]string{}
+	podToResourceMap := map[string]pool.Resource{}
```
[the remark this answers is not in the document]
jiechen0826: It is an IP address.

```go
// Resource represents a secondary IPv4 address or a prefix-deconstructed IPv4 address, uniquely identified by GroupID and ResourceID
type Resource struct {
	// could be IPv4 address or IPv4 prefix
	GroupID string
	// IPv4 address
	ResourceID string
}
```

pkg/provider/ip/provider.go
jiechen0826: > Does this lock guard `isPrevPDEnabled`?

Yes, this lock guards `isPrevPDEnabled` and `capacity`.

pkg/provider/ip/provider.go
```diff
@@ -120,11 +154,52 @@ func (p *ipv4Provider) DeInitResource(instance ec2.EC2Instance) error {
...
+	}
+	job := resourceProviderAndPool.resourcePool.SetToActive(ipv4ResourceConfig.WarmPoolConfig)
+	p.SubmitAsyncJob(job)
+
 	instanceType := instance.Type()
 	instanceName := instance.Name()
 	os := instance.Os()
 
-	capacity := getCapacity(instanceType, os)
+	capacity := resourceProviderAndPool.capacity
```
jiechen0826, line 202: @haouc `resourceProviderAndPool.capacity` is being read here so the lock is required as well.

pkg/provider/ip/provider.go
```diff
@@ -42,19 +44,27 @@ type ipv4Provider struct {
...
+	capacity int
+	// isPrevPDEnabled stores whether PD was enabled previously
+	isPrevPDEnabled bool
 }
 
 func NewIPv4Provider(log logr.Logger, apiWrapper api.Wrapper,
 	workerPool worker.Worker, resourceConfig config.ResourceConfig) provider.ResourceProvider {
 	return &ipv4Provider{
-		instanceProviderAndPool: make(map[string]ResourceProviderAndPool),
+		instanceProviderAndPool: make(map[string]*ResourceProviderAndPool),
```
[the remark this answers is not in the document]
haouc: I am not sure why. But we can look at this later when we have another PR including the related codes.

pkg/provider/ip/provider.go
haouc: Is that possible to have RLocks in the calling functions instead of the caller? For example, in functions like GetCapacity()?

jiechen0826: I see what you are saying. The reason I added here is the lock was already acquired at the beginning of this function and deferred releasing since it was needed for updating `isPDEnabled` so I thought not to acquire it again. This is all part of `UpdateResourceCapacity` function, but github added a few automatic checks that broke down the block into fragments which made it more confusing for reviewing. Sorry about that!

haouc: We can have the lock like this for now. When we have the entire package ready (the bigger PR), we can revisit the locks. We definitely need to avoid recall lock!

pkg/provider/ip/provider.go
```diff
@@ -66,35 +75,59 @@ func (p *ipv4Provider) InitResource(instance ec2.EC2Instance) error {
...
-		warmResources, instance.Name(), nodeCapacity)
+	nodeCapacity := getCapacity(instance.Type(), instance.Os()) - len(presentPrefixes)
 
-	p.putInstanceProviderAndPool(nodeName, resourcePool, eniManager)
+	// Set warm pool config to 0 if PD is enabled
+	ipV4WarmPoolConfig := p.config
+	isPDEnabled := p.conditions.IsWindowsPrefixDelegationEnabled()
+	if isPDEnabled {
+		ipV4WarmPoolConfig = &config.WarmPoolConfig{
+			DesiredSize:  0,
```
haouc, line 120: DesiredSize is 0? Is this intended?

jiechen0826: Ahh yes, this is intended because for secondary IP provider, once it sees PD enabled, it sets its own secondary IP pool to `desiredSize` of 0, effectively draining the pool.

haouc (approved):
lgtm

PR #217, opened 2023-05-05
updated vpc limits to include fields for hypervisor type and bare met…

jiechen0826:
…al status

*Issue #, if available:*

*Description of changes:*
Added two new fields `Hypervisor` and `IsBareMetal` to vpc limits. This can be used to check if prefix delegation is supported.

Hypervisor field can take on values of "xen", "nitro" or "" if it's bare metal. 


haouc (approved):
lgtm, thanks for updating this file.

PR #218, opened 2023-05-05
enable node events when instance type is not supported

haouc:
*Issue #, if available:*

*Description of changes:*
Since this controller runs in EKS control plane, if the instance type is not supported, the error message is not available for users. To make users' debugging easier and better visibility, we should expose some of critical failures that can be mitigated from user side. Events are common use for this purpose. This change propose sending node event when 
- new instance types are not included to support for this controller (Linux SGP feature or Windows)
- instance types are in the supported types list but doesn't support trunk interface (SGP feature)

Other changes:
- refactoring how passing client to node/instance. Since node is k8s object and also carry AWS instance, I think it is making sense to add k8s/ec2 clients into node struct instead of need pass them to functions when calling node resources.
- updating unit tests

Tests:
The unsupported and mocked missing instance types were used to test the events

case 1: the instance type doesn't support SGP/trunk interface
```
Warning  Unsupported              3m20s (x6 over 3m41s)  vpc-resource-controller  The instance type t3.medium is not supported for trunk interface (Security Group for Pods)
```

case 2: the (mocked) instance is new and hasn't been added into the supporting list
```
Warning  Unsupported              3m27s (x5 over 3m40s)  vpc-resource-controller  The instance type m5.large is not supported yet by the vpc resource controller
```

### DIFF
--- a/pkg/node/node_test.go
+++ b/pkg/node/node_test.go
@@ -20,14 +20,17 @@ import (
 
 	mock_ec2 "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/aws/ec2"
 	mock_api "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/aws/ec2/api"
+	mock_k8s "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
 	mock_provider "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/provider"
 	mock_resource "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/resource"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
@@ -49,6 +52,7 @@ type Mocks struct {
 	MockResourceManager *mock_resource.MockResourceManager
 	MockInstance        *mock_ec2.MockEC2Instance
 	MockEC2API          *mock_api.MockEC2APIHelper
+	MockK8sAPI          *mock_k8s.MockK8sWrapper
 	NodeWithMock        node
 }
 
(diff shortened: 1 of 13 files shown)


pkg/node/node_test.go
```diff
@@ -101,19 +110,19 @@ func TestNode_InitResources(t *testing.T) {
 	defer ctrl.Finish()
 
 	mock := NewMock(ctrl, 1)
-
+	fmt.Printf("Is EC2 API nil? %t\n", mock.MockEC2API == nil)
```
jdn5126, line 113: nit: I assume this was just for local testing and can be removed now

pkg/provider/ip/eni/eni_test.go
```diff
@@ -158,6 +159,23 @@ func TestEni_InitResources_Error(t *testing.T) {
 	assert.Error(t, mockError, err)
 }
 
+// TestEni_InitResources_Unspuuported_Type_Error tests that error is returned if the instance type is not supported
+func TestEni_InitResources_Unspuuported_Type_Error(t *testing.T) {
```
jdn5126, line 163: nit: "Unspuuported" -> "Unsupported"

pkg/provider/ip/provider.go
```diff
@@ -75,6 +78,11 @@ func (p *ipv4Provider) InitResource(instance ec2.EC2Instance) error {
 	eniManager := eni.NewENIManager(instance)
 	presentIPs, err := eniManager.InitResources(p.apiWrapper.EC2API)
 	if err != nil {
+		if errors.Is(err, utils.ErrNotFound) {
+			msg := fmt.Sprintf("The instance type %s is not supported for Windows yet", instance.Type())
```
jdn5126, line 82: nit: I think you can drop the "yet" in the warning. That implies it will be supported some day, which it may not

pkg/utils/events.go
```diff
@@ -0,0 +1,21 @@
...
+	unsupportedInstanceTypeReason = "Unsupported"
+)
+
+func SendNodeEvent(client k8s.K8sWrapper, nodeName, reason, msg, eventType string, logger logr.Logger) {
+	if node, err := client.GetNode(nodeName); err == nil {
+		// set UID to node name for kubelet filter the event to node description
+		node.SetUID(types.UID(nodeName))
+		client.BroadcastEvent(node, unsupportedInstanceTypeReason, msg, eventType)
+	} else {
+		logger.Error(err, "had an error to get the node for sending unsupported event", "Node", nodeName)
```
jdn5126, line 19: nit: does this log the actual error from the GET call?

jdn5126 (commented):
Changes LGTM, just a few nits and then I can approve

pkg/node/node_test.go
haouc: Thanks for catching this. Yes, it was.

pkg/provider/ip/eni/eni_test.go
haouc: Thanks, fixed the typo.

pkg/provider/ip/provider.go
haouc: Makes sense.

pkg/utils/events.go
haouc: Do you mean `logger.Error()`? Yes, this logs the not nil error from GetNode() API call.

jdn5126 (approved):

PR #220, opened 2023-05-08
updated warm pool to support resource group for secondary ip and pref…

jiechen0826:
…ix-ip

*Issue #, if available:*

*Description of changes:*
- Added a struct for `Resource` which either represents a secondary IP address where group ID is the same as resource ID, or represents a prefix-deconstructed IP address where group ID is a `/28` prefix, and resource ID is a `/32` IPv4 address.
- Updated `usedResources` to be a map from owner id to `Resource` struct
- Updated  `warmResources` to be a map from resource group ID to a list of `Resource` that belong to that group. For secondary IP resource group, each key is a secondary IPv4 address, and the value contains 1 `Resource`. For prefix IP resource group, each key is a prefix, and the value is the list of IPv4 addresses deconstructed from that prefix.
- Added API `IsManagedResource` so that pod controller can call this and check if a pod's IP is managed by a IP pool or a PD pool.
- Added APIs `SetToDraining` and `SetToActive` to handle toggle situations between secondary IP mode and prefix delegation mode.
- When assigning IPs from secondary IP pool, it picks a random IP from the warm pool, whereas for prefix IP pool, it chooses the IP from the resource group (the prefix) with fewest elements to avoid prefix starvation.
- `getPDDeviation` calculates number of prefixes to maintain in the warm pool to satisfy `WarmIPTarget`, `MinimumIPTarget` and `WarmPrefixTarget` requirements.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -15,13 +15,14 @@ package pool
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker"
-
 	"github.com/go-logr/logr"
 )
 
@@ -33,6 +34,7 @@ var (
 	ErrResourceAlreadyAssigned    = fmt.Errorf("resource is already assigned to the requestor")
 	ErrResourceDoesntExist        = fmt.Errorf("requested resource doesn't exist in used pool")
 	ErrIncorrectResourceOwner     = fmt.Errorf("resource doesn't belong to the requestor")
+	NumIPv4AddrPerPrefix          = 16
 )
 
 type Pool interface {
@@ -44,6 +46,7 @@ type Pool interface {
 	ReconcilePool() *worker.WarmPoolJob
 	ProcessCoolDownQueue() bool
 	Introspect() IntrospectResponse
+	IsManagedResource(resourceID string) bool
 }
 
 type pool struct {
@@ -56,9 +59,9 @@ type pool struct {
 	// lock to concurrently make modification to the poll resources
 	lock sync.RWMutex // following resources are guarded by the lock
 	// usedResources is the key value pair of the owner id to the resource id
-	usedResources map[string]string
-	// warmResources is the list of free resources available to be allocated to the pods
-	warmResources []string
+	usedResources map[string]Resource
+	// warmResources is the map of group id to a list of free resources available to be allocated to the pods
+	warmResources map[string][]Resource // use map  so that assign/free is fast, despite resync/reconsile always slow
 	// coolDownQueue is the resources that sit in the queue for the cool down period
 	coolDownQueue []CoolDownResource
 	// pendingCreate represents the number of resources being created asynchronously
@@ -70,24 +73,34 @@ type pool struct {
 	// reSyncRequired is set if the upstream and pool are possibly out of sync due to
 	// errors in creating/deleting resources
 	reSyncRequired bool
+	// isPDPool indicates whether the pool is for prefix IP provider or secondary IP provider
+	isPDPool bool
+}
+
+// Resource represents a secondary IPv4 address or a prefix-deconstructed IPv4 address, uniquely identified by GroupID and ResourceID
+type Resource struct {
+	// could be IPv4 address or IPv4 prefix
+	GroupID string
+	// IPv4 address
+	ResourceID string
 }
 
 type CoolDownResource struct {
 	// ResourceID is the unique ID of the resource
-	ResourceID string
+	Resource Resource
 	// DeletionTimestamp is the time when the owner of the resource was deleted
 	DeletionTimestamp time.Time
 }
 
 // IntrospectResponse is the pool state returned to the introspect API
 type IntrospectResponse struct {
-	UsedResources    map[string]string
-	WarmResources    []string
+	UsedResources    map[string]Resource
+	WarmResources    map[string][]Resource
 	CoolingResources []CoolDownResource
 }
 
-func NewResourcePool(log logr.Logger, poolConfig *config.WarmPoolConfig, usedResources map[string]string,
-	warmResources []string, nodeName string, capacity int) Pool {
+func NewResourcePool(log logr.Logger, poolConfig *config.WarmPoolConfig, usedResources map[string]Resource,
+	warmResources map[string][]Resource, nodeName string, capacity int, isPDPool bool) Pool {
 	pool := &pool{
 		log:            log,
 		warmPoolConfig: poolConfig,
@@ -95,6 +108,7 @@ func NewResourcePool(log logr.Logger, poolConfig *config.WarmPoolConfig, usedRes
 		warmResources:  warmResources,
 		capacity:       capacity,
 		nodeName:       nodeName,
+		isPDPool:       isPDPool,
 	}
 	return pool
 }
@@ -104,7 +118,7 @@ func NewResourcePool(log logr.Logger, poolConfig *config.WarmPoolConfig, usedRes
 // upstream has additional resources which are not present locally, these resources are added
 // to the warm pool. During ReSync all Create/Delete operations on the Pool should be halted
 // but Assign/Free on the Pool can be allowed.
-func (p *pool) ReSync(upstreamResource []string) {
+func (p *pool) ReSync(upstreamResourceGroupIDs []string) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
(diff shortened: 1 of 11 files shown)


pkg/pool/pool.go
```diff
@@ -359,26 +441,45 @@ func (p *pool) ReconcilePool() *worker.WarmPoolJob {
...
-			resourceToDelete = append(resourceToDelete, p.warmResources[i])
+		numToDelete := deviation
+		for numToDelete > 0 {
+			numFreeGroups, freeResourceGroups := p.findFreeGroup()
+			if numFreeGroups == 0 {
+				log.Info("no warm resources to delete", "deviation", deviation, "numToDelete", numToDelete)
+				break
+			}
+
+			// Remove resources to be deleted form the warm pool
```
haouc, line 462: nit: from

pkg/pool/pool.go
```diff
@@ -359,26 +441,45 @@ func (p *pool) ReconcilePool() *worker.WarmPoolJob {
...
+				delete(p.warmResources, groupID)
+			}
 		}
 
-		// Remove resources to be deleted form the warm pool
-		p.warmResources = p.warmResources[:len(p.warmResources)-deviation]
 		// Increment pending to the number of resource being deleted, once successfully deleted the count can be decremented
-		p.pendingDelete += deviation
-		// Submit the job to delete resources
+		p.pendingDelete += deviation - numToDelete
```
haouc, line 475: Shouldn't `numToDelete` be 0 here?

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+		isWarmPrefixTargetDefined = true
+	}
+
+	numExistingWarmResources := numResourcesFromMap(p.warmResources)
+
+	// if neither WarmIPTarget nor MinIPTarget defined but WarmPrefixTarget is defined, calculate deviation in number of prefixes first
+	if !isWarmIPTargetDefined && !isMinIPTargetDefined && isWarmPrefixTargetDefined {
+		// a prefix is free if it has all possible IPs available for allocation, e.g. for /28 prefix it means 16 IPs
+		numFreePrefix, _ := p.findFreeGroup()
+		deviationPrefix = p.warmPoolConfig.WarmPrefixTarget - numFreePrefix - p.pendingCreate/NumIPv4AddrPerPrefix
```
haouc, line 593: Is this `p.pendingCreate/NumIPv4AddrPerPrefix` accurate? Do you need to add 1?

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+			p.warmResources[groupID] = resources[1:]
+		}
+		break
+	}
+
+	return resource
+}
+
+// getPDDeviation returns the deviation in number of IPv4 addresses when PD is enabled, taking into account of all PD configuration params
+func (p *pool) getPDDeviation() int {
```
haouc, line 573: Please add unit test. It is hard to eye check for all these calculations.

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+			minCount = len(resources)
+			group = groupID
+		}
+	}
+
+	return
+}
+
+// assignResourceFromMinGroup assigns a resource from the resource group that has the lowest number of resources
+func (p *pool) assignResourceFromMinGroup() Resource {
```
haouc, line 536: Please add unit test

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+		p.warmResources[groupID] = nil
+	} else {
+		p.warmResources[groupID] = p.warmResources[groupID][1:]
+	}
+
+	return resource
+}
+
+// assignResourceFromAnyGroup assigns a resource from any resource group with at least one resource available
+func (p *pool) assignResourceFromAnyGroup() Resource {
```
haouc, line 553: Also we need unit test on this function.

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+	for _, resource := range p.usedResources {
+		if resource.ResourceID == resourceID {
+			return true
+		}
+	}
+	return false
+}
+
+// findMinGroup finds the resource group with the lowest number of resources, and returns the group id along with its count of resources
+func findMinGroup(resourceGroups map[string][]Resource) (group string, minCount int) {
```
haouc, line 519: Need uni tests on this as well.

pkg/pool/pool.go
```diff
@@ -70,31 +73,42 @@ type pool struct {
...
+	// isPDPool indicates whether the pool is for prefix IP provider or secondary IP provider
+	isPDPool bool
+}
+
+// Resource represents a secondary IPv4 address or a prefix-deconstructed IPv4 address, uniquely identified by GroupID and ResourceID
+type Resource struct {
+	// could be IPv4 address or IPv4 prefix
+	GroupID string
+	// IPv4 address
+	ResourceID string
```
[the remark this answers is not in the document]
jiechen0826: Since `ResourceID` was used everywhere else (e.g. in resource handler where pods are created/deleted), I wonder if it's better to keep it consistent. Maybe in the future `ResourceID` can be extended to mean other network resources. What do you think?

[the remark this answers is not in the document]
haouc: Makes sense. It sounds good to me. Thanks

pkg/pool/pool.go
jiechen0826: That's a good point. Inside the function `getPDDeviation()`, `p.pendingCreate` is always a multiplier of 16 for PD pool. But I will change it to ceiling division just in case.

pkg/pool/pool.go
jiechen0826: Yea you are right here. Updated accordingly.

pkg/utils/math.go
```diff
@@ -50,3 +50,7 @@ func MaxOf[T constraints.Ordered](vars ...T) T {
 	}
 	return result
 }
+
```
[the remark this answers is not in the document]
jiechen0826: Added comments and unit tests.

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+	networkAddr := addr[0] + "." + addr[1] + "." + addr[2] + "."
+
+	mask, err := strconv.Atoi(prefix[index+1:])
+	if err != nil {
+		return nil, err
+	}
+	numOfAddresses := (32 - mask) * 4
+
+	for i := 0; i < numOfAddresses; i++ {
+		hostAddr, err := strconv.Atoi(addr[3])
```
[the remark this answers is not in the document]
jiechen0826: Thanks for catching it. I added a condition to check for the length of the slice first.

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+	count := 0
+	for _, resources := range resourceGroups {
+		count += len(resources)
+	}
+
+	return count
+}
+
+// DeconstructIPsFromPrefix deconstructs a /28 IPv4 prefix into a list of /32 IPv4 addresses
+func DeconstructIPsFromPrefix(prefix string) ([]string, error) {
```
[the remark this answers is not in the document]
jiechen0826: I moved it to`utils` package and added unit tests there.

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+		if resources != nil && len(resources) == numResourcePerGroup {
+			freeGroupIDs = append(freeGroupIDs, groupID)
+			numFreeGroups++
+		}
+	}
+	return
+}
+
+// resources in a, but not in b
+func differenceResourceList(a, b []Resource) (diff []Resource) {
```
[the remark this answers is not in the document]
jiechen0826: Good idea, done!

pkg/pool/pool.go
```diff
@@ -401,3 +502,212 @@ func (p *pool) Introspect() IntrospectResponse {
...
+	if p.isPDPool {
+		// for IPv4 prefix pool, each resource group is a /28 prefix, which contains 16 resources, i.e. IPv4 addresses
+		numResourcePerGroup = NumIPv4AddrPerPrefix
+	}
+
+	// find resource groups that have all possible resources free to allocate/release
+	for groupID, resources := range p.warmResources {
+		if resources != nil && len(resources) == numResourcePerGroup {
+			freeGroupIDs = append(freeGroupIDs, groupID)
+			numFreeGroups++
```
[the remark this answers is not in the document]
jiechen0826: That's a good point. Updated it return only one value.

pkg/pool/pool.go
```diff
@@ -33,6 +34,7 @@ var (
 	ErrResourceAlreadyAssigned    = fmt.Errorf("resource is already assigned to the requestor")
 	ErrResourceDoesntExist        = fmt.Errorf("requested resource doesn't exist in used pool")
 	ErrIncorrectResourceOwner     = fmt.Errorf("resource doesn't belong to the requestor")
+	NumIPv4AddrPerPrefix          = 16
```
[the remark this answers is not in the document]
jiechen0826: Updated

pkg/pool/pool.go
jiechen0826: Added unit tests.

pkg/pool/pool.go
jiechen0826: Added unit tests.

pkg/pool/pool.go
jiechen0826: Added unit tests.

pkg/pool/pool.go
jiechen0826: Added unit tests.

pkg/pool/pool.go
```diff
@@ -496,17 +508,27 @@ func (p *pool) Introspect() IntrospectResponse {
 		usedResources[k] = v
 	}
 
+	warmResources := make(map[string][]Resource)
```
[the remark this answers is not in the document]
jiechen0826: I was thinking along the line of the lock being only accessible within the pool object. Callers who will be using the map for future operations are presumably not in the pool object and therefore do not have access to `p.RLock()`. A hypothetical caller can obtain the original map from the function, and then proceed to modify it without a lock.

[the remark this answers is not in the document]
haouc: hmm, this sounds contradict to single source principle if callers are relying on the accuracy of the map. The map can become out-dated, can't it?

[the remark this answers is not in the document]
jiechen0826: The caller is still getting the accurate snapshot of the map given that point in time. The map can be out-dated only if it's being written to right after the function returns, and in that case caller will get the updated `IntrospectResponse` by calling the function again.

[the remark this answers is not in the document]
haouc: This is an interface for auditing purpose. Outdated data is fine.

haouc (approved):
lgtm, thanks.

PR #223, opened 2023-05-17
added prefix provider and support for PD configuration params

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Added logic to load PD configuration params (`warm-ip-target`, `minimum-ip-target`, `warm-prefix-target`) from vpc cni config map + unit tests
- Added prefix provider to handle job requests for prefixes + unit tests
- Minor fixes on secondary IP provider and warm pool

### DIFF
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -94,7 +94,7 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	// Check if configurations for Windows prefix delegation have changed
 	var isPDConfigUpdated bool
-	warmIPTarget, minIPTarget, warmPrefixTarget := config.ParseWinPDTargets(configmap)
+	warmIPTarget, minIPTarget, warmPrefixTarget := config.ParseWinPDTargets(r.Log, configmap)
 	if r.curWinPDWarmIPTarget != warmIPTarget || r.curWinPDMinIPTarget != minIPTarget || r.curWinPDWarmPrefixTarget != warmPrefixTarget {
 		r.curWinPDWarmIPTarget = warmIPTarget
 		r.curWinPDMinIPTarget = minIPTarget
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -21,12 +21,14 @@ import (
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node/manager"
+
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 )
 
 // ConfigMapReconciler reconciles a ConfigMap object
@@ -39,6 +41,9 @@ type ConfigMapReconciler struct {
 	Condition                         condition.Conditions
 	curWinIPAMEnabledCond             bool
 	curWinPrefixDelegationEnabledCond bool
+	curWinPDWarmIPTarget              int
+	curWinPDMinIPTarget               int
+	curWinPDWarmPrefixTarget          int
 }
 
 //+kubebuilder:rbac:groups=core,resources=configmaps,namespace=kube-system,resourceNames=amazon-vpc-cni,verbs=get;list;watch
@@ -65,7 +70,7 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	// Check if the flag value has changed
+	// Check if the Windows IPAM flag has changed
 	newWinIPAMEnabledCond := r.Condition.IsWindowsIPAMEnabled()
 
 	var isIPAMFlagUpdated bool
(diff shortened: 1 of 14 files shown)


pkg/provider/prefix/provider.go
```diff
@@ -0,0 +1,428 @@
...
+	} else {
+		capacity = (limits.IPv4PerInterface - 1) * limits.Interface
+	}
+
+	return capacity
+}
+
+// Retrieve dynamic configuration for prefix delegation from config map, else use default warm pool config
+func (p *ipv4PrefixProvider) getPDWarmPoolConfig() *config.WarmPoolConfig {
+	resourceConfig := make(map[string]config.ResourceConfig)
```
haouc, line 419: nit: do we need to create a map since the variable will take whatever map returned from the functions?

haouc (commented):
Overall looks good to me. Just several comments.  Also please consider to create a PR to go over metrics and check if all necessary metrics have been added for better ops observability.

pkg/provider/prefix/provider.go
jiechen0826: I initialized the map here because the following `if-else` block requires `resourceConfig` to be declared first. The alternative is `var resourceConfig map[string]config.ResourceConfig` but that would leave the map uninitialized and have a value of `nil`. But I guess that should be harmless since it will be replace by the results returned from the functions anyway. I can change it to `var resourceConfig map[string]config.ResourceConfig` if that works.

pkg/config/loader.go
```diff
@@ -59,6 +65,61 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
+	warmPrefixTargetStr, foundWarmPrefix := vpcCniConfigMap.Data[WarmPrefixTarget]
+
+	// If no configuration is found, return 0
+	if !foundWarmIP && !foundMinIP && !foundWarmPrefix {
+		return warmIPTarget, minIPTarget, warmPrefixTarget
+	}
+
+	if foundWarmIP {
+		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
+		if err == nil {
```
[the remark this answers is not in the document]
jiechen0826: This file doesn't have a logger, should I introduce one here?

controllers/core/configmap_controller.go
```diff
@@ -87,8 +90,21 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		isPrefixFlagUpdated = true
 	}
 
+	// Check if configurations for Windows prefix delegation have changed
+	var isPDConfigUpdated bool
+	warmIPTarget, minIPTarget, warmPrefixTarget := config.ParseWinPDTargets(configmap)
+	if r.curWinPDWarmIPTarget != warmIPTarget || r.curWinPDMinIPTarget != minIPTarget || r.curWinPDWarmPrefixTarget != warmPrefixTarget {
+		r.curWinPDWarmIPTarget = warmIPTarget
+		r.curWinPDMinIPTarget = minIPTarget
+		r.curWinPDWarmPrefixTarget = warmPrefixTarget
```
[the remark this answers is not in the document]
jiechen0826: That's a valid concern. Currently this controller doesn't support concurrent reconciliation. As discussed offline, I will explicitly set the `MaxConcurrentReconciles` to 1 so that it is only run as a single goroutine. If the controller wants to support concurrent reconciliation in the future, we can revisit the logic later.

pkg/config/loader.go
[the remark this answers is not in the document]
haouc: hmm, if you feel logs will be critical for prefix debug, then yes please go ahead to pass in a logger instance.

pkg/provider/prefix/provider.go
haouc: yeah, as long as the function returns a map no matter if it succeeds. It should be safe here. If the function returns nil, it is also not safe here.

pkg/config/loader.go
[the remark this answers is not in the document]
jiechen0826: Yea I think logs would be helpful. Thanks for the suggestion. I've added logging here.

pkg/provider/prefix/provider.go
jiechen0826: That makes sense. I double checked both functions won't return nil, so it should be safe to declare `var resourceConfig map[string]config.ResourceConfig`. Updated.

jiechen0826:
> Overall looks good to me. Just several comments. Also please consider to create a PR to go over metrics and check if all necessary metrics have been added for better ops observability.

Thanks a lot for reviewing! Yea I will submit a new PR for all the metrics.

pkg/config/loader.go
```diff
@@ -59,6 +66,69 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
+	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmIPTarget = warmIPTarget
+	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.MinIPTarget = minIPTarget
+	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget
+
+	return resourceConfig
+}
+
+// ParseWinPDTargets parses config map for Windows prefix delegation configurations set by users
+func ParseWinPDTargets(vpcCniConfigMap *v1.ConfigMap) (warmIPTarget int, minIPTarget int, warmPrefixTarget int) {
+	var log = ctrl.Log.WithName("parse windows prefix delegation targets from config map")
```
haouc, line 88: I think we better pass logger from `ParseWinPDTargets`'s callers instead of creating a new log instance. I saw only two callers, provider and ConfigMap_controller both which have logger you can pass here.

jiechen0826: Sounds good, I've updated accordingly.

haouc (approved):
Looks good to me. Thanks.

Issue #231, opened 2023-05-27
Branch ENI provider emits warning when IPAM is enabled for Windows node


jiechen0826:
**What would you like to be enhanced**:
Using v1.1.6 image, describing node shows this warning message for Windows node:
  `  Warning  Unsupported         76s (x3 over 3m23s)  vpc-resource-controller  The instance type c5.xlarge is not supported for trunk interface (Security Group for Pods)`

Pods IP allocation works fine and the controller does not actually initialize the branch ENI. It's just that checking `IsInstanceSupported` will emit the warning to Windows node.

**Why is the change needed and what use case will it solve**:
The warning didn't cause any issues, but it would be confusing to users that may lead them to think something is wrong with the controller.

haouc:
Thanks for reporting this issue, @jiechen0826 . This should be an incorrect message being sent to Windows node when trunk ENI provider error out from provisioning Windows resources. We should either not invoking trunk provider for Windows workflow, or not generate event for Windows node. I will make a change soon.

haouc:
The fix has been merged as #232. The next release will include this fix. I am keeping this issue open for visibility till it has been rolled out to all clusters.

haouc:
The rollout has been finished. Closing this ticket.

PR #232, opened 2023-05-27
Windows events were incorrectly emitted

haouc:
*Issue #, if available:*
#231 
*Description of changes:*
When branch ENI provider error out from Windows provisioning, it shouldn't emit event for warning users the instance type is not supported.

Tests were done and confirmed the behavior is fixed.
- Unsupported Linux instance type node still have the warning message when trying to enable SGP feature
- Windows node no longer has the warning event when enabling Windows support through ConfigMap

### DIFF
--- a/pkg/provider/branch/provider.go
+++ b/pkg/provider/branch/provider.go
@@ -466,8 +466,12 @@ func (b *branchENIProvider) GetPool(_ string) (pool.Pool, bool) {
 
 // IsInstanceSupported returns true for linux node as pod eni is only supported for linux worker node
 func (b *branchENIProvider) IsInstanceSupported(instance ec2.EC2Instance) bool {
+	if instance.Os() != config.OSLinux {
+		return false
+	}
+
 	limits, found := vpc.Limits[instance.Type()]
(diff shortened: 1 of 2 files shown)


pkg/provider/branch/provider.go
```diff
@@ -469,7 +469,7 @@ func (b *branchENIProvider) IsInstanceSupported(instance ec2.EC2Instance) bool {
 	limits, found := vpc.Limits[instance.Type()]
 	supported := found && instance.Os() == config.OSLinux && limits.IsTrunkingCompatible
 
-	if !supported {
+	if !supported && instance.Os() == config.OSLinux {
```
sushrk, line 472: Nit: for readability, should we have the logic like

```
// SGPP is not supported on Windows nodes
if instance.Os() == config.OSWindows{
  return false
} else {
  limits, found := vpc.Limits[instance.Type()]
  if  !(found && limits.IsTrunkingCompatible) {
    (Linux node- raise event)
    return false
  }
}
return true
```

haouc: I agree. Returning Windows as false earlier makes sense to me. Will make the change.

sushrk (approved):
Nice, lgtm! Thank you.

PR #234, opened 2023-05-29
added nitro check to prefix provider and pod controller

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- added nitro check to prefix provider

### DIFF
--- a/pkg/provider/prefix/provider.go
+++ b/pkg/provider/prefix/provider.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker"
 	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 )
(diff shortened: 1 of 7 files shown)


pkg/provider/prefix/provider.go
```diff
@@ -334,8 +334,12 @@ func (p *ipv4PrefixProvider) GetPool(nodeName string) (pool.Pool, bool) {
...
-	//TODO add check for Nitro instances
-	if instance.Os() == config.OSWindows {
+	limits, found := vpc.Limits[instance.Type()]
+	if !found {
+		return false
+	}
+
+	if instance.Os() == config.OSWindows && (limits.IsBareMetal || limits.Hypervisor == "nitro") {
 		return true
 	}
```
haouc, line 345: Can you emit an event for cases of not found and not supported?

jiechen0826: Yea sure, I've added node events for both cases.

jiechen0826:
I'm checking why the unit test failed here, and it looks like it's a connectivity issue with Github. The unit tests are passing locally. I can force update to trigger the check again.

pkg/provider/prefix/provider.go
haouc: Thanks, @jiechen0826. Please also make sure these user faced messages are tested (positive and negative).

haouc (approved):
lgtm, thanks

PR #235, opened 2023-05-29
use GetAssignedResource to check if pod is managed by prefix IP pool

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Previously, we used resource annotation to check if the resource is managed by a given pool. But this method won't work for a special case, where a pod with TerminationGracePeriodSeconds set to 0 is created and deleted immediately, the delete event doesn't contain the resource annotation. In such cases, we won't be able to find the resource by resource annotation. Instead, we need to use pod ID to get the assigned resource. ([Please see this comment for same reasoning](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/handler/warm.go#L143-L148)) Therefore, in order to find the correct resource pool to handle pod event, we always choose to use pod ID to query the resource. 
- Remove previously added API `IsManagedResource(resourceID)` from pool interface


controllers/core/pod_controller.go
```diff
@@ -258,8 +252,10 @@ func (r *PodReconciler) updateResourceName(isDeletionEvent bool, pod *v1.Pod) st
 		}
 		resourcePool, ok := prefixProvider.GetPool(pod.Spec.NodeName)
 		// If IP is managed by prefix IP pool, update resource name so that prefix IP handler will be used
-		if ok && resourcePool.IsManagedResource(resourceID) {
-			resourceName = config.ResourceNameIPAddressFromPrefix
+		if ok {
+			if _, ownsResource := resourcePool.GetAssignedResource(string(pod.UID)); ownsResource {
```
haouc, line 256: Irrelavent to this PR. Maybe we need to revist the pool operations for checking if items in pool are dynamically managed well (no leakage).

haouc (approved):

PR #236, opened 2023-05-29
updated to not create new secondary ENI for Windows

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Since Windows doesn't support multi-ENI yet, we don't want to create a new ENI when the resource request cannot be fulfilled. The check for OS can be removed once multi-ENI is supported.
- Modified unit tests for above changes.

### DIFF
--- a/pkg/provider/ip/eni/eni.go
+++ b/pkg/provider/ip/eni/eni.go
@@ -148,8 +148,8 @@ func (e *eniManager) CreateIPV4Resource(required int, resourceType config.Resour
 				log.Error(err, "failed to assign all the requested resources",
 					"requested", want, "got", len(assigned))
 			}
-			// Update the remaining capacity
-			e.attachedENIs[index].remainingCapacity = remainingCapacity - canAssign
+			// Update the remaining capacity, assigned can be empty while err is nil
+			e.attachedENIs[index].remainingCapacity = remainingCapacity - len(assigned)
 			// Append the assigned IPs on this ENI to the list of IPs created across all the ENIs
 			assignedIPv4Resources = append(assignedIPv4Resources, assigned...)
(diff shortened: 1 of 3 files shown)


pkg/provider/ip/eni/eni.go
```diff
@@ -169,8 +169,11 @@ func (e *eniManager) CreateIPV4Resource(required int, resourceType config.Resour
 
 	// If the existing ENIs could not assign the required resources, loop till the new ENIs can assign the required
 	// number of IPv4 resources
+	// Note: Windows don't support multi-ENI yet, so don't create new ENI even if primary ENI cannot assign required resources.
 	for len(assignedIPv4Resources) < required &&
-		len(e.attachedENIs) < eniLimit {
+		len(e.attachedENIs) < eniLimit &&
+		e.instance.Os() != config.OSWindows {
```
haouc, line 175: What is this for? The controller doesn't provision IP resources for trunk/branch interface in Linux.

jiechen0826: Initially the existing code had implemented creating new ENI when there is not enough network interface to assign IPv4 resources on the primary ENI. However, Windows node doesn't support multi-ENI at the moment. Adding this check for `e.instance.Os() != config.OSWindows` here essentially skipped the following code but still kept the code there in case in the future when multi-ENI support for Windows is added, removing this check will re-enable the following code to create new ENI easily. Please let me know if you think it makes more sense to remove it entirely from the code,

haouc: I think this equality check is misleading here. Commenting this code block with a TODO comment makes more sense to me.

haouc (approved):
lgtm, thanks.

PR #237, opened 2023-05-29
added integration tests for Windows prefix delegation

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Added test cases for when PD is enabled, basic pod deployment works
- Added test cases for when PD is enabled initially, then toggled to secondary IP mode, prefixes should be released while secondary IPs are added to warm pool
- Added test cases for when PD is enabled, `warm-prefix-target` works
- Added test cases for when PD is enabled, if the old VPC RC is deployed, pod deployment should fail

### DIFF
--- a/test/framework/verify/pod.go
+++ b/test/framework/verify/pod.go
@@ -155,6 +155,15 @@ func (v *PodVerification) WindowsPodHaveIPv4Address(pod *v1.Pod) {
 
 }
 
+func (v *PodVerification) WindowsPodHaveIPv4AddressFromPrefix(pod *v1.Pod) {
+	By("matching the prefix-deconstructed IPv4 from annotation to the pod IP")
+	ipAddWithCidr, found := pod.Annotations["vpc.amazonaws.com/PrivateIPv4Address"]
+	Expect(found).To(BeTrue())
+	// Remove the CIDR and compare pod IP with the IP annotated from VPC Controller
+	Expect(pod.Status.PodIP).To(Equal(strings.Split(ipAddWithCidr, "/")[0]))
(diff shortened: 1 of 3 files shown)


test/framework/verify/pod.go
```diff
@@ -155,6 +155,15 @@ func (v *PodVerification) WindowsPodHaveIPv4Address(pod *v1.Pod) {
 
 }
 
+func (v *PodVerification) WindowsPodHaveIPv4AddressFromPrefix(pod *v1.Pod) {
+	By("matching the prefix-deconstructed IPv4 from annotation to the pod IP")
+	ipAddWithCidr, found := pod.Annotations["vpc.amazonaws.com/PrivateIPv4Address"]
+	Expect(found).To(BeTrue())
+	// Remove the CIDR and compare pod IP with the IP annotated from VPC Controller
+	Expect(pod.Status.PodIP).To(Equal(strings.Split(ipAddWithCidr, "/")[0]))
```
haouc, line 163: nit: I recall you added a func which can help extract IP and mask from the string? Can we use that here if I recall correctly?

haouc (approved):

test/framework/verify/pod.go
jiechen0826: Yea there is a helper function to deconstruct prefix into IPs `DeconstructIPsFromPrefix()`. Here we will need to check if the deconstructed IPs contain the pod's IP. Should I submit a new PR for the improvement or add commit to this PR?

haouc: Adding a commit is good enough. Thanks.

haouc: We can merge this and if you have another PR coming. We can update there. Thanks.

PR #240, opened 2023-05-31
fixed capactiy advertisement when node start on secondary IP mode

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Fixed the issue of missing node capacity advertisement when node start on secondary IP mode. The cause was an incorrect condition resulting in returning the function too early and skipping capacity advertisement.
- Added license statement to prefix provider file.
- Added node event for error in `InitResources` to prefix provider to be consistent with secondary IP provider.
- Added a check for nitro status when toggling from PD to secondary IP
- Refactored common code for nitro check

### DIFF
--- a/pkg/provider/ip/provider.go
+++ b/pkg/provider/ip/provider.go
@@ -30,9 +30,9 @@ import (
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider/ip/eni"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/worker"
-	v1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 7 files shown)


pkg/provider/ip/provider.go
```diff
@@ -295,7 +297,7 @@ func (p *ipv4Provider) ReSyncPool(job *worker.WarmPoolJob) {
 	}
 
 	ipV4Resources, err := providerAndPool.eniManager.InitResources(p.apiWrapper.EC2API)
-	if err != nil || ipV4Resources == nil {
```
haouc, line 298: Can ipV4Resources be nil? Seems it is a pointer and could be nil, right?

jiechen0826: Yes you are right. Thanks for catching it. I will add it back.

haouc (approved):

PR #241, opened 2023-06-01
fix an incorrect argument passed to SendNodeEvent

haouc:
*Issue #, if available:*

*Description of changes:*
The argument `reason` was incorrectly used with const instead of passed value.


jdn5126 (approved):

haouc:
Seems like a periodical issue from GitHub... , it is ongoing...
https://github.com/niemeyer/gopkg/issues/78#issuecomment-1307802989
Since this PR shouldn't have impact and local tests were passed, I am merging it.

PR #242, opened 2023-06-01
changed node capacity initialization to only exclude the used IPs from the alternative provider

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Changed node capacity initialization to only exclude the used IPs from the alternative provider. When a node joins, or the controller restarts, both IPv4 provider and prefix provider are initialized with respective calculated capacity. Previously this capacity calculation would exclude any used or warm IPs of the alternative provider. This change made the calculation only exclude the used IPs, not the warm IPs. That way when user toggles from one mode to the other, if there was no used secondary IP or prefix from the previous mode, the node capacity would be accurate for the new mode. 

### DIFF
--- a/pkg/provider/ip/provider.go
+++ b/pkg/provider/ip/provider.go
@@ -111,6 +111,21 @@ func (p *ipv4Provider) InitResource(instance ec2.EC2Instance) error {
 
 	podToResourceMap := map[string]pool.Resource{}
 	usedIPSet := map[string]struct{}{}
+
+	// Construct map of all possible IPs to prefix for each assigned prefix
+	ipToPrefixMap := make(map[string]string)
+	usedPrefixSet := make(map[string]struct{})
+	for _, prefix := range presentPrefixes {
+		prefixIPs, err := utils.DeconstructIPsFromPrefix(prefix)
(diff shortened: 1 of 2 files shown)


pkg/provider/ip/provider.go
```diff
@@ -111,6 +111,20 @@ func (p *ipv4Provider) InitResource(instance ec2.EC2Instance) error {
...
 	podToResourceMap := map[string]pool.Resource{}
 	usedIPSet := map[string]struct{}{}
+
+	// Construct map of all possible IPs to prefix for each assigned prefix
+	ipToPrefixMap := make(map[string]string)
+	usedPrefixSet := make(map[string]struct{})
+	for _, prefix := range presentPrefixes {
+		prefixIPs, err := utils.DeconstructIPsFromPrefix(prefix)
+		if err != nil {
+			continue
```
haouc, line 121: Should we at least log the error?

jiechen0826: Thanks for the suggestion. I've added the log here.

haouc (approved):
lgtm, thanks.

PR #244, opened 2023-06-02
added more integration tests for PD configs

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Added integ test for `warm-ip-target`
- Added integ test for `minimum-ip-target`
- Refactored function to check if pod's IP is from a prefix

### DIFF
--- a/test/framework/verify/pod.go
+++ b/test/framework/verify/pod.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider/branch/trunk"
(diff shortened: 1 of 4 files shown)


test/framework/verify/pod.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider/branch/trunk"
+	helper_utils "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
```
haouc, line 24: Let's not use underscore for imported package alias. By convention, all lower cases and compressed names are fine. I think helper or utils are fine here.

jiechen0826: Sounds good, I've changed it to `helper`.

haouc (approved):
lgtm

PR #246, opened 2023-06-03
added a feature flag so that Windows PD can be disabled from controll…

jiechen0826:
…er manifest

*Issue #, if available:*

*Description of changes:*
- Added a feature flag so that Windows PD can be disabled from controller manifest

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,6 +98,7 @@ func main() {
 	var introspectBindAddr string
 	var leaseOnly bool
 	var healthCheckTimeout int
+	var enableWindowsPrefixDelegation bool
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080",
 		"The address the metric endpoint binds to.")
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -130,6 +131,8 @@ func main() {
 	flag.StringVar(&introspectBindAddr, "introspect-bind-addr", ":22775",
 		"Port for serving the introspection API")
 	flag.BoolVar(&leaseOnly, "lease-only", false, "Controller uses lease only for leader election")
+	flag.BoolVar(&enableWindowsPrefixDelegation, "enable-windows-prefix-delegation", true,
```
haouc, line 134: Can we make the default to 'false' thus we only need to turn this on for 1.25+

jiechen0826: Yeah sure, updated it to `false`

haouc (approved):
Thanks for maki this configurable. Looks good to me.

PR #248, opened 2023-06-03
fixed unstable test case for minimum-ip-target

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Fixed the unstable test case for `minimum-ip-target`. It passed consistently for the last 5 runs.


haouc:
Do you have tests covering this use case after removing the test?

jiechen0826:
> Do you have tests covering this use case after removing the test?

The current code checks for `minimum-ip-target` when no pod is running. The part I removed also checks when removing all running pods still meets the requirement for `minimum-ip-target`, but it wasn't stable after a few minutes. I will try to find a way to make the test more stable.

haouc (approved):
Thanks for making them work. Looks good to me.

PR #249, opened 2023-06-04
add support for Windows prefix delegation

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- This PR adds support for Windows prefix delegation to increase pod density per node. Instead of assigning secondary IPv4 address, this feature assigns IPv4 prefix to primary ENI on Windows node.
- Can enable the feature by adding an entry "enable-windows-prefix-delegation: true" in `amazon-vpc-cni` config map after enabling IPAM.
- Can set either `warm-ip-target` and/or `minimum-ip-target`, or `warm-prefix-target` in `amazon-vpc-cni` config map   to fine-tune the warm pool configuration used in prefix delegation.


jiechen0826:
@haouc The PR still shows all the commits from the feature branch, but I would like to squash all commits into one. Should I squash them locally and then force-push to the feature branch, and then submit the PR to merge into master? Please advise how to merge the feature branch. Thanks a lot!

PR #261, opened 2023-06-07
fixed log statements with incorrect args

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Some log statements have incorrect number of args. Replaced `fmt.Errorf` with package-defined `utils.ErrNotFound` to be consistent.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -684,11 +684,11 @@ func (p *pool) getPDDeviation() int {
 	}
 
 	if deviationPrefix != 0 {
-		p.log.Info("calculating IP deviation for prefix pool", "p.warmPoolConfig", p.warmPoolConfig,
-			"existing warm resources", numExistingWarmResources, "p.pendingCreate", p.pendingCreate, "p.pendingDelete", p.pendingDelete,
-			"numTotalResources", numTotalResources, "numCurrPrefix", numCurrPrefix, "numTotalResForWarmIPTarget", numTotalResForWarmIPTarget,
-			"numPrefixForWarmIPTarget", numPrefixForWarmIPTarget, "numPrefixForMinIPTarget", numPrefixForMinIPTarget,
-			"deviation", deviationPrefix*NumIPv4AddrPerPrefix)
+		p.log.V(1).Info("calculating IP deviation for prefix pool", "warmPoolConfig", p.warmPoolConfig,
+			"used resources", len(p.usedResources), "existing warm resources", numExistingWarmResources, "pendingCreate", p.pendingCreate,
+			"pendingDelete", p.pendingDelete, "numTotalResources", numTotalResources, "numCurrPrefix", numCurrPrefix,
+			"numTotalResForWarmIPTarget", numTotalResForWarmIPTarget, "numPrefixForWarmIPTarget", numPrefixForWarmIPTarget,
+			"numPrefixForMinIPTarget", numPrefixForMinIPTarget, "deviation", deviationPrefix*NumIPv4AddrPerPrefix)
 	}
 
 	return deviationPrefix * NumIPv4AddrPerPrefix
--- a/pkg/provider/ip/provider.go
+++ b/pkg/provider/ip/provider.go
@@ -192,7 +192,7 @@ func (p *ipv4Provider) DeInitResource(instance ec2.EC2Instance) error {
 func (p *ipv4Provider) UpdateResourceCapacity(instance ec2.EC2Instance) error {
 	resourceProviderAndPool, isPresent := p.getInstanceProviderAndPool(instance.Name())
 	if !isPresent {
-		p.log.Error(nil, "cannot find the instance provider and pool form the cache", "node-name", instance.Name())
+		p.log.Error(utils.ErrNotFound, utils.ErrMsgProviderAndPoolNotFound, "node name", instance.Name())
 		return nil
 	}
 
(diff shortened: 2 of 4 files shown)


haouc (approved):
lgtm

pkg/provider/ip/provider.go
```diff
@@ -192,7 +192,7 @@ func (p *ipv4Provider) DeInitResource(instance ec2.EC2Instance) error {
 func (p *ipv4Provider) UpdateResourceCapacity(instance ec2.EC2Instance) error {
 	resourceProviderAndPool, isPresent := p.getInstanceProviderAndPool(instance.Name())
 	if !isPresent {
-		p.log.Error(nil, "cannot find the instance provider and pool form the cache", "node-name", instance.Name())
+		p.log.Error(utils.ErrNotFound, "cannot find the instance provider and pool form the cache", "node-name", instance.Name())
 		return nil
```
jiechen0826, line 196: Hi @haouc , the key-value pair has been fixed in this PR.

jiechen0826, line 196: My question pertains to the next line, where I should return `nil` or the actual error `utils.ErrNotFound` so that the caller can process it.

pkg/provider/ip/provider.go
```diff
@@ -320,8 +320,7 @@ func (p *ipv4Provider) CreatePrivateIPv4AndUpdatePool(job *worker.WarmPoolJob) {
 func (p *ipv4Provider) ReSyncPool(job *worker.WarmPoolJob) {
 	providerAndPool, found := p.instanceProviderAndPool[job.NodeName]
 	if !found {
-		p.log.Error(fmt.Errorf("instance provider not found"), "node is not initialized",
-			"name", job.NodeName)
+		p.log.Error(utils.ErrNotFound, "instance provider not found", "node is not initialized", "name", job.NodeName)
```
haouc, line 323: This seems not pairing correctly.

pkg/provider/prefix/provider.go
```diff
@@ -316,8 +316,7 @@ func (p *ipv4PrefixProvider) DeleteIPv4PrefixAndUpdatePool(job *worker.WarmPoolJ
 func (p *ipv4PrefixProvider) ReSyncPool(job *worker.WarmPoolJob) {
 	providerAndPool, found := p.instanceProviderAndPool[job.NodeName]
 	if !found {
-		p.log.Error(fmt.Errorf("instance provider not found"), "node is not initialized",
-			"name", job.NodeName)
+		p.log.Error(utils.ErrNotFound, "instance provider not found", "node is not initialized", "name", job.NodeName)
```
haouc, line 319: This seems also not pairing correctly?

haouc (changes requested):

pkg/provider/ip/provider.go
jiechen0826: Thanks for catching it. Fixed it now.

pkg/provider/prefix/provider.go
jiechen0826: Thanks for catching it. Fixed it now.

jiechen0826:
Hi @haouc, could you please take a look at these changes? Thanks a lot!

pkg/pool/pool.go
```diff
@@ -684,7 +684,7 @@ func (p *pool) getPDDeviation() int {
 	}
 
 	if deviationPrefix != 0 {
-		p.log.Info("calculating IP deviation for prefix pool", "p.warmPoolConfig", p.warmPoolConfig,
+		p.log.V(1).Info("calculating IP deviation for prefix pool", "p.warmPoolConfig", p.warmPoolConfig,
```
sushrk, line 687: nit: maybe change key to "warmPoolConfig" as 'p.' might not be relevant for logs.

pkg/provider/ip/provider.go
```diff
@@ -339,7 +339,7 @@ func (p *ipv4Provider) ReSyncPool(job *worker.WarmPoolJob) {
 func (p *ipv4Provider) DeletePrivateIPv4AndUpdatePool(job *worker.WarmPoolJob) {
 	instanceResource, found := p.getInstanceProviderAndPool(job.NodeName)
 	if !found {
-		p.log.Error(fmt.Errorf("cannot find the instance provider and pool form the cache"), "node", job.NodeName)
+		p.log.Error(utils.ErrNotFound, "cannot find the instance provider and pool from the cache", "node", job.NodeName)
```
sushrk, line 342: Should we define this error in utils as it is used in multiple statements? 
Also would be good to keep the node name key consistent, "nodename" should work.

jiechen0826: Hi @sushrk, `utils.ErrNotFound` is already defined in utils. Are you referring to the message `"cannot find the instance provider and pool from the cache"`? I can move it to the same `errors.go` if that works.

pkg/pool/pool.go
jiechen0826: Sounds good, I've updated the name.

pkg/provider/ip/provider.go
jiechen0826: Thanks for the suggestion. I've updated both.

pkg/provider/ip/provider.go
jiechen0826: After offline discussion, we decided to keep it as it is to return `nil` here so that the caller can continue with the same behavior.

haouc (approved):
lgtm

PR #262, opened 2023-06-08
add missing lock when ENI manager init resources

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- In `eniManage.InitResources`, the method didn't lock around the shared `resourceToENIMap`.  This could cause concurrency issue when the same map is used by another goroutine. For example, the same `eniManage.InitResources` could be called by `ipv4Provider.InitResource` and `ipv4Provider.ReSyncPool`.


haouc (approved):
looks good to me.

PR #264, opened 2023-06-09
emit node event when error is InsufficientCidrBlocks in Windows PD mode

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- When the subnet doesn't have enough CIDR Blocks, Windows prefix delegation mode can fail to get a new prefix from EC2, which leads to pods stuck in creating. This node event can help customers quickly identify the reason why pods fail to run and take action to resolve the issue quickly. 

### DIFF
--- a/pkg/provider/prefix/provider.go
+++ b/pkg/provider/prefix/provider.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/api"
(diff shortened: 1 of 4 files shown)


pkg/provider/prefix/provider.go
```diff
@@ -288,6 +289,10 @@ func (p *ipv4PrefixProvider) CreateIPv4PrefixAndUpdatePool(job *worker.WarmPoolJ
 	if err != nil {
 		p.log.Error(err, "failed to create all/some of the IPv4 prefixes", "created resources", resources)
 		didSucceed = false
+		if strings.HasPrefix(err.Error(), utils.InsufficientCidrBlocksReason) {
```
haouc, line 292: This adds a dependency on EC2 API error. It is ok for now since it is just for sending a node event. Can you add a comment here to make a note.

jiechen0826: Sounds good, comment added.

haouc (approved):

haouc (approved):
lgtm

PR #268, opened 2023-06-17
do not resync or reconcile pool for InsufficientCidrBlocks error and add longer wait time for retry

jiechen0826:
*Issue #, if available:*

*Description of changes:*
- Do not resync pool for `InsufficientCidrBlocks` error as it results in unnecessary calls of `DescribeInstances`.
- Change retry wait time for`InsufficientCidrBlocks` error to 2 min.
- Add a new field in pool to indicate whether prefix is available in the subnet, so that it can inform pod controller of the error and its corresponding wait time.
- Send pod event when `InsufficientCidrBlocks` happens and pod needs to retry after 2 min.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -29,6 +29,7 @@ var (
 	ErrResourceAreBeingCooledDown = fmt.Errorf("cannot assign resource now, resources are being cooled down")
 	ErrResourcesAreBeingCreated   = fmt.Errorf("cannot assign resource now, resources are being created")
 	ErrWarmPoolEmpty              = fmt.Errorf("warm pool is empty")
+	ErrInsufficientCidrBlocks     = fmt.Errorf("InsufficientCidrBlocks: The specified subnet does not have enough free cidr blocks to satisfy the request")
 	ErrResourceAlreadyAssigned    = fmt.Errorf("resource is already assigned to the requestor")
 	ErrResourceDoesntExist        = fmt.Errorf("requested resource doesn't exist in used pool")
 	ErrIncorrectResourceOwner     = fmt.Errorf("resource doesn't belong to the requestor")
@@ -42,7 +43,7 @@ type Pool interface {
 	AssignResource(requesterID string) (resourceID string, shouldReconcile bool, err error)
 	FreeResource(requesterID string, resourceID string) (shouldReconcile bool, err error)
 	GetAssignedResource(requesterID string) (resourceID string, ownsResource bool)
-	UpdatePool(job *worker.WarmPoolJob, didSucceed bool) (shouldReconcile bool)
+	UpdatePool(job *worker.WarmPoolJob, didSucceed bool, prefixAvailable bool) (shouldReconcile bool)
 	ReSync(resources []string)
 	ReconcilePool() *worker.WarmPoolJob
 	ProcessCoolDownQueue() bool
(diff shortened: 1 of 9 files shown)


pkg/provider/prefix/provider.go
```diff
@@ -296,11 +296,16 @@ func (p *ipv4PrefixProvider) CreateIPv4PrefixAndUpdatePool(job *worker.WarmPoolJ
 
 	if err != nil {
 		p.log.Error(err, "failed to create all/some of the IPv4 prefixes", "created resources", resources)
-		didSucceed = false
+
 		//TODO: This adds a dependency on EC2 API error. Refactor later.
 		if strings.HasPrefix(err.Error(), utils.InsufficientCidrBlocksReason) {
```
haouc, line 301: can you make not retryable errors into a list? if we have the list in config package, we can extend the list for not retry errors?

pkg/provider/prefix/provider.go
```diff
@@ -296,11 +296,16 @@ func (p *ipv4PrefixProvider) CreateIPv4PrefixAndUpdatePool(job *worker.WarmPoolJ
...
+			// Send node event to inform user of insufficient CIDR blocks error
 			utils.SendNodeEvent(p.apiWrapper.K8sAPI, job.NodeName, utils.InsufficientCidrBlocksReason,
 				utils.ErrInsufficientCidrBlocks.Error(), v1.EventTypeWarning, p.log)
+		} else {
+			// For all other errors, set didSucceed as false to re-sync the pool.
+			// However, since InsufficientCidrBlocks error is not retryable, do not set didSucceed to false.
```
haouc, line 307: Makes sense here. But the `didSucceed` will confuse people in the future if it hits not retryable error but `didSucceed` == true...
Does that make sense to change `didSucceed` to `retry`? Or maybe for now, to avoid changing downstream code, we can rename it to `notRetry` which is true for not-retryable error but false to others?

pkg/provider/prefix/provider.go
jiechen0826: Which file do you suggest to put the list?

jiechen0826: Thanks for the suggestion, made the changes.

pkg/provider/prefix/provider.go
jiechen0826: Sounds good, renamed it to `notRetry` for now.

pkg/utils/errors.go
```diff
@@ -1,9 +1,23 @@
...
 	ErrNotFound                   = errors.New("resource was not found")
 	ErrInsufficientCidrBlocks     = errors.New("InsufficientCidrBlocks: The specified subnet does not have enough free cidr blocks to satisfy the request")
 	ErrMsgProviderAndPoolNotFound = "cannot find the instance provider and pool from the cache"
+	NotRetryErrors                = []string{ErrInsufficientCidrBlocks.Error()}
 )
+
+// ShouldRetryOnError returns true if the error is retryable, else returns false
+func ShouldRetryOnError(err error) bool {
+	for _, e := range NotRetryErrors {
+		if strings.HasPrefix(err.Error(), e) {
```
haouc, line 18: How confident this prefix will be matched for a long prefix `InsufficientCidrBlocks: The specified subnet does not have enough free cidr blocks to satisfy the request`?

pkg/pool/pool.go
```diff
@@ -322,6 +330,13 @@ func (p *pool) UpdatePool(job *worker.WarmPoolJob, didSucceed bool) (shouldRecon
 		log.Error(fmt.Errorf("warm pool job failed: %v", job), "operation failed")
 	}
 
+	if p.isPDPool {
+		p.prefixAvailable = prefixAvailable
```
haouc, line 334: I think this is the only place we update this variable `prefixAvailable`, isn't it?

jiechen0826: Yes that's right.

pkg/utils/errors.go
jiechen0826: That's a good point. I will make it match the shorter string `InsufficientCidrBlocks`.

haouc (approved):
lgtm

PR #269, opened 2023-06-21
Define environment variable and set it to empty by default

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
Script fails if `EXTRA_GINKGO_FLAGS` parameters is not defined. We are using `set -u` in this file which treats this as an error. With this change, we define the variable as empty if it is not defined/passed to the script


sushrk (approved):
Thanks for catching this!

PR #270, opened 2023-06-26
use the new CNINode CRD to manage the trunk initialization on nodes

haouc:
*Issue #, if available:*

*Description of changes:*
This change is switching nodes initialization/trunk interface initialization from being triggered by node labels to newly added CRD CNINode. The CNINode objects are
- 1-to-1 mapping to every node
- the CNINode object is referring to k8s node and will be created by this controller and deleted by k8s
- this CNINode will deliver the feature states from VPC CNI to this controller

Tests (with a new VPC CNI version whose PR will be created soon):
- SGP: no labels added by VPC CNI and features were added to CNINode
- SGP plus custom networking: no labels added by VPC CNI and features were added to CNINode
```
 % k get cnd
NAME                                           FEATURES
ip-xxx-xxx-xx-xxx.us-west-2.compute.internal   [{"name":"CustomNetworking","value":"us-west-2b"},{"name":"SecurityGroupsForPods"}]
```
```
% k get pod hello-world-5d97946fc7-knzvp -oyaml | grep pod-eni
    vpc.amazonaws.com/pod-eni: '[{"eniId":"eni-xxxxxxxxxxx","ifAddress":"06:a9:xx:xx:xx:xx","privateIp":"xxx.xxx.x.xxx","vlanId":1,"subnetCidr":"xxx.xxx.x.x/24"}]'
        vpc.amazonaws.com/pod-eni: "1"
        vpc.amazonaws.com/pod-eni: "1"
```
- upgrade from current VPC CNI to new VPC CNI
- downgrade from new VPC CNI to older VPC CNI
- scale to 300 nodes with CNINode support (used 50 sec to update all 300 CNINodes and used 3 min 40 sec to initialize all nodes' trunk interface and capacity annotation)


apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -25,15 +25,22 @@ const (
 	CustomNetworking      FeatureName = "CustomNetworking"
 )
 
+// Feature is a type of feature being supported by VPC resource controller and other AWS Services
+type Feature struct {
+	Name  FeatureName `json:"name,omitempty"`
+	Value string      `json:"value,omitempty"`
```
ellistarn, line 31: What are these values?

config/rbac/role.yaml
```diff
@@ -54,16 +54,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
-  - events.k8s.io
-  resources:
-  - events
-  verbs:
-  - create
```
ellistarn, line 62: We don't create events?

controllers/apps/deployment_controller_test.go
```diff
@@ -17,10 +17,10 @@ import (
...
 	"testing"
 
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
+	mock_condition "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
+	mock_k8s "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
+	mock_node "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
+	mock_manager "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
```
ellistarn, line 23: Why did these change? Are we missing linters?

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ellistarn, line 1: Why is this inside /core/? Looks like they're partially ordered by API group, but CNINode isn't part of code.

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package controllers
+
+import (
+	"context"
+
+	"github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
+	rcHealthz "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/healthz"
```
ellistarn, line 20: I'd import packages using their native naming wherever possible, as it helps keep naming consistent across a codebase / multiple codebases

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s *CNINodeController) SetupWithManager(mgr ctrl.Manager, healthzHandler *rcHealthz.HealthzHandler) error {
+	healthzHandler.AddControllersHealthCheckers(
+		map[string]healthz.Checker{
+			"health-cninode-controller": s.Check(),
+		},
+	)
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.CNINode{}).WithEventFilter(
+		eventFilters(),
```
ellistarn, line 82: Why doesn't the default event filter work for us?

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+			return false
+		},
+	}
+}
+
+func (s *CNINodeController) Check() healthz.Checker {
+	s.Log.Info("CNINode controller's healthz subpath was added")
+	// We can revisit this to use PingWithTimeout() instead if we have concerns on this controller.
+	return rcHealthz.SimplePing("configmap controller", s.Log)
+}
```
ellistarn, line 101: I'm curious what this checker is checking. AFAICT there's nothing that could deadlock here.

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	Log         logr.Logger
+	NodeManager manager.Manager
+	Context     context.Context
+	K8sAPI      k8s.K8sWrapper
+}
+
+func (s *CNINodeController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	cniNode, err := s.K8sAPI.GetCNINode(req.Name, req.Namespace)
+	if err != nil {
+		s.Log.Error(err, "Failed to get CNINode")
```
ellistarn, line 42: This error should be printed by controller runtime when you return the error.

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	s.Log.Info("The CNINode has been updated", "CNINode", nodeName, "Features", cniNode.Spec.Features)
+
+	if node, err := s.K8sAPI.GetNode(nodeName); err != nil {
+		// if not found the node, we don't requeue and just wait VPC CNI to send another request
+		s.Log.Info("CNINode reconciler didn't find the node and will do exponential retry", "Node", node.Name)
+		return ctrl.Result{}, err
+	} else {
+		// make sure we have the node in cache already
+		if _, found := s.NodeManager.GetNode(nodeName); !found {
+			s.Log.Info("CNINode controller could not find node, will try again", "NodeName", nodeName)
```
ellistarn, line 60: Thoughts on emitting this at debug level to avoid spamming logs on every reconcile loop?

ellistarn, line 60: Why is it possible to not find the node? Can we simply call the controller-runtime client for this?

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+			return ctrl.Result{Requeue: true}, nil
+		} else {
+			// add the node into working queue as UPDATE event
+			s.Log.Info("CNINode controller found the node and sending the node update")
+			if err = s.NodeManager.UpdateNode(nodeName); err != nil {
+				return ctrl.Result{}, err
```
ellistarn, line 66: We need an `IgnoreNotFound(err)` here as well.

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+		return ctrl.Result{}, err
+	} else {
+		// make sure we have the node in cache already
+		if _, found := s.NodeManager.GetNode(nodeName); !found {
+			s.Log.Info("CNINode controller could not find node, will try again", "NodeName", nodeName)
+			return ctrl.Result{Requeue: true}, nil
+		} else {
+			// add the node into working queue as UPDATE event
+			s.Log.Info("CNINode controller found the node and sending the node update")
+			if err = s.NodeManager.UpdateNode(nodeName); err != nil {
```
ellistarn, line 65: As far as I can tell, the only purpose of this controller is to trigger UpdateNode events? I'm wondering if it would make sense to instead setup a Watch for CNINode for the Node reconciler.

pkg/k8s/cninode_cleanup.go
```diff
@@ -0,0 +1,119 @@
...
+		map[string]healthz.Checker{
+			"health-cninode-cleaner": rcHealthz.SimplePing("cninode cleanup", e.Log),
+		},
+	)
+
+	return mgr.Add(e)
+}
+
+// StartCNINodeCleaner starts the CNINOde Cleaner routine that cleans up dangling CNINode created by the controller
+func (e *CNINodeCleaner) Start(ctx context.Context) error {
```
ellistarn, line 53: Can we use OwnerReferences to handle garbage collection, instead of building a custom controller for this?

pkg/k8s/wrapper.go
```diff
@@ -220,3 +224,59 @@ func (k *k8sWrapper) ListEvents(ops []client.ListOption) (*eventsv1.EventList, e
 	}
 	return events, nil
 }
+
+func (k *k8sWrapper) GetCNINode(cniNodeName, cniNodeNamespace string) (*rcv1alpha1.CNINode, error) {
+	cninode := &rcv1alpha1.CNINode{}
+	if err := k.cacheClient.Get(k.context, types.NamespacedName{
```
ellistarn, line 230: Check out client.ObjectKeyFromObject(cniNode)

pkg/k8s/wrapper.go
```diff
@@ -220,3 +224,59 @@ func (k *k8sWrapper) ListEvents(ops []client.ListOption) (*eventsv1.EventList, e
 	}
 	return events, nil
 }
+
+func (k *k8sWrapper) GetCNINode(cniNodeName, cniNodeNamespace string) (*rcv1alpha1.CNINode, error) {
```
ellistarn, line 228: I'm not totally sold on why we need this wrapper function, as it seems more or less identical to the default behavior of the client. But if we do keep this, consider

```suggestion
func (k *k8sWrapper) GetCNINode(nn types.NamespaceName) (*rcv1alpha1.CNINode, error) {
```

pkg/k8s/wrapper.go
```diff
@@ -220,3 +224,59 @@ func (k *k8sWrapper) ListEvents(ops []client.ListOption) (*eventsv1.EventList, e
...
+		Name:      cniNodeName,
+		Namespace: cniNodeNamespace,
+	}, cninode); err != nil {
+		return cninode, err
+	}
+	return cninode, nil
+}
+
+func (k *k8sWrapper) CreateCNINode(node *v1.Node) error {
+	err := retry.OnError(retry.DefaultBackoff,
```
ellistarn, line 240: Why do we implement retry inside of this client, instead of relying on the standard reconciler retry mechanisms?

pkg/k8s/wrapper.go
```diff
@@ -220,3 +224,59 @@ func (k *k8sWrapper) ListEvents(ops []client.ListOption) (*eventsv1.EventList, e
...
+			return client.IgnoreAlreadyExists(k.cacheClient.Create(k.context, cniNode))
+		})
+
+	// do not return error if the error is "already existing"
+	// add a wrapper here to prevent we use retry on the error above
+	return client.IgnoreAlreadyExists(err)
+}
+
+func (k *k8sWrapper) DeleteCNINode(cniNodeName, cniNodeNamespace string) error {
+	err := retry.OnError(retry.DefaultBackoff, func(err error) bool { return true }, func() error {
```
ellistarn, line 271: Same question for retrying here.

pkg/node/manager/manager.go
```diff
@@ -132,12 +137,20 @@ func (m *manager) AddNode(nodeName string) error {
...
 		log.Info("node is already processed, not processing add event again")
 		return nil
 	}
 
+	if err = m.CreateCNINodeIfNotExisting(k8sNode); err != nil {
+		m.Log.Error(err, "Failed to create CNINode for k8sNode", "NodeName", k8sNode.Name)
+		return err
+	} else {
+		// use an event to notify users that a CNINode has been created for this node
+		utils.SendNodeEventWithNodeObject(m.wrapper.K8sAPI, k8sNode, utils.CNINodeCreatedReason, "A new CNINode was created for this node", v1.EventTypeNormal, m.Log)
```
ellistarn, line 151: Do we need to create a kube event for this? Seems like it may expose unnecessary information to customers, and they can watch the CNI Node object if they need.

pkg/node/manager/manager.go
```diff
@@ -167,6 +180,24 @@ func (m *manager) AddNode(nodeName string) error {
...
+		if apierrors.IsNotFound(err) {
+			m.Log.Info("Will create a new CNINode", "CNINodeName", node.Name)
+			return m.wrapper.K8sAPI.CreateCNINode(node)
+		}
+		return err
+	} else {
+		if cniNode != nil && cniNode.Name == node.Name {
+			m.Log.Info("The CNINode is already existing", "CNINode", cniNode)
+		} else {
+			m.Log.Error(errors.New("API error caused unmatched CNINode"), "Found existing unmatched CNINode when creating CNINode", "NodeName", node.Name)
```
ellistarn, line 195: How could the name not match?

controllers/core/cninode_controller.go
haouc: I don't think node controller has a good place to add a watch for this CRD since node controller is purely for watching node updates and operating on them. The independent controller for this CRD will operate on node object as a resource regarding the CRD updates and I want to set it up as an overlay mechanism for even more resources use cases.

pkg/k8s/cninode_cleanup.go
haouc: We are already using the OwnerReference for CNINode CR. I am just paranoid on possible leakage for any reason k8s couldn't delete the CNINodes... That is also why I set this for every 12 hours.

pkg/k8s/wrapper.go
haouc: Makes sense to use namespacedname. Using this pattern to be aligned with other APIs and also easier for setup tests and call the client's API from other places.

pkg/node/manager/manager.go
haouc: Some race condition, users' mis-operation, or other causes?

pkg/node/manager/manager.go
haouc: Make sense, we don't have to. I will remove this event.

config/rbac/role.yaml
haouc: I don't think we are using the API Group `events.k8s.io` which was added for the deprecated event solution.

pkg/k8s/cninode_cleanup.go
ellistarn: > I am just paranoid on possible leakage for any reason

I'd be very cautious of this line of reasoning. The kubernetes garbage collector is a core mechanism that drives things like pod deletion when replicasets are deleted -- if we can't rely on it, Kubernetes will break in many ways. We should keep systems as simple as possible, and test them thoroughly to have confidence that the mechanisms are working as intended.

apis/vpcresources/v1alpha1/cninode_types.go
haouc: This is for custom networking CRD eniconfig's name. I was planning to call CNI's API to get the value instead of asking CNI to add into the CRD. But I realized the API won't work since it depends on aws-node local ENVs. Checking ENVs in pods won't be 100% reliable since users can set them up by different ways.

pkg/node/manager/manager.go
ellistarn: On line 184: https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/270/files#diff-e9cec4c2460e089f0f518c07f3d4c2fb40273486a8346d499ec5188395417772R184, we retrieve the node by name. The only way that the name wouldn't match, is if our client is fundamentally broken, and we have bigger problems.

controllers/apps/deployment_controller_test.go
haouc: I can check if we have lint called during make.

controllers/core/cninode_controller.go
haouc: Make sense. API should be a better place for CRDs.

controllers/core/cninode_controller.go
haouc: Ack. makes sense on naming alias.

controllers/core/cninode_controller.go
haouc: I think the default filter is allowing all 4 ops, isn't it? I only want to watch for valid Update and not delete. I think even create dosen't make sense.

controllers/core/cninode_controller.go
haouc: The controller itself doesn't have lock to probe, but would be possible for any reason a lower level issue could block the goroutine and trigger this probe failure?

controllers/core/cninode_controller.go
haouc: You are right. This is redundant.

controllers/core/cninode_controller.go
```diff
@@ -0,0 +1,101 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	nodeName := cniNode.Name
+	s.Log.Info("The CNINode has been updated", "CNINode", nodeName, "Features", cniNode.Spec.Features)
+
+	if node, err := s.K8sAPI.GetNode(nodeName); err != nil {
+		// if not found the node, we don't requeue and just wait VPC CNI to send another request
+		s.Log.Info("CNINode reconciler didn't find the node and will do exponential retry", "Node", node.Name)
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
haouc: It will since Get prints error.

controllers/core/cninode_controller.go
haouc: hmm, enabling debug log will be hard now. We need add the flag in control plane manifest. If this is important log for better visibility. I think we should keep it?

haouc: NodeManager GetNode from its cache. We have a node cache locally.

controllers/core/cninode_controller.go
haouc: Thanks catching this. Yeah, NotFound shouldn't be retried.

controllers/core/cninode_controller_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"context"
+	"testing"
+
+	"github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
+	mock_k8s "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
+	mock_manager "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
haouc: I haven't. I can check them later.

pkg/k8s/cninode_cleanup.go
haouc: Good point. Relying on it and trusting it. Ack. I will remove the extra cleanup.

pkg/node/manager/manager.go
haouc: Agree. I will remove the redundant check.

controllers/core/cninode_controller.go
ellistarn: I wouldn't worry about an extra reconcile loop on the creation. Ideally, controllers should just operate against observed state and not be aware of whether or not the object was created, updated, or refreshed (after a crash/restart/resync interval). In general, less deviations from the defaults == simpler and easier to reason about.

haouc: Ack. I will remove the customied filter. Thanks

controllers/core/cninode_controller_test.go
```diff
@@ -0,0 +1,125 @@
...
+		NodeManager: mockManager,
+		K8sAPI:      mockK8sWrapper,
+		Context:     context.TODO(),
+		Log:         zap.New(),
+	}
+
+	mockK8sWrapper.EXPECT().GetCNINode(reconcileRequest.Name, config.KubeDefaultNamespace).Return(cniNode, nil).Times(1)
+	mockK8sWrapper.EXPECT().GetNode(mockNodeName).Return(k8sNode, nil).Times(0)
+	mockManager.EXPECT().GetNode(mockNodeName).Return(mockNode, true).Times(0)
+	mockManager.EXPECT().UpdateNode(mockNodeName).Return(nil).Times(0)
```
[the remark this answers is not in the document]
haouc: Good point Ellis. I will take a action to groom our UTs in both projects. I will sync with you offline when I start working on that.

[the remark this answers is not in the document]
haouc: I agree `Times(0)` is more like a self-explanation than a test needed code.

pkg/k8s/wrapper.go
haouc: I think letting reconciler to retry makes more sense in this case. I will do ignore exists and let reconciler retry on error.

pkg/k8s/wrapper.go
haouc: This func has been removed since we won't want to add redundant garbage collector.

controllers/core/cninode_controller.go
haouc: Yeah, the only vulnerable place is the node cache which has been guarded by node manager health check. I will remove this check.

pkg/k8s/wrapper.go
haouc: I am not sure why we need get the ObjectKey from the cninode after Get()? Can you elaborate your thought? Thanks

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(cniNode.Spec.Features) == 0 {
+		return ctrl.Result{}, nil
+	}
+
+	nodeName := cniNode.Name
+	s.Log.Info("The CNINode has been updated", "CNINode", nodeName, "Features", cniNode.Spec.Features)
+
+	if _, err := s.K8sAPI.GetNode(nodeName); err != nil {
+		return ctrl.Result{}, err
+	} else {
```
ellistarn, line 51: Do we need this else statement? Since the if will return.

haouc: As discussed offline, we are dropping the cninode controller and use node controller to watch its referred CNINode update.

controllers/apps/deployment_controller_test.go
haouc: Will update on this in another PR.

controllers/core/node_controller.go
```diff
@@ -66,6 +72,18 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{Requeue: true}, goErr.New("pod datastore hasn't been synced, node controller need wait to retry")
 	}
 
+	// CNINodes are created in default namespace
```
ellistarn, line 75: Woah! I just realized this. Because its 1:1 with Node, which is a cluster scoped object, we should make sure these CRDs are cluster scoped.

controllers/core/node_controller.go
```diff
@@ -113,7 +127,13 @@ func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager, healthzHandler *rcHe
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: MaxNodeConcurrentReconciles}).
-		Complete(r)
+		Watches(
```
ellistarn, line 130: If you're an owner, can you use .Owns()?

pkg/node/manager/manager.go
```diff
@@ -132,12 +139,21 @@ func (m *manager) AddNode(nodeName string) error {
...
 		return nil
 	}
 
+	if err = m.CreateCNINodeIfNotExisting(k8sNode); err != nil {
+		m.Log.Error(err, "Failed to create CNINode for k8sNode", "NodeName", k8sNode.Name)
+		// we shouldn't see AlreadyExists error here
+		// double-check here to avoid retry on this error
+		return client.IgnoreAlreadyExists(err)
+	} else {
+		m.Log.Info("A new CNINode was created or existing CNINode was found", "CNINodeName", k8sNode.Name)
```
ellistarn, line 154: Seems noisy to log here. Do you expect to use this as a signal of whether or not the feature is working? Is AddNode called idempotently, or regularly?

ellistarn, line 154: Can we log only when we create the node, and not when it already exists?

pkg/node/manager/manager.go
```diff
@@ -132,12 +139,21 @@ func (m *manager) AddNode(nodeName string) error {
...
+	_, nodeFound = m.dataStore[k8sNode.Name]
 	if nodeFound {
 		log.Info("node is already processed, not processing add event again")
 		return nil
 	}
 
+	if err = m.CreateCNINodeIfNotExisting(k8sNode); err != nil {
+		m.Log.Error(err, "Failed to create CNINode for k8sNode", "NodeName", k8sNode.Name)
+		// we shouldn't see AlreadyExists error here
+		// double-check here to avoid retry on this error
```
ellistarn, line 151: I'd be a bit cautious about "double checking". The code should work as designed, and not need double checking.

pkg/node/manager/manager.go
```diff
@@ -132,12 +139,21 @@ func (m *manager) AddNode(nodeName string) error {
...
+		return client.IgnoreAlreadyExists(err)
```
ellistarn, line 152: Maybe I'm confused, but shouldn't we continue with the logic below line 156 if the node exists? Perhaps we should move the IgnoreAlreadyExists code inside of `CreateCNINodeIfNotExisting`

pkg/node/manager/manager.go
```diff
@@ -167,6 +183,21 @@ func (m *manager) AddNode(nodeName string) error {
...
+	if cniNode, err := m.wrapper.K8sAPI.GetCNINode(
+		types.NamespacedName{Name: node.Name, Namespace: config.KubeDefaultNamespace},
+	); err != nil {
+		if apierrors.IsNotFound(err) {
+			m.Log.Info("Will create a new CNINode", "CNINodeName", node.Name)
+			return m.wrapper.K8sAPI.CreateCNINode(node)
+		}
+		return err
+	} else {
+		m.Log.V(1).Info("The CNINode is already existing", "CNINode", cniNode)
```
ellistarn, line 196: Seems potentially noisy to log here.

pkg/node/manager/manager.go
```diff
@@ -282,7 +313,21 @@ func (m *manager) DeleteNode(nodeName string) error {
 // using custom networking
 func (m *manager) updateSubnetIfUsingENIConfig(cachedNode node.Node, k8sNode *v1.Node) error {
 	eniConfigName, isPresent := k8sNode.Labels[config.CustomNetworkingLabel]
-	if isPresent {
+	if isPresent || m.customNetworkEnabledInCNINode(k8sNode) {
+		if !isPresent {
```
ellistarn, line 317: Since we're checking both isPresent and !isPresent, can we flatten these if statements into siblings, instead of nesting??

pkg/node/manager/manager.go
```diff
@@ -293,13 +338,26 @@ func (m *manager) updateSubnetIfUsingENIConfig(cachedNode node.Node, k8sNode *v1
...
 			return nil
 		}
-		return fmt.Errorf("failed to find subnet in eniconfig spec %s", eniConfigName)
 	} else {
 		cachedNode.UpdateCustomNetworkingSpecs("", nil)
 	}
 	return nil
 }
 
+func (m *manager) GetEniConfigNameFromCNINode(node *v1.Node) (string, error) {
```
ellistarn, line 347:
```suggestion
func (m *manager) GetEniConfigName(node *v1.Node) (string, error) {
```

pkg/node/manager/manager.go
```diff
@@ -293,13 +338,26 @@ func (m *manager) updateSubnetIfUsingENIConfig(cachedNode node.Node, k8sNode *v1
...
+	if err != nil {
+		return "", err
+	}
+	for _, feature := range cniNode.Spec.Features {
+		if feature.Name == v1alpha1.CustomNetworking && feature.Value != "" {
+			return feature.Value, nil
+		}
+	}
+
+	return "", apierrors.NewNotFound(schema.GroupResource{}, "couldn't find custom networking eniconfig name defined by aws-node")
```
ellistarn, line 358: schema.GroupResource{} seems wrong. In fact, I'm not sure we want to use this library, since it's built for detecting not found on API calls. Maybe better to use fmt.Errorf?

pkg/node/manager/manager.go
```diff
@@ -401,9 +459,35 @@ func isWindowsNode(node *v1.Node) bool {
...
+func (m *manager) canAttachTrunk(node *v1.Node) bool {
 	_, ok := node.Labels[config.HasTrunkAttachedLabel]
-	return ok
+	return ok || m.trunkEnabledInCNINode(node)
+}
+
+func (m *manager) trunkEnabledInCNINode(node *v1.Node) bool {
+	if cniNode, err := m.wrapper.K8sAPI.GetCNINode(
+		types.NamespacedName{Name: node.Name, Namespace: config.KubeDefaultNamespace},
+	); err == nil {
```
ellistarn, line 470: This ignores any error cases that might be inside of GetCNINode. We should bubble those error cases to the reconcile loop, instead of assuming the trunk is not enabled.

controllers/core/node_controller.go
haouc: changed to cluster scope.

controllers/core/node_controller.go
haouc: Owns() works. Simplify the code further. Thanks

pkg/node/manager/manager.go
haouc: Removed the log.

pkg/node/manager/manager.go
haouc: Cleaned up the log, thanks.

pkg/node/manager/manager.go
haouc: Makes sense. Change to fmt.

pkg/node/manager/manager.go
haouc: added error to return for bubbling up along the chain.

pkg/node/manager/manager.go
haouc: Removed it. You are right. It should work. The check is redundant.

pkg/node/manager/manager.go
haouc: I had to make some change here to work with bubbling up errors. But I feel nested if check in this case is easier to avoid repeat code.

pkg/node/manager/manager.go
haouc: Updated the name

pkg/node/manager/manager.go
haouc: Yeah, makes sense. this is the redundant error check as mentioned above. I have removed it. The inside already has an existing check.

pkg/k8s/wrapper_test.go
```diff
@@ -65,6 +66,13 @@ var (
 			Namespace: config.KubeSystemNamespace,
 		},
 	}
+
+	mockCNINode = &v1alpha1.CNINode{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nodeName,
+			Namespace: config.KubeDefaultNamespace,
```
ellistarn, line 73:
```suggestion
```

pkg/config/type.go
```diff
@@ -78,6 +78,7 @@ const (
 	MinimumIPTarget                  = "minimum-ip-target"
 	// Since LeaderElectionNamespace and VpcCniConfigMapName may be different in the future
 	KubeSystemNamespace            = "kube-system"
+	KubeDefaultNamespace           = "default"
```
ellistarn, line 81:
```suggestion
```

pkg/node/manager/manager.go
```diff
@@ -167,6 +182,21 @@ func (m *manager) AddNode(nodeName string) error {
...
+func (m *manager) CreateCNINodeIfNotExisting(node *v1.Node) error {
+	if cniNode, err := m.wrapper.K8sAPI.GetCNINode(
+		types.NamespacedName{Name: node.Name},
+	); err != nil {
+		if apierrors.IsNotFound(err) {
+			m.Log.Info("Will create a new CNINode", "CNINodeName", node.Name)
+			return m.wrapper.K8sAPI.CreateCNINode(node)
+		}
+		return err
+	} else {
```
ellistarn, line 194: AFAICT, this else doesn't do anything, since all other branches shortcircuit.

pkg/node/manager/manager.go
```diff
@@ -401,9 +478,40 @@ func isWindowsNode(node *v1.Node) bool {
...
+
+func (m *manager) trunkEnabledInCNINode(node *v1.Node) (bool, error) {
+	var err error
+	if cniNode, err := m.wrapper.K8sAPI.GetCNINode(
+		types.NamespacedName{Name: node.Name},
+	); err == nil {
+		if lo.ContainsBy(cniNode.Spec.Features, func(addedFeature v1alpha1.Feature) bool {
+			return addedFeature.Name == v1alpha1.SecurityGroupsForPods
+		}) {
+			return true, err
```
ellistarn, line 497:
```suggestion
			return true, nil
```

pkg/node/manager/manager.go
```diff
@@ -401,9 +478,40 @@ func isWindowsNode(node *v1.Node) bool {
...
+		}
+	}
+	return false, err
+}
+
+func (m *manager) customNetworkEnabledInCNINode(node *v1.Node) (bool, error) {
```
ellistarn, line 503: Consider unifying these functions:
```
func (m *manager) isFeatureEnabled(node *v1.Node, feature v1alpha1.Feature) {
...
   lo.Contains(cniNode.Spec.Features, feature) 
...
}
```

ellistarn (dismissed):
A few optional comments. Nice work!!

pkg/node/manager/manager.go
haouc: We don't need `else` but I think if...else should also work. This else is for the if check on `err != nil`

pkg/node/manager/manager.go
haouc: `err` must be nil in this block, but explicitly call this out may be better for reading.

PR #271, opened 2023-06-28
Adjust ginkgo timeout in test scripts

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Adjust ginkgo timeouts in canary and full integration test script. I checked the run times over 3-5 successful jobs which also runs the full Windows PD tests. 
This is required as we are adjusting our prow job schedule and need to ensure VPC-RC tests can finish within reasonable time.
We'll be monitoring the jobs and we can adjust the timeouts if any issues. 

Also includes changes to clean up the canary script, EKS 1.17 has been out of support for a long time and we do not run EKS 1.17 tests anymore.


haouc (approved):
LGTM. We probably should groom the tests instead of having to keep increasing timeout...

PR #278, opened 2023-07-10
Update unit tests and integration tests for CNINode

haouc:
*Issue #, if available:*

*Description of changes:*
We should add new test cases to new CNINode operation.

### DIFF
--- a/test/framework/resource/k8s/node/manager.go
+++ b/test/framework/resource/k8s/node/manager.go
@@ -16,6 +16,7 @@ package node
 import (
 	"context"
 
+	cninode "github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/test/framework/utils"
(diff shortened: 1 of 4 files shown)


test/framework/resource/k8s/node/manager.go
```diff
@@ -95,3 +99,21 @@ func (d *defaultManager) GetNode(node *v1.Node) (*v1.Node, error) {
 	err := d.k8sClient.Get(context.TODO(), utils.NamespacedName(node), observedNode)
 	return observedNode, err
 }
+
+func (d *defaultManager) GetCNINode(node *v1.Node) (*cninode.CNINode, error) {
```
ellistarn, line 103: I wonder how many of these functions you could remove with golang generics :).

ellistarn (approved):

test/framework/resource/k8s/node/manager.go
haouc: Yeah, I can explore them. 👍

PR #280, opened 2023-07-12
use individual cninode controller to avoid latency

haouc:
*Issue #, if available:*

*Description of changes:*
We are seeing some latency when using Owns() or Watches() CNINode from node controller. This PR is changing to independent CNINode controller to reconcile on the CRD and trigger node process.

Tests:
- using Owns(): initializing 200 nodes in EKS 1.25 cluster used ~ 7 min 30 sec
- using independent CNINode reconciler: initializing 200 nodes in EKS 1.25 cluster used ~ 2 min 30 sec which is same as the old way that CNI labels node to signal this controller.

### DIFF
--- a/controllers/core/node_controller.go
+++ b/controllers/core/node_controller.go
@@ -18,7 +18,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	rcHealthz "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/healthz"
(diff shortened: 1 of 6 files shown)


controllers/core/node_controller.go
```diff
@@ -113,7 +112,7 @@ func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager, healthzHandler *rcHe
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: MaxNodeConcurrentReconciles}).
-		Owns(&v1alpha1.CNINode{}).Complete(r)
+		Complete(r)
```
ellistarn, line 115: All this change does is doubles `MaxNodeConcurrentReconciles` since you're making two controllers. Have you thought about tuning the max concurrency?

PR #281, opened 2023-07-12
fix: Add controller=true to ownerreferences to correctly reconcile on .Owns()

ellistarn:
*Issue #, if available:*

*Description of changes:*

Currently the node_controller is not correctly wired up with CNINode Updates. It only triggers on node events. We're currently saved by the 5 minute node heartbeat, which happens to trigger attachment trunk attachment when it happens. This can cause trunk attachment tot take between 0-5 minutes instead of near-immediate reconciliation.

Controller-runtime expects controller=true on ownerreferences for `.Owns` to correctly enqueue the node keys. 


haouc (approved):
lgtm. Thanks for fixing this.

PR #287, opened 2023-08-03
Inject extra keys to STS client request header

haouc:
*Issue #, if available:*

*Description of changes:*
We want to add two extra context keys to header.

Tests have been done with extra condition in role and negative tests with incorrect key for account.

### DIFF
--- a/main.go
+++ b/main.go
@@ -101,6 +101,7 @@ func main() {
 	var leaseOnly bool
 	var healthCheckTimeout int
 	var enableWindowsPrefixDelegation bool
+	var region string
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080",
 		"The address the metric endpoint binds to.")
@@ -135,6 +136,7 @@ func main() {
 	flag.BoolVar(&leaseOnly, "lease-only", false, "Controller uses lease only for leader election")
 	flag.BoolVar(&enableWindowsPrefixDelegation, "enable-windows-prefix-delegation", false,
 		"Enable the feature flag for Windows prefix delegation")
+	flag.StringVar(&region, "aws-region", "", "The aws region of the k8s cluster")
 
 	flag.Parse()
 
@@ -256,7 +258,7 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
-	ec2Wrapper, err := ec2API.NewEC2Wrapper(roleARN, setupLog)
+	ec2Wrapper, err := ec2API.NewEC2Wrapper(roleARN, clusterName, region, setupLog)
 	if err != nil {
 		setupLog.Error(err, "unable to create ec2 wrapper")
 	}
--- a/pkg/aws/ec2/api/wrapper.go
+++ b/pkg/aws/ec2/api/wrapper.go
@@ -23,7 +23,6 @@ import (
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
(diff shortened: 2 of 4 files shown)


dims (approved):
/approve
/lgtm

pkg/aws/ec2/api/wrapper.go
```diff
@@ -477,8 +485,7 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion string, roleARN st
 	// If the regional STS endpoint is different than the global STS endpoint then add the global sts endpoint
 	if regionalSTSEndpoint.URL != globalSTSEndpoint.URL {
 		globalProvider := &stscreds.AssumeRoleProvider{
-			Client: sts.New(userStsSession, aws.NewConfig().WithHTTPClient(client).
-				WithEndpoint(regionalSTSEndpoint.URL).WithMaxRetries(MaxRetries)),
```
M00nF1sh, line 481: is this a bug fix? since the previous code here were using "regionalSTSEndpoint.URL" instead of globalSTSEndpoint. If so, why do we now want to enable globalSTS?

pkg/utils/helper.go
```diff
@@ -204,3 +209,20 @@ func IsNitroInstance(instanceType string) (bool, error) {
...
 	return false, nil
 }
+
+// GetSourceAcctAndArn constructs source acct and arn and return them for use
+func GetSourceAcctAndArn(roleARN, region, clusterName string) (string, string, error) {
+	// ARN format (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html)
+	// arn:partition:service:region:account-id:resource-type/resource-id
+	// IAM format, region is always blank
+	// arn:aws:iam::account:role/role-name-with-path
+	roleParts := strings.Split(roleARN, ":")
```
M00nF1sh, line 219: let's use https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/ pkg to deal with arn instead of split ourself.

main.go
```diff
@@ -177,6 +179,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if region == "" {
```
M00nF1sh, line 182: nit, this is ok to me, but since it's an new required arg, we should be extremely careful about this version especially for hot-patches. 
have we considered alternatively like only enable this sts header feature when it's specified or auto-discover region from ec2 metadata if not specified?

pkg/aws/ec2/api/wrapper.go
haouc: I think this was a bug. I don't think we ever used global.

main.go
haouc: I like the idea which only enable the sts header when region is provided. Will make an update.

pkg/aws/ec2/api/wrapper.go
haouc: Since we already compare `egionalSTSEndpoint.URL != globalSTSEndpoint.URL`, it makes sense to correct the endpoint to global one.

haouc: After syncing offline, we decided to use the old way (both use regional endpoint) and will investigate if we should remove global endpoint.

main.go
haouc: I have changed to check region. If empty, we use default client, otherwise we use enhanced client.

pkg/utils/helper.go
haouc: Updated to use sdk API.

pkg/aws/ec2/api/wrapper.go
```diff
@@ -492,6 +503,30 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion string, roleARN st
 
 }
 
+func (e *ec2Wrapper) createSTSClient(
+	userStsSession *session.Session,
+	client *http.Client,
+	endpoint endpoints.ResolvedEndpoint,
+	sourceAcct, sourceArn string) *sts.STS {
+	stsClient := sts.New(userStsSession, aws.NewConfig().WithHTTPClient(client).
```
orsenthil, line 511: nit: one line break could help with readability.

haouc: Added a new line. Thanks.

M00nF1sh (dismissed):
/lgtm

pkg/aws/ec2/api/wrapper.go
```diff
@@ -492,6 +503,31 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion string, roleARN st
...
+	parsedArn, err := arn.Parse(sourceArn)
+	if err == nil && parsedArn.Region != "" {
+		stsClient.Handlers.Sign.PushFront(func(s *request.Request) {
+			s.ApplyOptions(request.WithSetRequestHeaders(map[string]string{
+				SourceKey:  sourceArn,
+				AccountKey: sourceAcct,
+			}))
+		})
+	} else {
+		e.log.Info("Will use default STS client since unexpected error or cluster ARN", "Errors", err, "ParsedARN", parsedArn)
```
orsenthil, line 525: Just noting, falling to default client seems like better approach than crashing.

orsenthil (approved):
LGTM.

orsenthil (approved):
LGTM.

M00nF1sh (approved):
/lgtm
/approve

PR #288, opened 2023-08-09
handle the placeholder

haouc:
*Issue #, if available:*

*Description of changes:*
Due to the deployment gaps between systems, we may face the placeholder instead of real region in manifest. In that case, we want to use empty string to create default client.

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,9 +59,10 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


M00nF1sh (approved):
/approve
/lgtm

Issue #216, opened 2023-05-03
The controller should expose its version to users


haouc:
**What would you like to be enhanced**:
Since the controller runs in EKS control plane, users have limited visibility. But the controller should expose its version through a way then users can track for updates.
**Why is the change needed and what use case will it solve**:

PR #293, opened 2023-08-23
support version notice to nodes

haouc:
*Issue #, if available:*
#216 
*Description of changes:*
If the node is manageable (SGP or Windows), the controller will send one time event to the node with its version. Although we have mapping between EKS platform version and the controller version in every release note, this event can be a convenient way for customers to check from any client, such as `kubectl describe node`, in some use cases, such as automated workflows.

```
Events:
  Type    Reason                   Age   From                     Message
  ----    ------                   ----  ----                     -------
  Normal  ControllerVersionNotice  22s   vpc-resource-controller  The node is managed by VPC resource controller version v1.3.0
  Normal  NodeTrunkInitiated       21s   vpc-resource-controller  The node has trunk interface initialized successfully
```


sushrk (approved):
Nice, LGTM. Thanks!

PR #301, opened 2023-09-02
Patch aws node role for node update permissions

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This adds the update permission on aws-node for backward compatibility tests

### DIFF
--- a/test/integration/webhook/validating_webhook_suite_test.go
+++ b/test/integration/webhook/validating_webhook_suite_test.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
 )
 
 var frameWork *framework.Framework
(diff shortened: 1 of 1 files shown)


test/integration/webhook/validating_webhook_suite_test.go
```diff
@@ -97,4 +116,12 @@ var _ = AfterSuite(func() {
 	if securityGroupID != "" {
 		Expect(frameWork.EC2Manager.DeleteSecurityGroup(ctx, securityGroupID)).To(Succeed())
 	}
+
+	By("Removing the patch on aws-node cluster role")
+	if updatedClusterRole != nil {
+		updatedClusterRole.Rules = updatedClusterRole.Rules[:len(updatedClusterRole.Rules)-1]
```
haouc, line 122: Could this be a potential problem? I am not sure if the new permission can always be the last one when server returns. Although this will be the last test, I think we should still be resilient. Can you make it specifically delete the new permission on node? Thanks.

haouc (approved):
lgtm

orsenthil:
@jaydeokar , @haouc - To understand further, we did we need this in the VPC RC test suite?

Issue #295, opened 2023-08-29
Increase number of pod reconcilers


sushrk:
**What would you like to be enhanced**:
Currently, we use [`MaxPodConcurrentReconciles=4`](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/4e7c7ba2020409468646a69f30141cc70aea190a/controllers/core/pod_controller.go#L59) in the pod controller. 
We would like to increase this to an appropriate number to prevent delays in creation of pods using security groups. 

**Why is the change needed and what use case will it solve**:
We have seen cases where there is a long delay with creation of pods using security groups due to degraded performance of the reconciler in larger clusters with 2000+ SGP pods on controller restart. 
Increase the number of pod reconcilers by an appropriate amount to help with the performance issues.

PR #304, opened 2023-09-06
increasing node and pod worker number

haouc:
*Issue #, if available:*
#295 
*Description of changes:*
We are seeing pod queue latency in some extreme use case. Increasing the workers to help on performance.

### DIFF
--- a/controllers/core/node_controller.go
+++ b/controllers/core/node_controller.go
@@ -41,7 +41,7 @@ import (
 // on cache only a single go routine should be sufficient. Using more than
 // one routines to help high rate churn and larger nodes groups restarting
 // when the controller has to be restarted for various reasons.
-const MaxNodeConcurrentReconciles = 3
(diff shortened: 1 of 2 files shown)


sushrk:
General comment for my understanding, how did we choose these values?

haouc:
We are still in the process to tune the values. I don't think there is any standards or guidance from upstream on those values. Since they are pretty lightweight and increasing to 20 for pods and 10 for nodes are not large number, I think it is worth to make the change and watch how it works.

haouc:
After offline sync with @sushrk , we agree to increase the concurrency on node and pod controller by stages to minimize risks. If this change works stably, we will increase them to 10 for node and 20 for pod controller in the next release.

sushrk (approved):
LGTM, thanks Hao. 
We will be tuning this in the future as per performance metrics.

PR #309, opened 2023-09-29
Annotate pods with IPv6 details from branch ENI when available

jdn5126:
*Issue #, if available:*
N/A

*Description of changes:*
In order to support Security Groups for Pods in IPv6 clusters, the VPC Resource Controller must create trunk and branch ENIs with IPv6 addresses assigned. This happens already based on the subnet used in an IPv6 cluster, so VPC Resource Controller's next role is to annotate pods with branch ENI information.

This PR annotates pods with the IPv6 address and CIDR that is configured for the branch ENI, if there is one. The reason that IPv6 fields are added in-addition, instead of in-place is two-fold:
1. VPC Resource Controller must support older VPC CNI versions, so it cannot change existing JSON names.
2. If dual-stack is eventually supported, the pod will need IPv4 and IPv6 fields, so this change facilitates that.

I tested this code against an IPv4 and an IPv6 cluster to verify that it works. In an IPv4 cluster, the annotation looks as follows:
```
k get pod nginx-deployment-6c4d879f6c-7ktc5 -n sg-test -o jsonpath='{.metadata.annotations}'
{"vpc.amazonaws.com/pod-eni":"[{\"eniId\":\"eni-0f11cb35d7d5bf4d6\",\"ifAddress\":\"0e:ee:16:bd:41:31\",\"privateIp\":\"192.168.19.148\",\"ipv6Addr\":\"\",\"vlanId\":1,\"subnetCidr\":\"192.16
8.0.0/19\",\"subnetV6Cidr\":\"\"}]"}
```

In an IPv6 cluster, it looks as follows:
```
{"vpc.amazonaws.com/pod-eni":"[{\"eniId\":\"eni-0f11cb35d7d5bf4d6\",\"ifAddress\":\"0e:ee:16:bd:41:31\",\"privateIp\":\"192.168.19.148\",\"ipv6Addr\":\"2600::\",\"vlanId\":1,\"subnetCidr\":\"192.16
8.0.0/19\",\"subnetV6Cidr\":\"2600::/64\"}]"}
```

There are no backward-incompatible changes here, but I am still awaiting results from integration tests for further coverage.

### DIFF
--- a/pkg/aws/ec2/instance.go
+++ b/pkg/aws/ec2/instance.go
@@ -38,15 +38,18 @@ type ec2Instance struct {
 	// subnetId is the instance's subnet id
 	instanceSubnetID string
 	// instanceSubnetCidrBlock is the cidr block of the instance's subnet
-	instanceSubnetCidrBlock string
+	instanceSubnetCidrBlock   string
+	instanceSubnetV6CidrBlock string
 	// currentSubnetID can either point to the Subnet ID of the instance or subnet ID from the ENIConfig
 	currentSubnetID string
 	// currentSubnetCIDRBlock can either point to the Subnet CIDR block for instance subnet or subnet from ENIConfig
-	currentSubnetCIDRBlock string
+	currentSubnetCIDRBlock   string
+	currentSubnetV6CIDRBlock string
 	// currentInstanceSecurityGroups can either point to the primary network interface security groups or the security groups in ENIConfig
 	currentInstanceSecurityGroups []string
 	// subnetMask is the mask of the subnet CIDR block
-	subnetMask string
+	subnetMask   string
+	subnetV6Mask string
 	// deviceIndexes is the list of indexes used by the EC2 Instance
 	deviceIndexes []bool
 	// primaryENIGroups is the security group used by the primary network interface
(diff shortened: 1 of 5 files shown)


sushrk:
Changes LGTM overall, thanks @jdn5126! I have kicked off integration tests on Gitlab and will update the results.

pkg/aws/ec2/instance.go
```diff
@@ -110,8 +115,16 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
 			i.instanceSubnetID, i.instanceID)
 	}
 	i.instanceSubnetCidrBlock = *instanceSubnet.CidrBlock
-
 	i.subnetMask = strings.Split(i.instanceSubnetCidrBlock, "/")[1]
+	// Cache IPv6 CIDR block if one is present
+	for _, v6CidrBlock := range instanceSubnet.Ipv6CidrBlockAssociationSet {
+		if v6CidrBlock.Ipv6CidrBlock != nil {
+			i.instanceSubnetV6CidrBlock = *v6CidrBlock.Ipv6CidrBlock
+			i.subnetV6Mask = strings.Split(i.instanceSubnetV6CidrBlock, "/")[1]
```
haouc, line 123: Is instanceSubnetV6CidrBlock guaranteed to be split-able to at least two parts? Want to be make sure we won't have a panic in possible v6-not-supported cases.

haouc (approved):
lgtm

PR #315, opened 2023-10-02
increase pod ENI workers count

sushrk:
*Issue #, if available:*
Related to #295 

*Description of changes:*
We have seen poor performance through metrics in large clusters when creating pods using security group for pods. Bump up the number of pod ENI workers to improve performance. 

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -26,7 +26,7 @@ const (
 	WorkQueueDefaultMaxRetries = 5
 
 	// Default Configuration for Pod ENI resource type
-	PodENIDefaultWorker = 7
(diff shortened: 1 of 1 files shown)


haouc (approved):
lgtm, thanks.

pkg/config/loader.go
```diff
@@ -26,7 +26,7 @@ const (
 	WorkQueueDefaultMaxRetries = 5
 
 	// Default Configuration for Pod ENI resource type
-	PodENIDefaultWorker = 7
```
orsenthil, line 29: Just a note. Will we observe any other performance impact with this change?

PR #317, opened 2023-10-12
isolate node cleanup from worker queue

haouc:
*Issue #, if available:*

*Description of changes:*
Isolating the node based reconciliation/cleaning from worker queue to an independent one-off goroutine.

### DIFF
--- a/pkg/aws/errors/ec2_errors.go
+++ b/pkg/aws/errors/ec2_errors.go
@@ -1,0 +1,6 @@
+package errors
+
+const (
+	NotFoundAssociationID = "InvalidAssociationID.NotFound"
+	NotFoundInterfaceID   = "InvalidNetworkInterfaceID.NotFound"
+)
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/api"
@@ -33,6 +34,7 @@ import (
 	"github.com/samber/lo"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
@@ -63,6 +65,7 @@ type Manager interface {
 	AddNode(nodeName string) error
 	UpdateNode(nodeName string) error
 	DeleteNode(nodeName string) error
+	CheckNodeForLeakedENIs(nodeName string)
 }
 
 // AsyncOperation is operation on a node after the lock has been released.
@@ -113,6 +116,37 @@ func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager
 	)
 
 	return manager, worker.StartWorkerPool(manager.performAsyncOperation)
+}
+
+func (m *manager) CheckNodeForLeakedENIs(nodeName string) {
+	managedNode, found := m.GetNode(nodeName)
+	if !found {
+		m.Log.Info("Node manager couldn't find the node for reconciliation cleanup", "NodeName", nodeName)
+		return
+	}
+
+	// Only start a goroutine when need to
+	if time.Now().After(managedNode.GetNextReconciliationTime()) {
+		go func() {
+			if resourceProvider, found := m.resourceManager.GetResourceProvider(config.ResourceNamePodENI); found {
+				foundLeakedENI := resourceProvider.ReconcileNode(nodeName)
+				if foundLeakedENI {
+					managedNode.SetReconciliationInterval(node.NodeInitialCleanupInterval)
+				} else {
+					interval := wait.Jitter(managedNode.GetReconciliationInterval(), 5)
+					if interval > node.MaxNodeReconciliationInterval {
+						interval = node.MaxNodeReconciliationInterval
+					}
+					managedNode.SetReconciliationInterval(interval)
+				}
+				managedNode.SetNextReconciliationTime(time.Now().Add(managedNode.GetReconciliationInterval()))
+				m.Log.Info("reconciled cleanup node for leaking branch interfaces", "NodeName", nodeName, "NextInterval", managedNode.GetReconciliationInterval(), "NextReconciliationTime", managedNode.GetNextReconciliationTime())
+			} else {
+				// no SGP provider enabled
+				return
+			}
+		}()
+	}
 }
 
 // GetNode returns the node from in memory data store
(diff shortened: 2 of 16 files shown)


pkg/provider/branch/provider.go
```diff
@@ -270,27 +267,23 @@ func (b *branchENIProvider) UpdateResourceCapacity(instance ec2.EC2Instance) err
...
 // with the internal cache.
-func (b *branchENIProvider) ReconcileNode(nodeName string) (ctrl.Result, error) {
+func (b *branchENIProvider) ReconcileNode(nodeName string) bool {
 	trunkENI, isPresent := b.getTrunkFromCache(nodeName)
 	log := b.log.WithValues("node", nodeName)
 	if !isPresent {
-		log.Info("stopping the reconcile job")
-		return ctrl.Result{}, nil
+		log.Info("no trunk ENI is pointing to the given node", "NodeName", nodeName)
+		return true
```
sushrk, line 275: Why do we need to return `true` if no trunkENI found?

pkg/provider/branch/provider.go
```diff
@@ -270,27 +267,23 @@ func (b *branchENIProvider) UpdateResourceCapacity(instance ec2.EC2Instance) err
...
 	podList, err := b.apiWrapper.PodAPI.ListPods(nodeName)
 	if err != nil {
 		log.Error(err, "failed fo list pod")
-		return reconcileRequeueRequest, nil
-	}
-	err = trunkENI.Reconcile(podList.Items)
-	if err != nil {
-		b.log.Error(err, "failed to reconcile")
-		return reconcileRequeueRequest, nil
+		return true
```
sushrk, line 280: Same question as above.

pkg/provider/branch/provider.go
```diff
@@ -407,7 +400,11 @@ func (b *branchENIProvider) CreateAndAnnotateResources(podNamespace string, podN
 func (b *branchENIProvider) DeleteBranchUsedByPods(nodeName string, UID string) (ctrl.Result, error) {
 	trunkENI, isPresent := b.getTrunkFromCache(nodeName)
 	if !isPresent {
-		return ctrl.Result{}, fmt.Errorf("failed to find trunk ENI on the node %s", nodeName)
+		// trunk cache is local map with lock. it shouldn't return not found if trunk exists
```
sushrk, line 403: nit: did you mean we shouldn't return error if trunk not found?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -511,39 +545,6 @@ func (t *trunkENI) getBranchInterfacesUsedByPod(pod *v1.Pod) (eniDetails []*ENID
 	return
 }
 
-// GetBranchInterfacesFromEC2 returns the list of branch interfaces associated with the trunk ENI. This is not supported
-// yet
-func (t *trunkENI) GetBranchInterfacesFromEC2() (eniDetails []*ENIDetails, err error) {
```
sushrk, line 516: +1, nice!

pkg/node/manager/manager.go
```diff
@@ -19,20 +19,23 @@ import (
...
 	"sync"
+	"time"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/api"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/condition"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
 	rcHealthz "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/healthz"
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node"
+	nodePkg "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/node"
```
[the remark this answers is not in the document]
haouc: Thanks

pkg/node/manager/manager.go
```diff
@@ -115,6 +119,38 @@ func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager
...
+	if time.Now().After(node.GetNextReconciliationTime()) {
+		go func() {
+			var foundLeakedENI bool
+			if resourceProvider, found := m.resourceManager.GetResourceProvider(config.ResourceNamePodENI); found {
+				foundLeakedENI = resourceProvider.ReconcileNode(nodeName)
+			} else {
+				// no SGP provider enabled
+				return
+			}
+			if foundLeakedENI {
```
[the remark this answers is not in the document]
haouc: yes, it can be moved into `if`

pkg/provider/branch/provider.go
haouc: Returning `true` for these two uncommon errors is making sure we keep the shortest interval for next clean up since we don't know what exactly happen and want to retry again asap.

pkg/provider/branch/provider.go
haouc: This is not `error`
```
trunkENI, isPresent := b.getTrunkFromCache(nodeName)
```
The `isPresent` is a bool value to indicate if the key exists in the map.

pkg/provider/branch/provider.go
haouc: Returning `true` means if we reconcile on this node again we use the shortest interval.

pkg/provider/branch/provider.go
sushrk: I meant the comment looks confusing, seems it should be something like `do not return not found error if trunk does not exist`?

pkg/provider/branch/provider.go
sushrk: Got it, can we add a comment here?

pkg/provider/branch/provider.go
haouc: hmm, I see what you mean. Yeah I can make the comment better.

pkg/provider/branch/provider.go
haouc: Updated comments, thanks

pkg/aws/errors/ec2_errors.go
```diff
@@ -0,0 +1,6 @@
+package errors
+
+const (
+	NotFoundAssociationID = "InvalidAssociationID.NotFound"
```
sushrk, line 4: nit: seems this is not used anywhere, can we delete it?

sushrk (approved):

pkg/aws/errors/ec2_errors.go
haouc: This will be need for adding disassociating call.

pkg/node/manager/manager.go
```diff
@@ -115,6 +118,37 @@ func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager
...
+					interval := wait.Jitter(managedNode.GetReconciliationInterval(), 5)
+					if interval > node.MaxNodeReconciliationInterval {
+						interval = node.MaxNodeReconciliationInterval
+					}
+					managedNode.SetReconciliationInterval(interval)
+				}
+				managedNode.SetNextReconciliationTime(time.Now().Add(managedNode.GetReconciliationInterval()))
+				m.Log.Info("reconciled cleanup node for leaking branch interfaces", "NodeName", nodeName, "NextInterval", managedNode.GetReconciliationInterval(), "NextReconciliationTime", managedNode.GetNextReconciliationTime())
+			} else {
+				// no SGP provider enabled
```
jdn5126, line 145: In this case, should we set next reconciliation time to never? Otherwise, the `if` statement starting the routine above will always return true

haouc: Do you mean if the node is not using SGP feature? Yes, in this case, every time when a node event triggers a check a goroutine is created but immediately returns. I think there shouldn't be impact on performance since they are light and very short lifespan.

jdn5126: Yeah, discussed offline and this is fine for now. We can improve in the future with finalizers if needed

PR #330, opened 2023-11-03
adding govulnscheck to action

haouc:
*Issue #, if available:*

*Description of changes:*
We should enable go vulnerability check in actions.


sushrk (approved):

jdn5126:
nit: I think it would be better to copy the VPC CNI approach: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/deps.yml so that we have consistent checks between repos

Issue #306, opened 2023-09-29
Increasing "CoolDownPeriod" time from 30s to 60s


ssup2:
**What would you like to be enhanced**: Increasing [CoolDownPeriod](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/provider/branch/trunk/trunk.go#L40) time from 30s to 60s

**Why is the change needed and what use case will it solve**: Now vpc-rc use cool down period to propagate changed network rules when deleting branch ENI. But I think cool down period must be higher than 60s. Because linux kernel's default ARP Cache timeout is 60s. This means that when a new pod is created with a IP used by another deleted pod for less than 1 minute. Is there any reason why 30s is used?

jayanthvn:
I think the 30 seconds was used to be inline with AWS VPC CNI cooldown time and that was because kubeproxy takes upto 30 seconds to clean up the endpoints. During this time we don't want any other pod to get the same IP hence we add the IP to cooldown for 30 seconds.

haouc:
We are planning to make a change to be aligned with Linux settings. @sushrk will help fixing this issue. Thanks

jdn5126:
Closing as fixed by https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/342

PR #342, opened 2023-12-04
Add support to configure branch ENI cooldown period via configmap

sushrk:
*Issue #, if available:*
#306 

*Description of changes:*
Adding support to configure the cool down period via `amazon-vpc-cni` configmap using flag `branch-eni-cooldown`, in seconds. The minimum value for the cool period is 30s, if user updates configmap to a lower value, this will be overridden and set to 30s. 

Testing updated in comments. 


pkg/provider/branch/cooldown/cooldown.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/k8s"
+	"github.com/go-logr/logr"
+)
+
+// Global variable for CoolDownPeriod allows packages to Get and Set the coolDown period
+var coolDown *cooldown
+
+type cooldown struct {
+	mu sync.RWMutex
```
sushrk, line 31: This could also be done with `sync/atomic`. Did not realize earlier that `time.Duration` type is `int64`. 

Perhaps the code would be cleaner with `atomic`, would need to refactor and check this.

haouc:
@sushrk made some changes in the new commit. please let me know if you have comments on them. I will ask for reviews from the team.

haouc:
We allow the period configurable but force the lowest value as 30 seconds. Whenever the value is changed, the controller broadcast a node event to all nodes which ensure users the planned value being applied successfully. 
Testing some behaviors
Setting cooldown period to 120 seconds
```
Events:
  Type    Reason                          Age   From                     Message
  ----    ------                          ----  ----                     -------
  Normal  ControllerVersionNotice         11m   vpc-resource-controller  The node is managed by VPC resource controller version v1.4.2-9-gf21a4f9
  Normal  NodeTrunkInitiated              11m   vpc-resource-controller  The node has trunk interface initialized successfully
  Normal  BranchENICoolDownPeriodUpdated  41s   vpc-resource-controller  Branch ENI cool down period has been updated to 2m0s
```

Then change the value to 29 seconds which should be forced to 30 seconds although CM still has 29 seconds
```
apiVersion: v1
data:
  branch-eni-cooldown: "29"
  enable-network-policy-controller: "false"
  enable-windows-ipam: "false"
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"enable-network-policy-controller":"false","enable-windows-ipam":"false"},"kind":"ConfigMap","metadata":{"annotations":{},"labels":{"app.kubernetes.io/instance":"aws-vpc-cni","app.kubernetes.io/name":"aws-node","app.kubernetes.io/version":"v1.15.4","k8s-app":"aws-node"},"name":"amazon-vpc-cni","namespace":"kube-system"}}
  creationTimestamp: "2023-12-02T19:56:21Z"
  labels:
    app.kubernetes.io/instance: aws-vpc-cni
    app.kubernetes.io/name: aws-node
    app.kubernetes.io/version: v1.15.4
    k8s-app: aws-node
  name: amazon-vpc-cni
  namespace: kube-system
  resourceVersion: "989462"
  uid: 1c7920ed-5d6b-452d-8228-f8dee3c771c8
```
```
Events:
  Type    Reason                          Age    From                     Message
  ----    ------                          ----   ----                     -------
  Normal  ControllerVersionNotice         15m    vpc-resource-controller  The node is managed by VPC resource controller version v1.4.2-9-gf21a4f9
  Normal  NodeTrunkInitiated              15m    vpc-resource-controller  The node has trunk interface initialized successfully
  Normal  BranchENICoolDownPeriodUpdated  4m12s  vpc-resource-controller  Branch ENI cool down period has been updated to 2m0s
  Normal  BranchENICoolDownPeriodUpdated  37s    vpc-resource-controller  Branch ENI cool down period has been updated to 30s
```

controllers/core/configmap_controller.go
```diff
@@ -73,6 +74,17 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
+	// Check if branch ENI cooldown period is updated
+	curCoolDownPeriod := cooldown.GetCoolDown().GetCoolDownPeriod()
+	newCoolDownPeriod, err := cooldown.GetVpcCniConfigMapCoolDownPeriodOrDefault(r.K8sAPI, r.Log)
+	if curCoolDownPeriod != newCoolDownPeriod {
+		if err != nil {
```
achevuru, line 83: Shouldn't this `err` check come prior to `curCoolDownPeriod != newCoolDownPeriod`

achevuru: But why default? Shouldn't we instead retain the current value available in `curCoolDownPeriod` ?

pkg/provider/branch/cooldown/cooldown.go
```diff
@@ -0,0 +1,85 @@
...
+				return time.Second * time.Duration(coolDownPeriodInt), nil
+			}
+		}
+	}
+	// If configmap not found, or configmap data not found, or error in parsing coolDown period, return default coolDown period and error
+	return DefaultCoolDownPeriod, fmt.Errorf("failed to get coolDown period period:%v", err)
+}
+
+func (c *cooldown) GetCoolDownPeriod() time.Duration {
+	c.mu.Lock()
```
achevuru: Why do we need this? I mean this should be triggered only on a ConfigMap update and we only look at one particular ConfigMap in VPC RC...

controllers/core/configmap_controller.go
haouc: I think you are right. We should stay on what it was if failing on retrieving data from the CM.

pkg/provider/branch/cooldown/cooldown.go
haouc: We can revisit this later as an improvement on all locks.

pkg/provider/branch/cooldown/cooldown.go
achevuru: Are you saying we call a GET per ENI? Why not just update it once and use it as SET is responsible for updating the value? Even with that I'm not sure why we need a lock for GET.

controllers/core/configmap_controller.go
```diff
@@ -73,6 +76,24 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			utils.SendBroadcastNodeEvent(
+				r.K8sAPI,
+				utils.BranchENICoolDownUpdateReason,
+				fmt.Sprintf("Branch ENI cool down period has been updated to %s", cooldown.GetCoolDown().GetCoolDownPeriod()),
+				v1.EventTypeNormal,
+				r.Log,
+			)
+		}
+	} else {
+		r.Log.Error(err, "retrieving branch ENI cool down period from configmap failed and will keep the current cooldown time setting.")
```
achevuru, line 94: nit: `retrieving branch ENI cool down period from configmap failed and will keep the current cooldown time setting.` --> `retrieving branch ENI cool down period from ConfigMap (amazon-vpc-cni) failed. Will retain the current cooldown period value`

pkg/provider/branch/cooldown/cooldown.go
```diff
@@ -0,0 +1,89 @@
...
+			coolDownPeriodInt, err := strconv.Atoi(val)
+			if err != nil {
+				log.Error(err, "failed to parse branch ENI coolDown period", "cool down period", val)
+			} else {
+				return time.Second * time.Duration(coolDownPeriodInt), nil
+			}
+		}
+	}
+	// If configmap not found, or configmap data not found, or error in parsing coolDown period, return default coolDown period and error
+	return DefaultCoolDownPeriod, fmt.Errorf("failed to get coolDown period period:%v", err)
```
achevuru, line 73: Additional `period` towards the end in the log statement

pkg/provider/branch/cooldown/cooldown.go
achevuru: We're talking about a lock around read call isn't it. Let's discuss offline. At worst, we will have an ENI or 2 released back to the VPC pool using the prior cool down period value..

achevuru (approved):

achevuru (approved):

controllers/core/configmap_controller.go
```diff
@@ -73,6 +76,24 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
+	// Check if branch ENI cooldown period is updated
+	curCoolDownPeriod := cooldown.GetCoolDown().GetCoolDownPeriod()
```
jayanthvn, line 80: Is this defaulted to 30s since if first time get fails what will be the value?

pkg/provider/branch/cooldown/cooldown.go
```diff
@@ -0,0 +1,87 @@
...
+			coolDownPeriodInt, err := strconv.Atoi(val)
+			if err != nil {
+				log.Error(err, "failed to parse branch ENI coolDown period", "cool down period", val)
+			} else {
+				return time.Second * time.Duration(coolDownPeriodInt), nil
+			}
+		}
+	}
+	// If configmap not found, or configmap data not found, or error in parsing coolDown period, return default coolDown period and error
+	return DefaultCoolDownPeriod, fmt.Errorf("failed to get cool down period:%v", err)
```
jayanthvn, line 73: ok, I see here we return 60s..

pkg/provider/branch/cooldown/cooldown.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (c *cooldown) GetCoolDownPeriod() time.Duration {
+	if c.coolDownPeriod < 30*time.Second {
+		return MinimalCoolDownPeriod
```
jayanthvn, line 78: Do we bubble up this error to the customer i.e, we have set a lower limit of 30s? Since cx would assume they have reduced it below 30s..

PR #359, opened 2024-01-11
Update docs to add amazon-vpc-cni configmap options for SGPP

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Add `amazon-vpc-cni` configuration options for managing security groups for pods and update README. 
Also updated the SGPP workflow to include `CNINode` updates. 

### DIFF
--- a/docs/sgp/sgp_config_options.md
+++ b/docs/sgp/sgp_config_options.md
@@ -1,0 +1,16 @@
+# Configuration options for Security groups for pods
+
+Users are able to configure the controller functionality related to security group for pods by updating the `data` fields in EKS-managed configmap `amazon-vpc-cni`.
+
+* **branch-eni-cooldown**: Cooldown period for the branch ENIs, the period of time to wait before deleting the branch ENI for propagation of iptables rules for the deleted pod. The default cooldown period is 60s, and the minimum value for the cool period is 30s. If user updates configmap to a lower value than 30s, this will be overridden and set to 30s.
+
+Add `branch-eni-cooldown` field in the configmap to set the cooldown period, example:
+```
+apiVersion: v1
+data:
+  branch-eni-cooldown: "60"
+kind: ConfigMap
+metadata:
+  name: amazon-vpc-cni
+  namespace: kube-system
+```
--- a/docs/sgp/workflow.md
+++ b/docs/sgp/workflow.md
@@ -7,10 +7,10 @@ Security Group for Pods is supported only on Nitro Based Instances.
 
 ![New Nitro Based Node Create Event Diagram](../images/sgp-node-create.png)
 
-1. User adds a new supported Node or enables ENI Trunking with existing nodes present in the cluster.
-2. VPC CNI Plugin adds label `vpc.amazonaws.com/has-trunk-attached: false` if the Node has capacity to create 1 additional ENI.
-3. Controller watches for Node events and acts on node with the above label by creating a Trunk ENI. 
-4. Controller updates the resource capacity on this node to `vpc.amazonaws.com/pod-eni: # Supported Branch ENI`.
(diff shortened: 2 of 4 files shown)


docs/sgp/sgp_config_options.md
```diff
@@ -0,0 +1,27 @@
+# Configuration options for Security groups for pods
+
+Users are able to configure the controller functionality related to security group for pods by updating the `data` fields in EKS-managed configmap `amazon-vpc-cni`.
+
+* **branch-eni-cooldown**: Cooldown period for the branch ENIs, the period of time to wait before deleting the branch ENI for propagation of iptables rules for the deleted pod. The default cooldown period is 60s, and the minimum value for the cool period is 30s. If user updates configmap to a lower value than 30s, this will be overridden and set to 30s.
+
+Minimal configmap example:
```
haouc, line 7: "Minimal" sounds misleading to me.

docs/sgp/sgp_config_options.md
```diff
@@ -0,0 +1,27 @@
...
+  branch-eni-cooldown: "60"
+kind: ConfigMap
+metadata:
+  name: amazon-vpc-cni
+  namespace: kube-system
+```
+
+Note that if you set the cooldown period to be lower than the default, you would also need to adjust the kernel's ARP cache timeout to be equal to the new cooldown period. Else, this will result in intermittent DNS timeouts or packet drops due to re-use of the IP address of a recently terminated pod by a new pod before the kernel's ARP cache can be updated.
+
+Customers may use the below commands to tune the ARP cache timeout, make sure you update it for all interfaces on the node.
```
haouc, line 20: default will auto update the new interfaces. can we make this more clear? Thanks.

docs/sgp/sgp_config_options.md
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: v1
+data:
+  branch-eni-cooldown: "60"
+kind: ConfigMap
+metadata:
+  name: amazon-vpc-cni
+  namespace: kube-system
+```
+
+Note that if you set the cooldown period to be lower than the default, you would also need to adjust the kernel's ARP cache timeout to be equal to the new cooldown period. Else, this will result in intermittent DNS timeouts or packet drops due to re-use of the IP address of a recently terminated pod by a new pod before the kernel's ARP cache can be updated.
```
haouc, line 18: Let's not call out this strongly since only one report which hit this issue so far.

docs/sgp/workflow.md
```diff
@@ -7,10 +7,10 @@ Security Group for Pods is supported only on Nitro Based Instances.
 
 ![New Nitro Based Node Create Event Diagram](../images/sgp-node-create.png)
 
-1. User adds a new supported Node or enables ENI Trunking with existing nodes present in the cluster.
-2. VPC CNI Plugin adds label `vpc.amazonaws.com/has-trunk-attached: false` if the Node has capacity to create 1 additional ENI.
-3. Controller watches for Node events and acts on node with the above label by creating a Trunk ENI. 
-4. Controller updates the resource capacity on this node to `vpc.amazonaws.com/pod-eni: # Supported Branch ENI`.
+1. User adds a new supported node or enables ENI Trunking with existing nodes present in the cluster.
+2. VPC CNI Plugin updates EKS-managed CRD `CNINode <NODE-NAME>` to add feature `SecurityGroupsForPods` if the node has capacity to create 1 additional ENI.
```
haouc, line 11: Thanks for updating this. I think this was missed in previous doc update.

haouc (changes requested):

docs/sgp/sgp_config_options.md
sushrk: Thanks, I moved this section to the troubleshooting guide.

haouc (approved):
lgtm

Issue #370, opened 2024-02-19
Rename Windows warm/minimum IP targets


jdn5126:
**What would you like to be enhanced**:
I would like to rename the Windows warm/minimum IP target variables that are read from the `amazon-vpc-cni` ConfigMap: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/config/type.go#L76

The current names cause confusion with the environment variables in the `aws-node` Daemonset that are used for warm/minimum targets on Linux nodes. I propose that we rename these to `windows-*`.

Since we cannot do the switchover all at once, I propose the VPC Resource Controller starts accepting both names, which will allow the AWS VPC CNI chart to change its names to the `windows-*` values.

**Why is the change needed and what use case will it solve**:
This change would help avoid customer confusion when it comes to deploying the AWS VPC CNI chart.

haouc:
Thanks @jdn5126 , it makes sense to rename them to be aligned with Windows use case and avoid confusion with Linux use cases.

PR #371, opened 2024-02-20
rename windows flags

haouc:
*Issue #, if available:*
#370 
*Description of changes:*
Current Windows' flags name can cause confusion for Linux users. We need rename them with prefix `windows` without breaking current users' setting.

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -92,8 +92,17 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
 	}
 
 	warmIPTargetStr, foundWarmIP := vpcCniConfigMap.Data[WarmIPTarget]
+	if !foundWarmIP {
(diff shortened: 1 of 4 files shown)


pkg/config/loader.go
```diff
@@ -92,8 +92,17 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
 	}
 
 	warmIPTargetStr, foundWarmIP := vpcCniConfigMap.Data[WarmIPTarget]
+	if !foundWarmIP {
```
jdn5126, line 95: nit: can we just add a comment here that we are transitioning these variables to use the `windows-*` name to avoid confusion? And the older variables will eventually be deprecated?

jdn5126 (approved):
Thanks for adding this so quickly! The changes look good to me, just one nit about adding another comment

pkg/config/loader.go
haouc: Sure, updated this PR.

jdn5126 (approved):

PR #372, opened 2024-02-21
Call DisassociateTrunkInterface before deleting branch ENI

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Call `DisassociateTrunkInterface` to remove association between branch and trunk ENI before deleting the branch ENI as per recommendation by EC2 team. 

Tests done after adding `DisassociateTrunkInterface` permission in the control plane instance role: 
* Scale up and down pods using security group for pods and verify branch ENIs can be disassociated and deleted
* Terminate nodes and verify branch ENIs are deleted- Sometimes the association is removed when trunk ENI is deleted with EC2 instance termination, in which case branch ENI is still deleted
* Upgrade from old vpc-resource-controller which is missing association ID and verify branch ENIs can be deleted for backwards compatibility. 

SGPP tests:
```
Ran 19 of 23 Specs in 929.417 seconds
FAIL! -- 18 Passed | 1 Failed | 0 Pending | 4 Skipped
```
Failed at `[It] should have same number of CNINode no matter which mode`, unrelated to the changes in this PR as I was testing finalizers on CNINode and updated(downgraded) controller on my cluster. 

### DIFF
--- a/pkg/aws/ec2/api/helper.go
+++ b/pkg/aws/ec2/api/helper.go
@@ -620,8 +620,8 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
 }
 
 func (h *ec2APIHelper) DisassociateTrunkInterface(associationID *string) error {
-	ip := &ec2.DisassociateTrunkInterfaceInput{
+	input := &ec2.DisassociateTrunkInterfaceInput{
 		AssociationId: associationID,
 	}
-	return h.ec2Wrapper.DisassociateTrunkInterface(ip)
+	return h.ec2Wrapper.DisassociateTrunkInterface(input)
 }
--- a/pkg/aws/ec2/api/helper.go
+++ b/pkg/aws/ec2/api/helper.go
@@ -93,6 +93,7 @@ type EC2APIHelper interface {
 	GetInstanceDetails(instanceId *string) (*ec2.Instance, error)
 	AssignIPv4ResourcesAndWaitTillReady(eniID string, resourceType config.ResourceType, count int) ([]string, error)
 	UnassignIPv4Resources(eniID string, resourceType config.ResourceType, resources []string) error
+	DisassociateTrunkInterface(associationID *string) error
 }
 
 // CreateNetworkInterface creates a new network interface
@@ -616,4 +617,11 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
 		return err
 	}
 	return nil
+}
+
+func (h *ec2APIHelper) DisassociateTrunkInterface(associationID *string) error {
+	ip := &ec2.DisassociateTrunkInterfaceInput{
+		AssociationId: associationID,
+	}
+	return h.ec2Wrapper.DisassociateTrunkInterface(ip)
 }
--- a/pkg/provider/branch/trunk/trunk.go
+++ b/pkg/provider/branch/trunk/trunk.go
@@ -138,6 +138,8 @@ type ENIDetails struct {
 	deletionTimeStamp time.Time
 	// deleteRetryCount is the
 	deleteRetryCount int
+	// ID of association between branch and trunk ENI
+	AssociationID string
 }
 
 type IntrospectResponse struct {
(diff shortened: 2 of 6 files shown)


jdn5126 (approved):
Changes LGTM

pkg/aws/ec2/api/helper.go
```diff
@@ -617,3 +618,10 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
 	}
 	return nil
 }
+
+func (h *ec2APIHelper) DisassociateTrunkInterface(associationID *string) error {
+	ip := &ec2.DisassociateTrunkInterfaceInput{
```
haouc, line 623: nit: let's use another variable name. IP is commonly reserved for "IP address".

pkg/provider/branch/trunk/trunk.go
```diff
@@ -138,6 +138,8 @@ type ENIDetails struct {
 	deletionTimeStamp time.Time
 	// deleteRetryCount is the
 	deleteRetryCount int
+	// ID of association between branch and trunk ENI
+	AssociationID string
```
haouc, line 142: Let's add json tag for parsing as other fields.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -398,12 +400,14 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
 		newENIs = append(newENIs, newENI)
 
 		// Associate Branch to trunk
-		_, err = t.ec2ApiHelper.AssociateBranchToTrunk(&t.trunkENIId, nwInterface.NetworkInterfaceId, vlanID)
+		var associationOutput *awsEC2.AssociateTrunkInterfaceOutput
```
haouc, line 403: Can we remove this statement and directly assigning the returned value to the variable in the next line?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -499,7 +503,19 @@ func (t *trunkENI) DeleteCooledDownENIs() {
 
 // deleteENIs deletes the provided ENIs and frees up the Vlan assigned to then
 func (t *trunkENI) deleteENI(eniDetail *ENIDetails) (err error) {
-	// Delete Branch network interface first
+	// Disassociate branch ENI from trunk if association ID exists and delete branch network interface
+	if eniDetail.AssociationID != "" {
+		err = t.ec2ApiHelper.DisassociateTrunkInterface(&eniDetail.AssociationID)
+		if err != nil {
+			trunkENIOperationsErrCount.WithLabelValues("disassociate_trunk_error").Inc()
+			if !strings.Contains(err.Error(), ec2Errors.NotFoundAssociationID) {
```
haouc, line 511: Would be easier and ok if we just log the error instead of branching them to two conditions?

pkg/provider/branch/trunk/trunk.go
sushrk: The declaration here needs to be separate. We do not want to re-declare `err` here since we use it in the next check to see if creation/association of branch ENIs failed in which case we move all branch ENIs to the delete queue.

pkg/provider/branch/trunk/trunk.go
sushrk: This is following the workflow in `DeleteNetworkInterface` call below, I think we did not want to log error when association or the branch ENI is missing and instead only log as info for debugging. Any other failure will be logged as error.

pkg/provider/branch/trunk/trunk.go
haouc: I don't think that is how error works here. If create fails, error leads to break the loop already. If error reach here, it is nil. Here we check error returned from associate call. If it is not nil, we break from the loop. Actually we should avoid reusing err if possible.
err = createNetworkInterface()
if err != nil break
else
...
err = associateNetworkInterface()
if err != nil break
else
...

pkg/provider/branch/trunk/trunk.go
haouc: I think this is debatable if we need a branching for logging. But I am ok if you prefer to keep them as is.

haouc (approved):
lgtm. Thanks.

pkg/provider/branch/trunk/trunk.go
sushrk: Synced offline on this; this cannot be moved to short assignment as `err` will be re-declared and we miss catching issues with association. Resolving this.

PR #374, opened 2024-02-23
feat: Centralize leaked ENI cleanup

sushrk:
*Issue #, if available:*
N/A
*Description of changes:*
Centralize leaked ENI cleanup in the controller to delete leaked ENIs provisioned by the controller and VPC-CNI:
* Added finalizer in CNINode: when kubernetes deletes an object that has finalizers specified for it, the Kubernetes API marks the object for deletion by populating .metadata.deletionTimestamp, and returns a 202 status code (HTTP "Accepted"). The target object remains in a terminating state while the control plane, or other components, take the actions defined by the finalizers. After these actions are complete, the controller removes the relevant finalizers from the target object. When the metadata.finalizers field is empty, Kubernetes considers the deletion complete and deletes the object.
* Add a new field in Spec to add Tag: This is currently used to specify tags that need to added to network interfaces provisioned by the controller and VPC-CNI. VPC-CNI will read the CNINode to get the tags. 

The CNINode reconciler will add the finalizer and tag if missing on existing CNINode objects and handles the finalizer routine to clean up leaked resources.

* Re-create CNINode if deleted by user: The controller will re-create CNINode if deleted by user when node object is still present. 
If failed to create, node event will be published:
```
  Warning  CNINodeDeleted                  2m7s  vpc-resource-controller  CNINode was deleted and could not be re-created by controller
```
* Removed deleting branch ENIs on node termination
* Added metrics to count number of available ENIs found and number of leaked ENIs that failed to be deleted
* Added metrics to count the number of times controller needed to recreate CNINode
Tests: 
```
SGPP test suite

Ran 19 of 23 Specs in 923.345 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 4 Skipped
PASS

Ginkgo ran 1 suite in 15m25.59382642s
Test Suite Passed
```
* Verified CNINode finalizers and tags are added(if missing) to existing CNINode objects
* Verified finalizer routine is called when CNINode object is deleted at node termination
* Verified CNINode is re-created when deleted if the node object exists
* Verified new metrics are registered on the controller

Note, PR for the test suite is pending until the changes are merged in. 


main.go
```diff
@@ -332,11 +334,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := (&ec2API.ENICleaner{
+	cleaner := &eniCleaner.ClusterENICleaner{
```
haouc, line 337: nit: why do we not use the same pattern as other setup with manager?

pkg/node/manager/manager.go
```diff
@@ -248,6 +250,12 @@ func (m *manager) UpdateNode(nodeName string) error {
 
 	log := m.Log.WithValues("node name", nodeName, "request", "update")
 
+	// Create CNINode if not existing while updating node
```
haouc, line 253: Let's not recreate CNINode that is accidentally deleted by other users.

pkg/provider/branch/provider.go
```diff
@@ -76,7 +76,7 @@ var (
 
 	// NodeDeleteRequeueRequestDelay represents the time after which the resources belonging to a node will be cleaned
 	// up after receiving the actual node delete event.
-	NodeDeleteRequeueRequestDelay = time.Minute * 5
+	NodeDeleteRequeueRequestDelay = time.Minute * 1
```
haouc, line 79: What is the reason that you want to use a much short requeue delay?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -425,29 +435,20 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
-	// Delete all the branch used by the pod on this trunk ENI
+// DisassociateAllBranchENIs removes association between all the branch ENIs with the trunk. The branch ENIs will be deleted by the node termination finalizer
+// This is the last API call to the the Trunk ENI before it is removed from cache
+func (t *trunkENI) DisassociateAllBranchENIs() {
+	// Disassociate all the branch used by the pod on this trunk ENI, it will be cleaned up by the finalizer routine
 	// Since after this call, the trunk will be removed from cache. No need to clean up its branch map
 	for _, podENIs := range t.uidToBranchENIMap {
 		for _, eni := range podENIs {
-			err := t.deleteENI(eni)
+			err := t.ec2ApiHelper.DisassociateTrunkInterface(&eni.AssociationID)
```
haouc, line 445: Since this is called upon node deletion, we can afford some retries. Can you put the API call into retry.OnError()? EC2 APIs are eventually consistent and could fail intermittent for any reason. Having a 3/5 retries when disassociating branch ENI from trunk may better safeguard the consistency on our side.

haouc: Another option is using retry in the wrapper which calls EC2 but I personally prefer adding retry in the immediate caller.

pkg/provider/branch/trunk/trunk_test.go
```diff
@@ -700,7 +713,8 @@ func TestTrunkENI_InitTrunk(t *testing.T) {
 				f.mockEC2APIHelper.EXPECT().GetInstanceNetworkInterface(&InstanceId).Return([]*awsEc2.InstanceNetworkInterface{}, nil)
 				f.mockInstance.EXPECT().GetHighestUnusedDeviceIndex().Return(freeIndex, nil)
 				f.mockInstance.EXPECT().SubnetID().Return(SubnetId)
-				f.mockEC2APIHelper.EXPECT().CreateAndAttachNetworkInterface(&InstanceId, &SubnetId, SecurityGroups, nil,
+				// f.mockInstance.EXPECT().Name().Return(FakeInstance.Name())
```
haouc, line 716: nit: no longer in use? if yes, we can remove it.

pkg/aws/ec2/api/wrapper.go
```diff
@@ -360,10 +388,12 @@ func prometheusRegister() {
...
 			ec2APICallLatencies,
-			vpcCniLeakedENICleanupCnt,
-			vpcrcLeakedENICleanupCnt,
 			ec2DisassociateTrunkInterfaceCallCnt,
 			ec2DisassociateTrunkInterfaceErrCnt,
+			VpcCniAvailableENICnt,
+			VpcRcAvailableENICnt,
+			LeakedENIClusterCleanupCnt,
+			LeakedENINodeCleanupCnt,
 		)
```
haouc, line 397: Can you also add a counter for disassociate call?

pkg/aws/ec2/api/cleanup/node_cleanup.go
```diff
@@ -0,0 +1,91 @@
...
+			_, err := e.EC2Wrapper.DeleteNetworkInterface(&ec2.DeleteNetworkInterfaceInput{
+				NetworkInterfaceId: networkInterface.NetworkInterfaceId,
+			})
+			if err != nil {
+				// append err and continue, we will retry deletion in the next period/reconcile
+				errors = append(errors, fmt.Errorf("failed to delete leaked network interface %v:%v", *networkInterface.NetworkInterfaceId, err))
+				e.Log.Error(err, "failed to delete the leaked network interface",
+					"id", *networkInterface.NetworkInterfaceId)
+				// increment leaked ENIs count
+				api.LeakedENINodeCleanupCnt.Inc()
```
haouc, line 75: Doesn't sound right to me. For example, ENI A fails to be deleted twice, this counter counts the same ENI twice, doesn't it?

pkg/aws/ec2/api/cleanup/node_cleanup.go
```diff
@@ -0,0 +1,91 @@
...
+
+		nextToken = describeNetworkInterfaceOp.NextToken
+		if nextToken == nil {
+			break
+		}
+
+	}
+	if len(errors) > 0 {
+		return fmt.Errorf("failed to delete atleast one leaked network interface, will be retried: %v", errors)
+	}
```
haouc, line 89: Seems rudundant error log for me since you already log error for every time they happen in the for loop.

pkg/aws/ec2/api/cleanup/eni/eni_cleanup.go
```diff
@@ -100,79 +103,80 @@ func (e *ENICleaner) Start(ctx context.Context) error {
...
+func (e *ClusterENICleaner) DeleteLeakedResources() error {
+	var errors []error
+	var nextToken *string
+	eniFilters := e.GetENITagFilters()
 	availableENIs := make(map[string]struct{})
-
+	api.VpcCniAvailableENICnt.Set(0)
+	api.VpcRcAvailableENICnt.Set(0)
+	api.LeakedENIClusterCleanupCnt.Set(0)
 	for {
```
haouc, line 114: Do we need a jitter in the loop to flatten possible API requests peak on EC2? Since this is in a cleanup process, a small delay between every request is ok and it could help protecting the account quota.

pkg/aws/ec2/api/cleanup/node_cleanup.go
```diff
@@ -0,0 +1,91 @@
...
+		{
+			Name:   aws.String("tag:" + config.NetworkInterfaceNodenameKey),
+			Values: []*string{aws.String(n.NodeName)},
+		},
+	}...)
+	return nodeFilter
+}
+
+// DeleteLeakedResources deletes all available network interfaces on the node at node termination
+func (e *NodeTerminationCleaner) DeleteLeakedResources() error {
```
haouc, line 48: This seems like a copy of eni_cleanup. Can you check if you can reuse the function with some arguments for their specific counters and tagset?

controllers/vpcresources/cninode_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	r.finalizers = finalizer.NewFinalizers()
+	err := r.finalizers.Register(config.NodeTerminationFinalizer, &nodeTerminationFinalizer{
+		EC2Wrapper: r.EC2Wrapper,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to register the node termination finalizer on CNINode: %v", err)
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.CNINode{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
```
haouc, line 110: Any concern on performance if we only use 1?

controllers/vpcresources/cninode_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		if err := r.Client.Patch(ctx, newCNINode, client.MergeFrom(cniNode)); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to patch CNINode %v", err)
+		}
+	}
+
+	// Finalize will add the finalizer if it does not exist and object is not being delete, and will remove the finalizer if it exists
+	// and object is being deleted. Also sets result.Updated to true so we need to update the object
+	finalizationRes, err := r.finalizers.Finalize(ctx, cniNode)
```
haouc, line 87: I think we should add deletion protection for CNINode here. If the finalizer is updated (deletion timestamp is added), we can copy and recreate the CNINode iif the node is still existing (which means accidently or incorrectly deleted the CNINode object). Then we should have all data from the original object and create a copy soon. Something like
1, deletion timestamp is found
2, check if the node is deleted
3, the node is gone --> continue finalizing the CNINode
4, the node is still existing --> giving up finalizing and copying and recreating the CNINode

We have to implementation the protection until upstream [KEP for in-use protection](https://github.com/kubernetes/enhancements/pull/2840) is implemented and graduated.

pkg/aws/ec2/api/wrapper.go
sushrk: Metrics for disassociate call is already added in a previous PR #372

controllers/vpcresources/cninode_controller.go
sushrk: Updated to use the same number as max concurrent node reconciles

pkg/aws/ec2/api/cleanup/eni/eni_cleanup.go
sushrk: Fixed by using the paginated API call with jitter

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			r.Log.Info("CNINode is deleted", "CNINode", req.NamespacedName)
+		}
+		// Ignore not found error; CNINode will be re-created on node reconcile if the node is not terminated
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Add cluster name tag if it does not exist
+	clusterNameTagKey := fmt.Sprintf(config.ClusterNameTagKeyFormat, r.ClusterName)
+	val, ok := cniNode.Spec.Tags[clusterNameTagKey]
+	if !ok || val != config.ClusterNameTagValue {
```
haouc, line 67: I am wonder what would be the case `val != config.ClusterNameTagValue` could occur? If this happens, that should mean another user set this up then if this is safe to reset them at line 70 or 73?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+
+		fmt.Println("patching CNINode to add tags")
+		if err := r.Client.Patch(ctx, cniNodeCopy, client.MergeFrom(cniNode)); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to patch CNINode %v", err)
+		}
+
+		// Re-fetch the CNINode after updating the Tag so we have the latest state of the CRD before updating it again in later operations
+		if err := r.Client.Get(ctx, req.NamespacedName, cniNode); err != nil {
+			return ctrl.Result{}, err
+		}
```
haouc, line 85: I am not sure if I follow what this if check wants to do here? You `Get` the object and `Patch` it. Why do you need `Get` it again here?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// Add the finalizer if it does not exist, so we can define cleanup routine which should be run before CNINode can be deleted
+	if !controllerutil.ContainsFinalizer(cniNode, config.NodeTerminationFinalizer) {
+		r.Log.Info("adding finalizer for CNINode", "cniNode", cniNode.Name)
+		if ok := controllerutil.AddFinalizer(cniNode, config.NodeTerminationFinalizer); !ok {
+			r.Log.Info("failed to add finalizer for CNINode, requeuing the request", "cniNode", cniNode.Name)
```
haouc, line 92: Is this qualified as an Error? Especially when we need depend on the finalizer to clean up associated resources.

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			return ctrl.Result{Requeue: true}, nil
+		}
+		if err := r.Client.Update(ctx, cniNode); err != nil {
```
haouc, line 95: Why do you use `Update` instead of `Patch` for this ops?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+					// Return err if failed to delete leaked ENIs on node so it can be retried
+					if err := cleaner.DeleteLeakedResources(); err != nil {
+						return ctrl.Result{}, err
+					}
+					// Remove finalizer and update CNINode to continue deletion
+					if ok := controllerutil.RemoveFinalizer(cniNode, config.NodeTerminationFinalizer); !ok {
+						r.Log.Info("failed to remove finalizer for CNINode, requeuing the request", "cniNode", cniNode.Name)
+						return ctrl.Result{Requeue: true}, nil
+					}
+					if err := r.Client.Update(ctx, cniNode); err != nil {
```
haouc, line 128: Same here, why do you use `Update` instead of `Patch` here?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+					r.Log.Info("failed to get the node object in CNINode reconciliation, will retry")
+					// Requeue request so it can be retried
+					return ctrl.Result{}, err
+				}
+			} else {
+				// node exists, do not run the cleanup routine(periodic cleanup routine will anyway delete leaked ENIs), remove the finalizer
+				// and let object be deleted and re-create similar object
+
+				// Create a copy without deletion timestamp for creation
+				newCNINode := &v1alpha1.CNINode{
```
haouc, line 142: Can you use `newCNINode := cniNode.DeepCopy()`?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+						// TODO: should we include finalizers at object creation or let finalizer add it on Create event?
+						Finalizers: cniNode.Finalizers,
+					},
+					Spec: cniNode.Spec,
+				}
+				if ok := controllerutil.RemoveFinalizer(cniNode, config.NodeTerminationFinalizer); !ok {
+					r.Log.Info("failed to remove finalizer for CNINode, requeuing the request", "cniNode", cniNode.Name)
+					return ctrl.Result{Requeue: true}, nil
+				}
+				if err := r.Client.Update(ctx, cniNode); err != nil {
```
haouc, line 156: Would `Patch` be safer?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+				}
+
+				// Re-fetch CNINode to verify it is deleted before re-creating it as we use the same name
+				// TODO: do we need to wait till CNINode deletion here?
+				if err := r.Client.Get(ctx, req.NamespacedName, cniNode); err != nil && errors.IsNotFound(err) {
+					if err = retry.OnError(retry.DefaultBackoff, func(error) bool {
+						if err != nil {
+							return true
+						}
+						return false
```
haouc, line 168: This is retriable func which is the definition for should we retry. I think in this case, you just need always `return true` to retry on all errors till the backoff completes.

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+					return ctrl.Result{Requeue: true}, nil
+				}
+				if err := r.Client.Update(ctx, cniNode); err != nil {
+					// Return err so update can be retried
+					return ctrl.Result{}, err
+				}
+
+				// Re-fetch CNINode to verify it is deleted before re-creating it as we use the same name
+				// TODO: do we need to wait till CNINode deletion here?
+				if err := r.Client.Get(ctx, req.NamespacedName, cniNode); err != nil && errors.IsNotFound(err) {
```
haouc, line 163: I think we should retry on this. Otherwise we may end up as 
- the old cniNode is pending for deleteion
- the old cniNode is deleted
- the reconciler gets the delete event
- the reconciler couldn't find the cniNode and return ok
- the resource is not protected

pkg/config/type.go
```diff
@@ -51,6 +51,8 @@ const (
 	OSWindows = "windows"
 	// OSLinux is the the linux Operating System
 	OSLinux = "linux"
+	// Node termination finalizer on CNINode CRD
+	NodeTerminationFinalizer = "node.k8s.aws.com/termination"
```
haouc, line 55: Let's discuss on the naming since this will be very hard to change in the future if we have to.

haouc: Probably `networking.k8s.aws/cninode`? For example, the finalizer key used by network policy controller: https://github.com/aws/amazon-network-policy-controller-k8s/blob/f34356594e417edac53b6532ec88f305b3bbb128/internal/controllers/policy_controller.go#L45

main.go
```diff
@@ -141,6 +144,7 @@ func main() {
 	flag.BoolVar(&enableWindowsPrefixDelegation, "enable-windows-prefix-delegation", false,
 		"Enable the feature flag for Windows prefix delegation")
 	flag.StringVar(&region, "aws-region", "", "The aws region of the k8s cluster")
+	flag.StringVar(&vpcID, "vpc-id", "", "The vpc-id where EKS cluster is deployed")
```
haouc, line 147: nit: VPC ID

pkg/provider/branch/trunk/trunk.go
```diff
@@ -425,29 +437,37 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
+						return false
+					}
+					return true
+				}, func() error {
+					return t.ec2ApiHelper.DisassociateTrunkInterface(&eni.AssociationID)
+				},
+			)
+			if err != nil && !strings.Contains(err.Error(), ec2Errors.NotFoundAssociationID) {
+				// Just log, if the ENI disassociation fails, it will be force deleted
+				t.log.Error(err, "failed to disassociate eni", "eni id", eni.ID)
```
haouc, line 467: Should we also count this new API call success and error in metrics? Also I think latency metrics may help as well at least after we just add this call.

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,194 @@
...
+						return r.Client.Create(ctx, newCNINode)
+					}); err != nil {
+						// The CNINode is deleted at this point, we cannot re-create the same object
+						// raise node object to publish warning that CNINode is deleted
+						utils.SendNodeEventWithNodeName(r.K8sAPI, node.Name, utils.CNINodeDeleted,
+							fmt.Sprint("CNINode was deleted and could not be re-created by controller"), v1.EventTypeWarning, r.Log)
+
+					}
+					// Requeue request to ensure the state of the object
+					return ctrl.Result{Requeue: true}, nil
```
haouc, line 180: Can you add a metrics on recreating the resource for protection? I want to know how many customers may trigger this behavior. I am not really preferring to requeue this after retries on a supposed to be reliable API call. Having it for now is ok to me. I would like to remove the requeueing after we have metrics data from customers and prove the behavior should be very rare and the retry succeed.

pkg/provider/branch/trunk/trunk.go
sushrk: Both error and latency metrics was already added in the previous PR [here](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/eni-cleanup/pkg/aws/ec2/api/wrapper.go#L820)

controllers/crds/cninode_controller.go
sushrk: Since currently it is updating both Tag and Finalizer field on the same reconcile, we need to re-fetch the latest version of the object else it fails with `object has been modified, please apply your changes to the latest version and try again`. 

Do you mean we just return after patching the tag so we add the finalizer on next reconcile triggered for CNINode? That should work too, but it takes two reconciles to get to desired state

controllers/crds/cninode_controller.go
sushrk: Hm, `AddFinalizer` and `RemoveFinalizer` updates the same object, but patch might be more efficient here, will update this.

controllers/crds/cninode_controller.go
sushrk: Yeah, this is not directly settable by clients per documentation: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata

controllers/crds/cninode_controller.go
```diff
@@ -57,128 +88,99 @@ func (r *CNINodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-					// Requeue request to ensure the state of the object
-					return ctrl.Result{Requeue: true}, nil
-				}
+			// Create a copy without deletion timestamp for creation
+			newCNINode := &v1alpha1.CNINode{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            cniNode.Name,
+					Namespace:       "",
+					OwnerReferences: cniNode.OwnerReferences,
+					// TODO: should we include finalizers at object creation or let controller patch it on Create/Update event?
```
haouc, line 157: I agree. We should create the CRs with finalizer. We can do this as a follow-up.

controllers/crds/cninode_controller.go
```diff
@@ -202,10 +202,10 @@ func (r *CNINodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *CNINodeReconciler) waitTillCNINodeDeleted(ctx context.Context, nameSpacedCNINode types.NamespacedName) error {
 	oldCNINode := &v1alpha1.CNINode{}
 
-	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*10)
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*3)
 	defer cancel()
 
-	return wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextCancel(ctx, 500*time.Millisecond, true, func(ctx context.Context) (bool, error) {
 		if err := r.Client.Get(ctx, nameSpacedCNINode, oldCNINode); err != nil {
```
haouc, line 209: nit: would this be more readable?

```
err := r.Client.Get(ctx, nameSpacedCNINode, oldCNINode)
if err == nil || !errors.IsNotFound(err) {
  return false, err
}
return true, nil
```

controllers/crds/cninode_controller.go
```diff
@@ -202,10 +202,10 @@ func (r *CNINodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *CNINodeReconciler) waitTillCNINodeDeleted(ctx context.Context, nameSpacedCNINode types.NamespacedName) error {
 	oldCNINode := &v1alpha1.CNINode{}
 
-	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*10)
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*3)
 	defer cancel()
 
-	return wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextCancel(ctx, 500*time.Millisecond, true, func(ctx context.Context) (bool, error) {
```
haouc, line 208: nit: Why don't you use simpler option (`PollUntilContextTimeout`)?

controllers/crds/cninode_controller.go
```diff
@@ -0,0 +1,227 @@
...
+
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*3)
+	defer cancel()
+
+	return wait.PollUntilContextCancel(ctx, 500*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+		if err := r.Client.Get(ctx, nameSpacedCNINode, oldCNINode); err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
+			} else {
+				return false, err
```
haouc, line 213: I think this is not working regarding
> PollUntilContextCancel tries a condition func until it returns true, an error, or the context is cancelled or hits a deadline.

Returning an error here will quit the polling

controllers/crds/cninode_controller.go
sushrk: My bad, I misread that `PollUntilContextTimeout` was deprecated, have updated this in the latest commit.

sushrk:
Had a sync offline, and addressed all the comments in subsequent commits. PTAL, thanks.

haouc (approved):
lgtm

controllers/crds/cninode_controller.go
haouc: Close this conversation since the question has been addressed in later commits.

PR #375, opened 2024-02-28
fix: paginate DescribeNetworkInterfaces with deep filters

### LINKED ISSUE #188: Unpaginated calls to DescribeNetworkInterfaces getting blocked by EC2
**Describe the Bug**:

We've been in the process of rolling out security groups for pods across various EKS clusters in multiple AWS accounts that we own. Recently, we've attempted to do the same in one of our "larger" AWS accounts, and noticed that pods would hang indefinitely waiting for an ENI to be attached by vpc-resource-controller.

We recently raised the ENI limits on this account to a very large value, and was told that unpaginated calls to `DescribeNetworkInterfaces` would be blocked as a result, and it seems that this is impacting vpc-resource-controller as you can see here in this CloudTrail event:

```json
{
    "eventVersion": "1.08",
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": "<REDACTED>:amazon-vpc-resource-controller-k8s",
        "arn": "arn:aws:sts::<REDACTED>:assumed-role/<REDACTED>/amazon-vpc-resource-controller-k8s",
        "accountId": "<REDACTED>",
        "sessionContext": {
            "sessionIssuer": {
                "type": "Role",
                "principalId": "<REDACTED>",
                "arn": "arn:aws:iam::<REDACTED>:role/<REDACTED>-control-20210716225641340500000009",
                "accountId": "<REDACTED>",
                "userName": "<REDACTED>-control-20210716225641340500000009"
            },
            "webIdFederationData": {},
            "attributes": {
                "creationDate": "2023-03-14T21:23:53Z",
                "mfaAuthenticated": "false"
            }
        },
        "invokedBy": "eks.amazonaws.com"
    },
    "eventTime": "2023-03-14T21:35:11Z",
    "eventSource": "ec2.amazonaws.com",
    "eventName": "DescribeNetworkInterfaces",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "eks.amazonaws.com",
    "userAgent": "eks.amazonaws.com",
    "errorCode": "Client.OperationNotPermitted",
    "errorMessage": "This operation is not permitted.",
    "requestParameters": {
        "networkInterfaceIdSet": {},
        "filterSet": {
            "items": [
                {
                    "name": "tag:vpcresources.k8s.aws/trunk-eni-id",
                    "valueSet": {
                        "items": [
                            {
                                "value": "eni-019dd4ee9f4aa06c8"
                            }
                        ]
                    }
                }
            ]
        }
    },
    "responseElements": null,
    "requestID": "8fd9723c-c4fa-4dd2-82be-47d8f328bf8e",
    "eventID": "fd8c44a0-b04c-430d-9e77-8c478b0f2de3",
    "readOnly": true,
    "eventType": "AwsApiCall",
    "managementEvent": true,
    "recipientAccountId": "<REDACTED>",
    "eventCategory": "Management"
}
```

For AWS engineers, our case # is 12229078111

**Observed Behavior**:

Calls to `DescribeNetworkInterfaces` are blocked by EC2 with a `Client.OperationNotPermitted` error code

**Expected Behavior**:

SGP should work even if unpaginated calls are blocked.

**How to reproduce it (as minimally and precisely as possible)**:

Increase ENI limits in an AWS account high enough that EC2 blocks unpaginated calls to `DescribeNetworkInterfaces`

**Additional Context**:

**Environment**:
- Kubernetes version (use `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.16-eks-ffeb93d", GitCommit:"52e500d139bdef42fbc4540c357f0565c7867a81", GitTreeState:"clean", BuildDate:"2022-11-29T18:41:42Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}`
- CNI Version 1.11.4
- OS (Linux/Windows): Amazon Linux (EKS AMI)

sushrk:
*Issue #, if available:*
Fixes #188 
*Description of changes:*
Paginate DescribeNetworkInterfaces API call by specifying a deep filter like vpc-id and subnet-id as unpaginated API call causes the operation to fail when there is a large number of network interfaces. 
https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-network-interfaces.html

Unpaginated calls were used in the periodic leaked ENI cleanup routine and during trunk initialization to fetch all branch ENIs on the node. These are now paginated to use vpc-id and subnet-id respectively, and also added jitter to avoid EC2 API throttling and increased call volume to address earlier issue with pagination.

Testing:
Security group for pods test succeeded:
```
Ran 19 of 23 Specs in 975.331 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 4 Skipped
PASS

Ginkgo ran 1 suite in 16m18.325964387s
Test Suite Passed
```

Manual tests:
* Node scale up/down
* Pods using security groups scale up/down
* Created 500+ ENIs in same VPC to confirm call volume has not increased drastically
```
ec2_describe_network_interfaces_pages_api_call_count 4
ec2_describe_network_interfaces_pages_api_err_count 0
```
* Soak test with 100 nodes and scaling up pods using to security group from 0 to 1000 every 15mins. Did not observe a drastic increase in the API call volume. 

Release notes:

```
Paginate DescribeNetworkInterfaces API call
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -107,6 +107,7 @@ func main() {
 	var healthCheckTimeout int
 	var enableWindowsPrefixDelegation bool
 	var region string
+	var vpcID string
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080",
 		"The address the metric endpoint binds to.")
@@ -141,6 +142,7 @@ func main() {
 	flag.BoolVar(&enableWindowsPrefixDelegation, "enable-windows-prefix-delegation", false,
 		"Enable the feature flag for Windows prefix delegation")
 	flag.StringVar(&region, "aws-region", "", "The aws region of the k8s cluster")
+	flag.StringVar(&vpcID, "vpc-id", "", "The vpc-id where EKS cluster is deployed")
 
 	flag.Parse()
 
@@ -180,6 +182,11 @@ func main() {
 
 	if clusterName == "" {
 		setupLog.Error(fmt.Errorf("cluster-name is a required parameter"), "unable to start the controller")
+		os.Exit(1)
+	}
+
+	if vpcID == "" {
+		setupLog.Error(fmt.Errorf("vpc-id is a required parameter"), "unable to start the controller")
 		os.Exit(1)
 	}
 
@@ -336,6 +343,7 @@ func main() {
 		EC2Wrapper:  ec2Wrapper,
 		ClusterName: clusterName,
 		Log:         ctrl.Log.WithName("eni cleaner"),
+		VPCID:       vpcID,
 	}).SetupWithManager(ctx, mgr, healthzHandler); err != nil {
 		setupLog.Error(err, "unable to start eni cleaner")
 		os.Exit(1)
--- a/pkg/aws/ec2/api/eni_cleanup.go
+++ b/pkg/aws/ec2/api/eni_cleanup.go
@@ -43,16 +43,22 @@ type ENICleaner struct {
 }
 
 var (
-	vpcCniLeakedENICleanupCnt = prometheus.NewCounter(
-		prometheus.CounterOpts{
-			Name: "vpc_cni_created_leaked_eni_cleanup_count",
-			Help: "The number of leaked ENIs created by VPC-CNI that is cleaned up by the controller",
+	vpccniAvailableENICnt = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "vpc_cni_created_available_eni_count",
+			Help: "The number of available ENIs created by VPC-CNI that controller will try to delete in each cleanup cycle",
 		},
 	)
-	vpcrcLeakedENICleanupCnt = prometheus.NewCounter(
-		prometheus.CounterOpts{
-			Name: "vpc_rc_created_leaked_eni_cleanup_count",
-			Help: "The number of leaked ENIs created by VPC-RC that is cleaned up by the controller",
+	vpcrcAvailableENICnt = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "vpc_rc_created_available_eni_count",
+			Help: "The number of available ENIs created by VPC-RC that controller will try to delete in each cleanup cycle",
+		},
+	)
+	leakedENICnt = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "leaked_eni_count",
+			Help: "The number of available ENIs that failed to be deleted by the controller in each cleanup cycle",
 		},
 	)
 )
@@ -102,6 +108,9 @@ func (e *ENICleaner) Start(ctx context.Context) error {
 // interval between cycle 1 and 2 and hence can be safely deleted. And we can also conclude that Interface 1 was
 // created but not attached at the the time when 1st cycle ran and hence it should not be deleted.
 func (e *ENICleaner) cleanUpAvailableENIs() {
+	vpccniAvailableENICnt.Set(0)
+	vpcrcAvailableENICnt.Set(0)
+	leakedENICnt.Set(0)
 	describeNetworkInterfaceIp := &ec2.DescribeNetworkInterfacesInput{
 		Filters: []*ec2.Filter{
 			{
@@ -140,12 +149,13 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
 			}); tagIdx != -1 {
 				switch *networkInterface.TagSet[tagIdx].Value {
 				case config.NetworkInterfaceOwnerTagValue:
-					vpcrcLeakedENICleanupCnt.Inc()
+					vpcrcAvailableENICnt.Inc()
 				case config.NetworkInterfaceOwnerVPCCNITagValue:
-					vpcCniLeakedENICleanupCnt.Inc()
+					vpccniAvailableENICnt.Inc()
 				default:
-					// We will not hit this case as we only filter for above two tag values, adding it for any future use cases
-					e.Log.Info("found available ENI not created by VPC-CNI/VPC-RC")
+					// We should not hit this case as we only filter for relevant tag values, log error and continue if unexpected ENIs found
+					e.Log.Error(fmt.Errorf("found available ENI not created by VPC-CNI/VPC-RC"), "eni id", *networkInterface.NetworkInterfaceId)
+					continue
 				}
 			}
 
@@ -155,6 +165,7 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
 				NetworkInterfaceId: networkInterface.NetworkInterfaceId,
 			})
 			if err != nil {
+				leakedENICnt.Inc()
 				// Log and continue, if the ENI is still present it will be cleaned up in next 2 cycles
 				e.Log.Error(err, "failed to delete the dangling network interface",
 					"id", *networkInterface.NetworkInterfaceId)
--- a/pkg/aws/ec2/api/eni_cleanup.go
+++ b/pkg/aws/ec2/api/eni_cleanup.go
@@ -108,9 +108,9 @@ func (e *ENICleaner) Start(ctx context.Context) error {
 // interval between cycle 1 and 2 and hence can be safely deleted. And we can also conclude that Interface 1 was
 // created but not attached at the the time when 1st cycle ran and hence it should not be deleted.
 func (e *ENICleaner) cleanUpAvailableENIs() {
-	vpccniAvailableENICnt.Set(0)
-	vpcrcAvailableENICnt.Set(0)
-	leakedENICnt.Set(0)
+	vpcrcAvailableCount := 0
+	vpccniAvailableCount := 0
+	leakedENICount := 0
 	describeNetworkInterfaceIp := &ec2.DescribeNetworkInterfacesInput{
 		Filters: []*ec2.Filter{
 			{
@@ -137,7 +137,7 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
 
 	networkInterfaces, err := e.EC2Wrapper.DescribeNetworkInterfacesPages(describeNetworkInterfaceIp)
 	if err != nil {
-		e.Log.Error(err, "failed to describe network interfaces, will retry")
+		e.Log.Error(err, "failed to describe network interfaces, cleanup will be retried in next cycle")
 		return
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/aws/ec2/api/eni_cleanup.go
```diff
@@ -116,61 +117,57 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
...
+				return *tag.Key == config.NetworkInterfaceOwnerTagKey
+			}); tagIdx != -1 {
+				switch *networkInterface.TagSet[tagIdx].Value {
+				case config.NetworkInterfaceOwnerTagValue:
+					vpcrcLeakedENICleanupCnt.Inc()
+				case config.NetworkInterfaceOwnerVPCCNITagValue:
+					vpcCniLeakedENICleanupCnt.Inc()
+				default:
+					// We will not hit this case as we only filter for above two tag values, adding it for any future use cases
+					e.Log.Info("found available ENI not created by VPC-CNI/VPC-RC")
```
haouc, line 148: I think we should log this as error if we get unexpected ENIs being returned.

pkg/aws/ec2/api/wrapper.go
```diff
@@ -639,6 +656,38 @@ func (e *ec2Wrapper) DescribeNetworkInterfaces(input *ec2.DescribeNetworkInterfa
 	return describeNetworkInterfacesOutput, err
 }
 
+// DescribeNetworkInterfacesPages returns network interfaces that match the filters specified in the input with MaxResult set to 1000(max value)
+// This API is used during periodic ENI cleanup routine and trunk initialization to list all network interfaces that match the given filters (vpc-id or subnet-id, and tag)
+// Only required fields, network interface ID and tag set, is populated to avoid consuming extra memory
+func (e *ec2Wrapper) DescribeNetworkInterfacesPages(input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
+	var networkInterfaces []*ec2.NetworkInterface
+	input.MaxResults = aws.Int64(1000)
```
haouc, line 664: Can you make the 1000 as constant?

pkg/aws/ec2/api/wrapper.go
```diff
@@ -639,6 +656,38 @@ func (e *ec2Wrapper) DescribeNetworkInterfaces(input *ec2.DescribeNetworkInterfa
 	return describeNetworkInterfacesOutput, err
 }
 
+// DescribeNetworkInterfacesPages returns network interfaces that match the filters specified in the input with MaxResult set to 1000(max value)
+// This API is used during periodic ENI cleanup routine and trunk initialization to list all network interfaces that match the given filters (vpc-id or subnet-id, and tag)
+// Only required fields, network interface ID and tag set, is populated to avoid consuming extra memory
+func (e *ec2Wrapper) DescribeNetworkInterfacesPages(input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
```
haouc, line 662: Why do you need retrun a slice of NetworkInterface instead of a slice of string (ENI ID)? Since I see in `append` you just populate ENI ID and TagSet to every NetworkInterface, I am not sure why you need `TagSet`?

pkg/aws/ec2/api/eni_cleanup.go
```diff
@@ -116,61 +117,57 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
...
-					continue
+	for _, networkInterface := range networkInterfaces {
+		if _, exists := e.availableENIs[*networkInterface.NetworkInterfaceId]; exists {
+			// Increment promethues metrics for number of leaked ENIs cleaned up
+			if tagIdx := slices.IndexFunc(networkInterface.TagSet, func(tag *ec2.Tag) bool {
+				return *tag.Key == config.NetworkInterfaceOwnerTagKey
+			}); tagIdx != -1 {
+				switch *networkInterface.TagSet[tagIdx].Value {
+				case config.NetworkInterfaceOwnerTagValue:
+					vpcrcLeakedENICleanupCnt.Inc()
```
haouc, line 143: I think in the other PR (#374) you propose to use resetting counter and only counting failed `delete` in leakedENICounter. Do you want to add the same strategy here, or want to wait for the other PR?

pkg/aws/ec2/api/eni_cleanup.go
sushrk: Makes sense, fixed this in latest commit

pkg/aws/ec2/api/eni_cleanup.go
sushrk: Added the new metrics in this PR too, thanks!

pkg/aws/ec2/api/eni_cleanup.go
```diff
@@ -140,12 +149,13 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
...
-					vpcrcLeakedENICleanupCnt.Inc()
+					vpcrcAvailableENICnt.Inc()
 				case config.NetworkInterfaceOwnerVPCCNITagValue:
-					vpcCniLeakedENICleanupCnt.Inc()
+					vpccniAvailableENICnt.Inc()
 				default:
-					// We will not hit this case as we only filter for above two tag values, adding it for any future use cases
-					e.Log.Info("found available ENI not created by VPC-CNI/VPC-RC")
+					// We should not hit this case as we only filter for relevant tag values, log error and continue if unexpected ENIs found
+					e.Log.Error(fmt.Errorf("found available ENI not created by VPC-CNI/VPC-RC"), "eni id", *networkInterface.NetworkInterfaceId)
```
haouc, line 157: nit: Can you remove the space? Cloudwatch filter is not friendly for spaces on key words.

pkg/aws/ec2/api/eni_cleanup.go
```diff
@@ -102,6 +108,9 @@ func (e *ENICleaner) Start(ctx context.Context) error {
 // interval between cycle 1 and 2 and hence can be safely deleted. And we can also conclude that Interface 1 was
 // created but not attached at the the time when 1st cycle ran and hence it should not be deleted.
 func (e *ENICleaner) cleanUpAvailableENIs() {
+	vpccniAvailableENICnt.Set(0)
```
haouc, line 111: I think for gauge type you don't need to reset the value. Every time when you report a value you can use set() instead of inc().

main.go
```diff
@@ -183,6 +185,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if vpcID == "" {
```
orsenthil, line 188: This is backwards incompatible change. How do we plan to introduce this in our deployment. What if we are not able to update VPC RC for some customers (we typically call broken clusters)?

pkg/aws/ec2/api/wrapper.go
```diff
@@ -639,6 +657,38 @@ func (e *ec2Wrapper) DescribeNetworkInterfaces(input *ec2.DescribeNetworkInterfa
...
+	var networkInterfaces []*ec2.NetworkInterface
+	input.MaxResults = aws.Int64(config.DescribeNetworkInterfacesMaxResults)
+
+	start := time.Now()
+	if err := e.userServiceClient.DescribeNetworkInterfacesPages(input, func(output *ec2.DescribeNetworkInterfacesOutput, _ bool) bool {
+		ec2APICallCnt.Inc()
+		ec2DescribeNetworkInterfacesPagesAPICallCnt.Inc()
+		//Currently only network interface ID and the tag set is require, only add required details to avoid consuming extra memory
```
orsenthil, line 671: What does `avoid consuming extra memory` mean?

pkg/aws/ec2/api/eni_cleanup.go
```diff
@@ -116,61 +126,59 @@ func (e *ENICleaner) cleanUpAvailableENIs() {
...
-					case config.NetworkInterfaceOwnerVPCCNITagValue:
-						vpcCniLeakedENICleanupCnt.Inc()
-					default:
-						// We will not hit this case as we only filter for above two tag values, adding it for any future use cases
-						e.Log.Info("found available ENI not created by VPC-CNI/VPC-RC")
-					}
-				}
+	networkInterfaces, err := e.EC2Wrapper.DescribeNetworkInterfacesPages(describeNetworkInterfaceIp)
+	if err != nil {
+		e.Log.Error(err, "failed to describe network interfaces, will retry")
```
orsenthil, line 140: When does the retry stop?

sushrk: The cleanup routine runs every 30 mins until controller shutdown, this means that deletion of leaked ENIs will be retried in the next run of the cleanup. 
The error log could be clearer here, just copied it over from before. I'll update this.

main.go
sushrk: The controller manifest will be updated to add the required parameter in the next release.

sushrk:
> **Some overall comments**
> 
> ```
>     Created 500+ ENIs in same VPC to confirm call volume has not increased drastically
> 
> ec2_describe_network_interfaces_pages_api_call_count 4
> ec2_describe_network_interfaces_pages_api_err_count 0
> ```
> 
>     1. Do we have _before_ and _after_ call count.
> 
>     2. We could measure is the pod create latency of the SGPP operation _before_ and _after_ the change in cluster with 1000 ENI and ensure that they there is no change.
> 
>     3. Does the paginated version return any empty pages at all, even with a filter? Can we assert against it?
> 
>     4. What are the drawbacks on including a required vpc-id as a parameter in the new version of VPC RC?
> 
>     5. So the trunk Init routine, `WaitForNetworkInterfaceStatusChange` will continue to use DescribeNetworkInterfaces (unpaginated), but only the ENI cleanup will use the paginated. Is that correct? Do we see any problems with trunk init using the unpaginated, with large number of ENI cases?

1. Yes, we do a benchmark test for every release, the first graph is from `v1.4.3` vs second graph from the test. This is running the integration test cases in a loop for 5 times. 
Screenshot 2024-02-28 at 1 01 12 PM

Screenshot 2024-02-28 at 12 19 09 PM

2. SGPP pod creation latency is comparable to previous releases when scaled up to 1000 pods, approx 4 mins.
3. The paginated version could return empty pages, this is not deterministic. But using deep filter seems to have alleviated the problem of increased call volume. Graph for soak test with 100 nodes and scaling pods using security pods from 0 to 1000 every 15 mins, running overnight
Screenshot 2024-02-28 at 10 48 54 AM

4. There should be no drawbacks, we will update the controller manifest accordingly with the release to populate the VPC-ID. This is required for the cleanup routine to pass the vpc-id in `DescribeNetworkInterfaces`
5. All other calls to `DescribeNetworkInterfaces` pass in the network interface ID, like in `WaitForNetworkInterfaceStatusChange`, and is not an issue with respect to throttling.

main.go
sushrk: Also, if we do rollback, the manifest will be rolled back, not just the controller image image. So. IMO this should still be safe.

haouc (approved):
lgtm

pkg/aws/ec2/api/wrapper.go
orsenthil: This can explained in the comment too. It will help us in future.

orsenthil:
> All other calls to DescribeNetworkInterfaces pass in the network interface ID, like in WaitForNetworkInterfaceStatusChange, and is not an issue with respect to throttling.

Got it. Thank you!.

orsenthil (approved):
LGTM

pkg/aws/ec2/api/wrapper.go
sushrk: Ack, I can add more info here in a follow-up PR.

PR #386, opened 2024-03-18
fix:update cluster tag name in CNINode

sushrk:
*Issue #, if available:*
NA
*Description of changes:*

Update the cluster tag key to be deterministic in the CNINode CRD for use in other components like VPC-CNI where cluster name might be unknown.


haouc (approved):
lgtm

PR #388, opened 2024-03-23
add node os label in CNINode and run finalizer routine for linux only

sushrk:
*Issue #, if available:*
N/A
*Description of changes:*
1. Adding node OS label to CNINodes to run the finalizer routine for Linux nodes only. 
2. Added retry to get CNINode during node reconcile: When CNINode is deleted, node update is triggered and since CNINode is not immediately found, the controller updates the node to be unmanaged and adds node to delete queue and removes the trunk, this is a regression bug. CNINode is shortly re-created by the CNINode controller, so added retry with backoff to GetCNINode which prevents node being unmanaged. 

Tests:
1. CNINode creation: Validated new CNINodes are created with required fields
3. Node deletion: Validated finalizer routine is only run for Linux nodes
4. Reconcile tests: Edit CNINode to remove required fields and validated it is updated correctly

### DIFF
--- a/controllers/crds/cninode_controller.go
+++ b/controllers/crds/cninode_controller.go
@@ -91,11 +91,24 @@ func (r *CNINodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	nodeFound := true
+	node := &v1.Node{}
+	if err := r.Client.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			nodeFound = false
+		} else {
+			r.Log.Error(err, "failed to get the node object in CNINode reconciliation, will retry")
+			// Requeue request so it can be retried
+			return ctrl.Result{}, err
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


controllers/crds/cninode_controller.go
```diff
@@ -91,33 +91,67 @@ func (r *CNINodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 			} else {
 				cniNodeCopy.Spec.Tags = map[string]string{
 					config.CNINodeClusterNameKey: r.ClusterName,
 				}
 			}
+			shouldPatch = true
+		}
+		// if node exists, get & add OS label if it does not exist on CNINode
+		if nodeFound {
+			nodeLabelOS := node.ObjectMeta.Labels[config.NodeLabelOS]
```
haouc, line 123: If we have access to node and node metadata has OS info, why can't we use them directly when checking condition for finalizer cleanup?

haouc: Synced with @sushrk offline, this is for deleted nodes in which case we don't have access the node object. This makes sense to me.

haouc (approved):
lgtm

haouc (approved):
lgtm

Issue #20, opened 2021-02-12
Ability to use Security Group Name in Security Group Policy


abhipth:
**Is your feature request related to a problem? Please describe.**
Security Groups are tied to VPC which means you cannot reuse Security Group Policy on clusters belonging to different VPC. 

**Describe the solution you'd like**
SGP should allow specifying security groups using security group name.

swade1987:
This will make reviewing PRs much easier.

jdnurmi:
Just to add to this, it would make using security groups defined in the EC2 ack usable - as it is, we can't reference a constructed security group automatically unless we can refer to it by name or pull it from a configMap - it'd be great if I could have a single helm chart define the security groups _and_ the SGP instead of having to define at least one of them out-of-band or by external tooling.

Aleksei-Poliakov:
@jdnurmi has a great point about ACK integration - something like this would be very useful not only for `SecurityGroupPolicy`, but also for other places that require a security group, for example `ENIConfig`

GnatorX:
I opened a PR for this but not sure who I should engage to see where we are on supporting this feature

jayanthvn:
@sjastis @hao for viz..

mattbrandman:
I was literally just looking for this feature. Makes it way easier to assign when you generate the SGs automatically in terraform. Having to pick out the ID by hand is a pain and not stable if it gets destroyed

GnatorX:
Hey wanted to check in since I haven't seen any update in the PR or the issue. Where are we in supporting this with the PR?

taer:
I wrote a small mutation webhook controller that I have 90% done. It was pretty straight forward. It had 2 features
* annotation with a list of sec-group names to apply
* annotation with a "also add the node security group" which probably needs to be default.

I'll see if company has OSS or not

PR #389, opened 2024-03-27
Support security group names in VPC Resource controller

GnatorX:
*Issue #, if available:* https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/20

*Description of changes:*
Add support for Security group names

Credit: Most of the code was implemented by @abeer-stripe


GnatorX:
I have tested this, with some augmentation, on our cluster and it seems to work. We would need EKS to test this because we had to remove CreateNetworkInterfacePermission from the code (not something we can / need to do).

mattbrandman:
This would also be super useful for tying together security group policies generated by TF with a naming schema. Right now we use a cli to generate the sg pull the id out and then set it as a tf var for import and a variable for argocd. This would be a huge improvement in that workflow

sushrk:
Thanks for the PR, I'm going to take a look this week.

pkg/aws/ec2/api/helper.go
```diff
@@ -594,3 +619,81 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
...
 }
+
+func (h *ec2APIHelper) GetSecurityGroupIdsForSecurityGroupNames(securityGroupNames []string) ([]string, error) {
+	sgIds := make([]string, 0)
+	if len(securityGroupNames) < 1 {
+		return sgIds, nil
+	}
+	sgNamesNotInCache := make([]string, 0)
+	for _, sgName := range securityGroupNames {
+		if sgId, ok := h.securityGroupNameToIdMap[sgName]; !ok {
```
sushrk, line 630: I'm not sure caching is a good idea here. 

Consider the scenario where a security group policy includes `sg-example-name` corresponding to `sg-1111`. This is added to `securityGroupNameToIdMap`, and some pod A is launched with `sg-1111` attached to the branchENI. Now, customer deleted `sg-1111` and created `sg-2222` with the same name `sg-example-name`. But the map still points to old `sg-1111` which no longer exists. So any pods launched after would be stuck in creation process as branchENI cannot be created(due to missing SG). 

Need to evaluate if we always have to do an API call during pod creation or we can consider periodically refreshing the cache (TTL caching).

pkg/aws/ec2/api/helper.go
```diff
@@ -594,3 +619,81 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
...
+		},
+		{
+			Name:   aws.String("vpc-id"),
+			Values: aws.StringSlice([]string{h.workerNodeVpcId}),
+		},
+	}
+	input := &ec2.DescribeSecurityGroupsInput{Filters: filters}
+	foundSgNames := map[string]bool{}
+	for {
+		output, err := h.ec2Wrapper.DescribeSecurityGroups(input)
```
sushrk, line 653: Can we consider using the `GetSecurityGroupsForVpc` API call instead of `DescribeSecurityGroups` here?

pkg/aws/ec2/api/helper.go
```diff
@@ -594,3 +619,81 @@ func (h *ec2APIHelper) DetachAndDeleteNetworkInterface(attachmentID *string, nwI
...
+	}
+	if len(missingSgNames) > 0 {
+		sort.Strings(missingSgNames)
+		return nil, fmt.Errorf("security groups %s not found", missingSgNames)
+	}
+	return sgIds, nil
+}
+
+// GetSourceAcctAndArn constructs source acct and arn and return them for use
+func GetSourceAcctAndArn(roleARN, region, clusterName string) (string, string, error) {
```
sushrk, line 681: Any reason why this is moved from `pkg/utils/helper.go`?

pkg/utils/helper.go
```diff
@@ -155,11 +162,18 @@ func (s *SecurityGroupForPods) filterPodSecurityGroups(
 			continue
 		}
 
+		sgNameList = append(sgNameList, sgp.Spec.SecurityGroupNames.GroupNames...)
 		sgList = append(sgList, sgp.Spec.SecurityGroups.Groups...)
 	}
 
+	sgNameList = RemoveDuplicatedSg(sgNameList)
```
sushrk, line 169: It would be preferable to enforce this via CRD validation in `apis/vpcresources/v1beta1/securitygrouppolicy_types.go` for both SG IDs and names, like `+kubebuilder:validation:UniqueItems=true`.

sushrk (changes requested):

apis/vpcresources/v1beta1/securitygrouppolicy_types.go
```diff
@@ -24,6 +24,15 @@ type SecurityGroupPolicySpec struct {
 	PodSelector            *metav1.LabelSelector `json:"podSelector,omitempty"`
 	ServiceAccountSelector *metav1.LabelSelector `json:"serviceAccountSelector,omitempty"`
 	SecurityGroups         GroupIds              `json:"securityGroups,omitempty"`
+	SecurityGroupNames     GroupNames            `json:"securityGroupNames,omitempty"`
```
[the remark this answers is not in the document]
GnatorX: Totally

apis/vpcresources/v1beta1/securitygrouppolicy_types.go
```diff
@@ -24,6 +24,15 @@ type SecurityGroupPolicySpec struct {
...
+}
+
+// GroupNames contains the list of security group names that will be applied to the network interface of the pod matching the criteria.
+type GroupNames struct {
+	// Groups is the list of EC2 Security Group Names that need to be applied to the ENI of a Pod.
+	// +kubebuilder:validation:MinItems=1
```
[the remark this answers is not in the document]
GnatorX: ya thats a good call out, let me see what we can do here

pkg/aws/ec2/api/helper.go
GnatorX: Yup let me look into this

pkg/aws/ec2/api/helper.go
GnatorX: There is a circular dependency which required moving a few things out of utils such as this and the httpclient code

pkg/utils/helper.go
GnatorX: make sense let me do that. It does make it slightly harder to validate just at the CRD if we are mixing SG id and SG names though

pkg/aws/ec2/api/helper.go
GnatorX: I think a TTL caching make sense given (at least from our side) how rare we change delete security group and recreate with the same name. The cache was added to reduce the number of calls we hit EC2 API with.

pkg/aws/ec2/api/helper.go
```diff
@@ -60,17 +62,39 @@ var (
...
-	return &ec2APIHelper{ec2Wrapper: ec2Wrapper}
+	return &ec2APIHelper{
+		ec2Wrapper:               ec2Wrapper,
+		workerNodeVpcId:          workerNodeVpcId,
+		securityGroupNameToIdMap: make(map[string]string),
+	}
+}
+
+// For unit testing
+func newEC2APIHelper(ec2Wrapper EC2Wrapper, clusterName, workerNodeVpcId string,
```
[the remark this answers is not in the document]
GnatorX: We added this here because need to to expose the internal cache for testing.

GnatorX:
Updated the PR. Let me know what you think about TTL cache. If you want to move in that direction. Do you have a recommendation of TTL cache, if not I will just pick up https://github.com/kubernetes/client-go/blob/v0.29.3/tools/cache/expiration_cache.go#L38

apis/vpcresources/v1beta1/securitygrouppolicy_types.go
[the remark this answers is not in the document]
GnatorX: The non-empty I believe is checked on CreateNetworkInterface. Do you have different place in mind?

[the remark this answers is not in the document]
sushrk: Yeah, it would be good to include in CRD validation, let me see how to do this with `kubebuilder`.

sushrk:
> Updated the PR. Let me know what you think about TTL cache. If you want to move in that direction. Do you have a recommendation of TTL cache, if not I will just pick up https://github.com/kubernetes/client-go/blob/v0.29.3/tools/cache/expiration_cache.go#L38

Thanks, will go through this. 
Can you also rebase and squash the commits? Would make it easier to review and merge the PR.

GnatorX:
I am going to close this against https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/408 because it was simpler to rebase + squash

PR #390, opened 2024-03-27
pin envtest version due to an upstream bug

haouc:
*Issue #, if available:*

*Description of changes:*
We have to pin the envtest version before [the upstream issue](https://github.com/kubernetes-sigs/controller-runtime/issues/2720#issuecomment-2016461180) is fixed.


sushrk (approved):
lgtm, thanks!

PR #398, opened 2024-04-02
add github action to run gosec static analysis

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Add `gosec` static analysis check in github actions. 
By default, this scans for high & medium issues. Excluding low issues as they are not very useful to fix(only low issues reported from local scan were many unhandled errors), also excludes test and generated files. 

Note on G108,G114: The go profiler is not enabled listening on a port be default, it is wrapped by an optional flag `enable-profiling`. This complies with a security best practice:
* only expose to localhost at a port which is not used for http server
* the controller doesn't open any endpoint to internet as a server
* the controller runs in isolated EKS control plane instance and doesn't server any traffic to/from outside world

### DIFF
--- a/pkg/k8s/pod/converter.go
+++ b/pkg/k8s/pod/converter.go
@@ -56,6 +56,7 @@ func (c *PodConverter) ConvertList(originalList interface{}) (convertedList inte
 		},
 	}
 	for _, pod := range podList.Items {
+		pod := pod // Fix gosec G601, so we can use &node
 		strippedPod := c.StripDownPod(&pod)
 		strippedPodList.Items = append(strippedPodList.Items, *strippedPod)
 	}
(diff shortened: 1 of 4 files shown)


pkg/k8s/pod/converter.go
```diff
@@ -56,6 +56,7 @@ func (c *PodConverter) ConvertList(originalList interface{}) (convertedList inte
 		},
 	}
 	for _, pod := range podList.Items {
+		pod := pod // Fix gosec G601, so we can use &node
```
haouc, line 59: I am not sure if we want to fix this for now since golang 1.22 changes the variable in for loop. https://tip.golang.org/doc/go1.22

Or can you create a backlog to revisit this when we upgrade to v1.22? I think we will upgrade soon.

haouc: Offline syncing with @sushrk. We can have a backlog to revisit this change after upgrading to go 1.22.

haouc (approved):
lgtm

PR #400, opened 2024-04-04
remove DisassociateAllBranchENIs as it is not useful

sushrk:
*Issue #, if available:*
N/A

*Description of changes:*
Removing `DisassociateAllBranchENIs` function as it is not useful. 
There is a delay in cleaning up resources after node deletion, see [`NodeDeleteRequeueRequestDelay`](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/provider/branch/provider.go#L79). So [`DeleteNode`](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/provider/branch/provider.go#L203) is called 1 min after node object is deleted. 

After adding the node termination finalizer, the branch ENIs will be cleaned up right after the node object(and CNINode) is deleted. So by the time `DeleteNode` is called after a minute, we do not expect to have branchENIs associated with a trunk ENI to call `DisassociateNetworkInterface`. Also, once the trunk ENI is deleted by node deletion, the association between branch and trunk ENI is automatically deleted. 

If any ENI is leaked, the periodic cleanup routine will clean it up in the next cycle as long as the cluster is active. So far in tests, we have not observed branchENIs being leaked after the node termination finalizer is run. 


pkg/aws/ec2/api/cleanup/eni_cleanup.go
```diff
@@ -149,7 +151,7 @@ func (e *ENICleaner) DeleteLeakedResources() error {
 				_, err := e.EC2Wrapper.DeleteNetworkInterface(&ec2.DeleteNetworkInterfaceInput{
 					NetworkInterfaceId: nwInterface.NetworkInterfaceId,
 				})
-				if err != nil {
+				if err != nil && !strings.Contains(err.Error(), ec2Errors.NotFoundInterfaceID) { // ignore InvalidNetworkInterfaceID.NotFound error
```
haouc, line 154: nit: if we not skip `NotFound`, the following log will be kind of misleading in case the ENI is not found.

haouc: Actually I think we should skip `NotFound` instead of continue to log them as success. `NotFound` could be caused by various reasons. I don't think we should log `NotFound` as success here.

sushrk: Makes sense, let me update the log.

haouc (approved):
lgtm

PR #405, opened 2024-04-11
Update controller_auth_proxy_patch.yaml

orsenthil:
Update the reference from gcr.io to registry.k8s.io

>  kube-rbac-proxy is moving to registry.k8s.io/kubebuilder/kube-rbac-proxy (from gcr.io/kubebuilder/kube-rbac-proxy) because GCR is being sunset. We need to update these references.

### DIFF
--- a/config/default/controller_auth_proxy_patch.yaml
+++ b/config/default/controller_auth_proxy_patch.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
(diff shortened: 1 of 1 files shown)


jaydeokar (approved):
Same.. Verify if images are already present in upstream

orsenthil:
Yes, the image is present https://explore.ggcr.dev/?repo=registry.k8s.io%2Fkubebuilder%2Fkube-rbac-proxy

PR #413, opened 2024-04-22
Modify SecurityGroups Maxitems

changhyuni:
*Issue #, if available:*
Modify the maxitems in SecurityGroups.
According to "Amazon VPC Quotas," it should be able to use up to 16.

*Description of changes:*
We need to use more than 10 security groups.
Increase the maxitems to the maximum number of allowable (16), and modify the descritions.
https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html#vpc-limits-security-groups

### DIFF
--- a/apis/vpcresources/v1beta1/securitygrouppolicy_types.go
+++ b/apis/vpcresources/v1beta1/securitygrouppolicy_types.go
@@ -30,7 +30,7 @@ type SecurityGroupPolicySpec struct {
 type GroupIds struct {
 	// Groups is the list of EC2 Security Groups Ids that need to be applied to the ENI of a Pod.
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:MaxItems=5
+	// +kubebuilder:validation:MaxItems=16
 	Groups []string `json:"groupIds,omitempty"`
 }
 
--- a/config/crd/bases/vpcresources.k8s.aws_securitygrouppolicies.yaml
+++ b/config/crd/bases/vpcresources.k8s.aws_securitygrouppolicies.yaml
@@ -99,7 +99,7 @@ spec:
                       need to be applied to the ENI of a Pod.
                     items:
                       type: string
-                    maxItems: 5
+                    maxItems: 16
                     minItems: 1
                     type: array
                 type: object


sushrk:
The default number of security groups per Network Interface is 5. Updating the `MaxItems=16` could lead to customers erroneously adding more than 5 SGs in the security group policy prior to increasing the service quota. 

The recommended way is to define more than 1 security group if you need more than 5 SGs to be specified, as documented in the [README](https://github.com/aws/amazon-vpc-resource-controller-k8s?tab=readme-ov-file#security-group-for-pods). 

Can you help me understand if any issues with multiple CRDs? We understand this is harder to manage than 1 CRD, but hopefully it is a one-time setup effort for SGP CRDs.

sushrk:
>If I apply it without increasing the qutos, won't it be rejected by the vpc qutos anyway?

I need to check on this. 

>Is there any advantage to enforcing this through maxitems?

Yeah, as mentioned before this avoids customers from specifying more than the default number(5) of security groups in the CRD.

>how about making it possible to dynamically modify maxitems?

Hm, this might be possible if we set maxItems in CRD based on the service quota value. Let me see if it is feasible.

changhyuni:
@sushrk 
How about using servicequotas to look up the vpc service qutos(securitygroup) value to set maxitems when vpc cni is deployed?
Let me see if it is feasible too.

GnatorX:
We would love this feature also. We have increased service quotas and currently rely on a fork of the VPC RC to get around this

changhyuni:
@sushrk 
Hi!
How's it going? I'm curious because I haven't seen any updates.

sushrk:
Sorry for the late reply. Thinking about this, it makes sense to me to set the `maxItems` to absolute maximum of 16 in CRD definition and internally check that the number of SGs specified is lower than the service quota value at CRD creation. 

We'd need to document this more clearly though, especially when customers want to lower their quota so it does not break existing workloads. 

I'll need to confirm the customer experience with PM, will post an update tomorrow. 

@changhyuni are you planning to add the support for this?

changhyuni:
@sushrk 
Hi. Thank you for your comment.
I think I can create a webhook to check the number of security groups in the CRD. 
What do you think? I'v tried to make webhook and will add test code soon. 
Can I use the aws-go-sdk-v2? I looked at your other code and it doesn't seem to use it.

sushrk:
>I think I can create a webhook to check the number of security groups in the CRD.

Can you explain why we are adding a webhook to check the number of SGs? Does this validate the CRD at creation/update/both?
Also, we need to handle the case where more than one SGP CRD can be applied to a pod, probably will not be sufficient to validate individual CRD objects. We should add this validation at pod creation. 

>Can I use the aws-go-sdk-v2? I looked at your other code and it doesn't seem to use it, it uses an api package. 

We use aws-go-sdk for EC2 operations, but through a wrapper and helper functions.

changhyuni:
@sushrk 
Hi I had some time and made a few additions.
Created an API via ec2 wrapper and added logic to validate on pod webhooks

changhyuni:
@sushrk 
Could you check this out when you have a time?

haouc:
We need avoid adding upstream service support check in pod mutating webhook. The webhook need to be light and respond quickly. Always calling upstream API will be more and more expensive for most of users who don't have the need to support more than 5 security groups. Since users should be aware that their service quotas before applying the setting to the CRD, I think fail a pod creation and send a pod event as a message should be good enough.

changhyuni:
@haouc 
Hi. Thanks you for comment I agree that API calls increase 
I'll remove the pod webhook, so could you merge the crd settings change for the 16 maxitems?

changhyuni:
@haouc @sushrk 
Hi 
If you have time could you check this?

changhyuni:
@haouc 
Hello, How is the requirement coming along?
I'd like to assist if possible. How can I make any necessary changes?

GnatorX:
I have created https://github.com/aws/containers-roadmap/issues/2608 to try to drive traction

PR #415, opened 2024-04-24
Add new target for building docker images with no tests

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This creates a new target to build the image for vpc resource controller.
- We pull the latest image for golang using the minor version instead of pinning the exact patch version
- Update the golang image to "public.ecr.aws/docker/library/golang"


Makefile
```diff
@@ -13,7 +13,8 @@ VERSION ?= $(GIT_VERSION)
 IMAGE ?= $(REPO):$(VERSION)
 BASE_IMAGE ?= public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest.2
 GOLANG_VERSION ?= $(shell cat .go-version)
-BUILD_IMAGE ?= public.ecr.aws/bitnami/golang:$(GOLANG_VERSION)
+GO_IMAGE_TAG=$(shell cat .go-version | cut -f 1-2 -d".")
```
sushrk, line 16: Should we update [.go-version](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/.go-version) file to include the minor version?

Makefile
```diff
@@ -13,13 +13,10 @@ VERSION ?= $(GIT_VERSION)
 IMAGE ?= $(REPO):$(VERSION)
 BASE_IMAGE ?= public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest.2
 GOLANG_VERSION ?= $(shell cat .go-version)
-BUILD_IMAGE ?= public.ecr.aws/bitnami/golang:$(GOLANG_VERSION)
+BUILD_IMAGE ?= public.ecr.aws/docker/library/golang:$(GOLANG_VERSION)
 GOARCH ?= amd64
 PLATFORM ?= linux/amd64
 
-export GOSUMDB = sum.golang.org
```
jaydeokar, line 20: We don't need to set these variables as we are building the binaries using the latest golang patch version for the family..  These variables are already set in the go env variables inside the container

```
GOSUMDB='sum.golang.org'
GOTOOLCHAIN='local'
GOVERSION='go1.22.2'
``` 
More context here - https://go.dev/doc/toolchain
```
The go command selects the Go toolchain to use based on the GOTOOLCHAIN setting. To find the GOTOOLCHAIN setting, the go command uses the standard rules for any Go environment setting:

If GOTOOLCHAIN is set to a non-empty value in the process environment (as queried by [os.Getenv](https://go.dev/pkg/os/#Getenv)), the go command uses that value.

Otherwise, if GOTOOLCHAIN is set in the user’s environment default file (managed with [go env -w and go env -u](https://go.dev/cmd/go/#hdr-Print_Go_environment_information)), the go command uses that value.

Otherwise, if GOTOOLCHAIN is set in the bundled Go toolchain’s environment default file ($GOROOT/go.env), the go command uses that value.
```

Go toolchain local will fall back to using what's packed in the local go environment.. i.e. the patch version of the image that was pulled

Makefile
jaydeokar: Updated to Minor version

Makefile
```diff
@@ -79,6 +76,11 @@ docker-buildx: check-env test
 docker-build: check-env test
 	docker build --build-arg BASE_IMAGE=$(BASE_IMAGE) --build-arg ARCH=$(GOARCH) --build-arg BUILD_IMAGE=$(BUILD_IMAGE) . -t ${IMAGE}
 
+
+# Build the docker image with buildx and no tests
+docker-buildx-no-test:
+	docker buildx build --platform=$(PLATFORM) -t $(IMAGE)_$(GOARCH) --build-arg BASE_IMAGE=$(BASE_IMAGE) --build-arg BUILD_IMAGE=$(BUILD_IMAGE) --build-arg $(GOARCH) --load .
```
orsenthil, line 82: How is no test ensured in this?

orsenthil: Oh, you mean, removing the dependency on the target test , to save time?

jaydeokar: Yes, correct

sushrk (approved):
Lgtm, thanks.

PR #417, opened 2024-04-30
Remove deprecated flag from test script

jaydeokar:
*Issue #, if available:*
NA

*Description of changes:*
Removes the deprecated flag `--no-cli-pager` as per awscli v2


orsenthil (approved):

orsenthil:
Discussed offline. We don't need this change, the flag is present in awscli v2.

PR #427, opened 2024-05-29
Fix CNINodes integration test for internal testbed 

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This ensures there is always a node running in the cluster while running the cni nodes integration suite

image

### DIFF
--- a/test/integration/cninode/cninode_test.go
+++ b/test/integration/cninode/cninode_test.go
@@ -44,6 +44,11 @@ var _ = Describe("[CANARY]CNINode test", func() {
 			Expect(err).ToNot(HaveOccurred())
 			oldMinSize = *asg[0].MinSize
 			oldMaxSize = *asg[0].MaxSize
+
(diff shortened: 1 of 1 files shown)


test/integration/cninode/cninode_test.go
```diff
@@ -44,6 +44,11 @@ var _ = Describe("[CANARY]CNINode test", func() {
 			Expect(err).ToNot(HaveOccurred())
 			oldMinSize = *asg[0].MinSize
 			oldMaxSize = *asg[0].MaxSize
+
+			// Keep atleast one instance running in the cluster when restoring the values in AfterEach
+			if oldMinSize == 0 {
+				oldMinSize += 1
```
sushrk, line 50: Can we just set `oldMinSize=1` in this case?

jaydeokar: Okay. Updated

sushrk (approved):
Thanks for fixing it, lgtm.

PR #436, opened 2024-06-19
QPS and busrt adjustment

haouc:
*Issue #, if available:*

*Description of changes:*
- The burst is incorrectly set and qps is capped to smaller number. 
- We should use the same rate limiting for related API calls.
- Update the kubernetes client qps and burst to be same as controller-runtime default values.
- Double the pod reconciler workers.


jaydeokar (approved):
lgtm

Issue #428, opened 2024-05-30
Add Windows configuration options to control IP address allocation in Secondary IP Mode


tzifudzi:
### Background
- The current default behavior of the VPC Resource Controller in Amazon EKS for Windows nodes using secondary IP mode is to allocate a fixed number of warm secondary IP addresses on each Windows node. However, in some scenarios and use cases, this approach while beneficial to enable faster pod startup times, can lead to IP address wastage, especially in subnets with limited IPv4 address space. 

### What would you like to be enhanced
Add the following configuration options to the VPC Resource Controller to provide more flexibility in managing IP address allocation when using the Secondary IP Mode for Windows nodes:

1. `minimum-ip-target`: Specify the minimum number of minimum number of IP addresses to be available at any time
2. `warm-ip-target`: Specify the target number of unused IP addresses allocated in excess of current need to enable faster pod startup times

### Why is the change needed and what use case will it solve
- EKS users who need to carefully manage their IP address utilization may find this default behavior problematic, as it can result in unused IP addresses on their nodes and reduced IP address availability. To address this issue, adding configuration options to control the IP address allocation process would provide more flexibility and allow users to optimize their IP address usage and availability based on their specific requirements and tradeoffs for warm IP availability.

### Task list
- [ ] Create PR with initial changes to support the new configuration parameters
- [ ] Respond to PR review feedback and merge PR
- [ ] Add documentation about new configuration parameters
- [ ] Verify changes are effected and generally available across all regions :tada:

tzifudzi:
I will be working on this issue. @sushrk/@haouc Please assign this to me.

PR #443, opened 2024-07-17
Add Windows secondary IP mode configurable options for managing IP address allocation

tzifudzi:
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

## Issue 
Resolves https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/428

### Description of changes
This change introduces changes to provide configurable options for managing IP address allocation on Windows nodes when using secondary IP mode. The VPC Resource Controller will now allow configuring of the `windows-minimum-ip-target` and `windows-warm-ip-target` parameters in the `amazon-vpc-cni` ConfigMap. These parameters will control the minimum number of IP addresses to be allocated on each Windows node, and the number of 'warm' IP addresses to be pre-allocated respectively. The VPC RC will always ensure that based on the configurations, there will always be enough warm IPs to satisfy the configurations. 

Previously, the VPC Resource Controller utilized a fixed and hardcoded number of 3 warm secondary IP addresses on each Windows node. This approach, while beneficial for enabling faster pod startup times, wasn't ideal for the following example scenarios. 

1. _When there is a need to configure fewer warm IPs to improve IP address utilization in the subnet._ The fixed warm IP target could lead to inefficient IP address utilization, particularly in subnets with limited IPv4 address space.
2. _When there is a need to configure a higher target for warm IPs to ensure faster pod startup times._ The fixed warm IP target may not provide enough pre-allocated IP addresses to meet the requirements for faster pod launches in a case where a warmpool target greater than 3 is desired.
3. _When there is a need to configure a minimum IP target after the node is set up._ The lack of a configurable minimum IP target could result in insufficient IP addresses being allocated to the node, potentially impacting pod startup times in cases where high availability is desired.

### Scope
- This change only affects Windows nodes when `enable-windows-ipam` is set to `true`
- This change only affects Windows modes when running in the Secondary IP mode, in contrast to running in prefix delegation mode

### Notes for reviewers
- A warmpool target of 1 is enforced even if a zero value is set because the current code design workflow relies on the warmpool always being populated with at least 1 warm IP. In future, a follow up change can be applied to make it both minimum IP and warm IP targets configurable to zero such that zero IPs are kept warm for use cases where no warm IPs are desired.

Todo 
- [x] Integration tests will be added later to further validate the proposed changes. To start with unit tests have been added and manual e2e tests have been performed. After reaching consensus on the core design and implementation of the initial changes, I can follow up with integration tests either in this PR or a different one.
- [x] Documentation will be updated to reflect the new configuration options and their intended usage once the core design and default behavior have been finalized and agreed upon. This will ensure that the documentation accurately represents the final implementation.

### Integration Test Output
Output as of Tue Jul 23 11:52AM PDT
```
[tzifudzi:~/GolandProjects/amazon-vpc-resource-controller-k8s/test/integration/windows] feature/secondary-ip-configs(+11/-15,1)+* 3h0m40s 130 ± ginkgo --skip=Stress -v --timeout 180m -- --cluster-kubeconfig='/Users/tzifudzi/.kube/config' --cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id=$VPC_ID --ginkgo.skip="stress"
Running Suite: Windows Integration Test Suite - /Users/tzifudzi/GolandProjects/amazon-vpc-resource-controller-k8s/test/integration/windows
...
Ran 34 of 40 Specs in 6833.774 seconds
SUCCESS! -- 34 Passed | 0 Failed | 0 Pending | 6 Skipped
PASS
...
```

Integration test runs
- July 22 [integration-test-output.txt](https://github.com/user-attachments/files/16353208/integration-test-output.txt)
- Aug 29 [output.txt](https://github.com/user-attachments/files/16808386/output.txt)

### DIFF
--- a/controllers/core/configmap_controller.go
+++ b/controllers/core/configmap_controller.go
@@ -46,8 +46,8 @@ type ConfigMapReconciler struct {
 	Condition                         condition.Conditions
 	curWinIPAMEnabledCond             bool
 	curWinPrefixDelegationEnabledCond bool
-	curWinPDWarmIPTarget              int
-	curWinPDMinIPTarget               int
+	curWinWarmIPTarget                int
+	curWinMinIPTarget                 int
 	curWinPDWarmPrefixTarget          int
 	Context                           context.Context
 }
@@ -116,21 +116,33 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		isPrefixFlagUpdated = true
 	}
 
-	// Check if configurations for Windows prefix delegation have changed
-	var isPDConfigUpdated bool
-	warmIPTarget, minIPTarget, warmPrefixTarget := config.ParseWinPDTargets(r.Log, configmap)
-	if r.curWinPDWarmIPTarget != warmIPTarget || r.curWinPDMinIPTarget != minIPTarget || r.curWinPDWarmPrefixTarget != warmPrefixTarget {
-		r.curWinPDWarmIPTarget = warmIPTarget
-		r.curWinPDMinIPTarget = minIPTarget
-		r.curWinPDWarmPrefixTarget = warmPrefixTarget
-		logger.Info("updated PD configs from configmap", config.WarmIPTarget, r.curWinPDWarmIPTarget,
-			config.MinimumIPTarget, r.curWinPDMinIPTarget, config.WarmPrefixTarget, r.curWinPDWarmPrefixTarget)
+	// Check if Windows IP target configurations in ConfigMap have changed
+	var isWinIPConfigsUpdated bool
 
-		isPDConfigUpdated = true
+	warmIPTarget, minIPTarget, warmPrefixTarget, isPDEnabled := config.ParseWinIPTargetConfigs(r.Log, configmap)
+	var winMinIPTargetUpdated = r.curWinMinIPTarget != minIPTarget
+	var winWarmIPTargetUpdated = r.curWinWarmIPTarget != warmIPTarget
+	var winPDWarmPrefixTargetUpdated = r.curWinPDWarmPrefixTarget != warmPrefixTarget
+	if winWarmIPTargetUpdated || winMinIPTargetUpdated {
+		r.curWinWarmIPTarget = warmIPTarget
+		r.curWinMinIPTarget = minIPTarget
+		isWinIPConfigsUpdated = true
+	}
+	if isPDEnabled && winPDWarmPrefixTargetUpdated {
+		r.curWinPDWarmPrefixTarget = warmPrefixTarget
+		isWinIPConfigsUpdated = true
+	}
+	if isWinIPConfigsUpdated {
+		logger.Info(
+			"Detected update in Windows IP configuration parameter values in ConfigMap",
+			config.WinWarmIPTarget, r.curWinWarmIPTarget,
+			config.WinMinimumIPTarget, r.curWinMinIPTarget,
+			config.WinWarmPrefixTarget, r.curWinPDWarmPrefixTarget,
+			config.EnableWindowsPrefixDelegationKey, isPDEnabled,
+		)
 	}
 
-	// Flag is updated, update all nodes
-	if isIPAMFlagUpdated || isPrefixFlagUpdated || isPDConfigUpdated {
+	if isIPAMFlagUpdated || isPrefixFlagUpdated || isWinIPConfigsUpdated {
 		err := UpdateNodesOnConfigMapChanges(r.K8sAPI, r.NodeManager)
 		if err != nil {
 			// Error in updating nodes
--- a/controllers/core/configmap_controller_test.go
+++ b/controllers/core/configmap_controller_test.go
@@ -16,14 +16,9 @@ package controllers
 import (
 	"context"
 	"errors"
+	"strconv"
 	"testing"
 
-	mock_condition "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
-	mock_k8s "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
-	mock_node "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
-	mock_manager "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
-	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
-	cooldown "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider/branch/cooldown"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
@@ -35,18 +30,36 @@ import (
 	fakeClient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	mock_condition "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/condition"
+	mock_k8s "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/k8s"
+	mock_node "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node"
+	mock_manager "github.com/aws/amazon-vpc-resource-controller-k8s/mocks/amazon-vcp-resource-controller-k8s/pkg/node/manager"
+	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/config"
+	cooldown "github.com/aws/amazon-vpc-resource-controller-k8s/pkg/provider/branch/cooldown"
 )
 
 var (
 	mockConfigMap = &corev1.ConfigMap{
 		TypeMeta:   metav1.TypeMeta{},
 		ObjectMeta: metav1.ObjectMeta{Name: config.VpcCniConfigMapName, Namespace: config.KubeSystemNamespace},
-		Data:       map[string]string{config.EnableWindowsIPAMKey: "true", config.EnableWindowsPrefixDelegationKey: "true"},
+		Data: map[string]string{
+			config.EnableWindowsIPAMKey:             "true",
+			config.EnableWindowsPrefixDelegationKey: "false",
+			config.WinMinimumIPTarget:               strconv.Itoa(config.IPv4DefaultWinMinIPTarget),
+			config.WinWarmIPTarget:                  strconv.Itoa(config.IPv4DefaultWinWarmIPTarget),
+		},
 	}
 	mockConfigMapPD = &corev1.ConfigMap{
 		TypeMeta:   metav1.TypeMeta{},
 		ObjectMeta: metav1.ObjectMeta{Name: config.VpcCniConfigMapName, Namespace: config.KubeSystemNamespace},
-		Data:       map[string]string{config.EnableWindowsIPAMKey: "false", config.EnableWindowsPrefixDelegationKey: "true"},
+		Data: map[string]string{
+			config.EnableWindowsIPAMKey:             "true",
+			config.EnableWindowsPrefixDelegationKey: "true",
+			config.WinMinimumIPTarget:               strconv.Itoa(config.IPv4PDDefaultMinIPTargetSize),
+			config.WinWarmIPTarget:                  strconv.Itoa(config.IPv4PDDefaultWarmIPTargetSize),
+			config.WinWarmPrefixTarget:              strconv.Itoa(config.IPv4PDDefaultWarmPrefixTargetSize),
+		},
 	}
 	mockConfigMapReq = reconcile.Request{
 		NamespacedName: types.NamespacedName{
@@ -89,11 +102,13 @@ func NewConfigMapMock(ctrl *gomock.Controller, mockObjects ...client.Object) Con
 	return ConfigMapMock{
 		MockNodeManager: mockNodeManager,
 		ConfigMapReconciler: &ConfigMapReconciler{
-			Client:      client,
-			Log:         zap.New(),
-			NodeManager: mockNodeManager,
-			K8sAPI:      mockK8sWrapper,
-			Condition:   mockCondition,
+			Client:             client,
+			Log:                zap.New(),
+			NodeManager:        mockNodeManager,
+			K8sAPI:             mockK8sWrapper,
+			Condition:          mockCondition,
+			curWinMinIPTarget:  config.IPv4DefaultWinMinIPTarget,
+			curWinWarmIPTarget: config.IPv4DefaultWinWarmIPTarget,
 		},
 		MockNode:                   mockNode,
 		MockK8sAPI:                 mockK8sWrapper,
@@ -103,11 +118,30 @@ func NewConfigMapMock(ctrl *gomock.Controller, mockObjects ...client.Object) Con
 	}
 }
 
-func Test_Reconcile_ConfigMap_Updated(t *testing.T) {
+func Test_Reconcile_ConfigMap_Updated_Secondary_IP(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	mock := NewConfigMapMock(ctrl, mockConfigMap)
+	mock.MockCondition.EXPECT().IsWindowsIPAMEnabled().Return(true)
+	mock.MockCondition.EXPECT().IsWindowsPrefixDelegationEnabled().Return(false)
+	mock.MockK8sAPI.EXPECT().ListNodes().Return(nodeList, nil)
+	mock.MockNodeManager.EXPECT().GetNode(mockNodeName).Return(mock.MockNode, true)
+	mock.MockNodeManager.EXPECT().UpdateNode(mockNodeName).Return(nil)
+
+	mock.MockK8sAPI.EXPECT().GetConfigMap(config.VpcCniConfigMapName, config.KubeSystemNamespace).Return(createCoolDownMockCM("30"), nil).AnyTimes()
+
+	cooldown.InitCoolDownPeriod(mock.MockK8sAPI, zap.New(zap.UseDevMode(true)).WithName("cooldown"))
+	res, err := mock.ConfigMapReconciler.Reconcile(context.TODO(), mockConfigMapReq)
+	assert.NoError(t, err)
+	assert.Equal(t, res, reconcile.Result{})
+}
+
+func Test_Reconcile_ConfigMap_Updated_PD(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mock := NewConfigMapMock(ctrl, mockConfigMapPD)
 	mock.MockCondition.EXPECT().IsWindowsIPAMEnabled().Return(true)
 	mock.MockCondition.EXPECT().IsWindowsPrefixDelegationEnabled().Return(true)
 	mock.MockK8sAPI.EXPECT().ListNodes().Return(nodeList, nil)
@@ -120,14 +154,13 @@ func Test_Reconcile_ConfigMap_Updated(t *testing.T) {
 	res, err := mock.ConfigMapReconciler.Reconcile(context.TODO(), mockConfigMapReq)
 	assert.NoError(t, err)
 	assert.Equal(t, res, reconcile.Result{})
-
 }
 
 func Test_Reconcile_ConfigMap_PD_Disabled_If_IPAM_Disabled(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	mock := NewConfigMapMock(ctrl, mockConfigMapPD)
+	mock := NewConfigMapMock(ctrl, mockConfigMap)
 	mock.MockCondition.EXPECT().IsWindowsIPAMEnabled().Return(false)
 	mock.MockCondition.EXPECT().IsWindowsPrefixDelegationEnabled().Return(false)
 	mock.MockK8sAPI.EXPECT().GetConfigMap(config.VpcCniConfigMapName, config.KubeSystemNamespace).Return(createCoolDownMockCM("30"), nil).AnyTimes()
--- a/docs/windows/secondary_ip_mode_config_options.md
+++ b/docs/windows/secondary_ip_mode_config_options.md
@@ -1,0 +1,63 @@
+# Configuration options when using secondary IP addresses Windows
+
+We provide multiple configuration options that allow you to fine-tune the IP address allocation behavior on Windows
+nodes using the secondary IP address mode. These configuration options can be set in the `amazon-vpc-cni` ConfigMap in
+the `kube-system` namespace.
+
+- `windows-warm-ip-target` → The total number of IP addresses that should be allocated to each Windows node in excess of
+  the current need at any given time. The excess IPs can be used by newly launched pods, which aids in faster pod
+  startup times since there is no wait time for additional IP addresses to be allocated. The VPC Resource Controller
+  will attempt to ensure that this excess desired threshold is always met.
+
+  Defaults to 3 if unspecified or invalid. Must be greater than or equal to 1.
+
+  For example, if no pods were running on a given Windows node, and if you set `windows-warm-ip-target` to 5, the VPC
+  Resource Controller will aim to ensure that each Windows node always has at least 5 IP addresses in excess, ready for
+  use, allocated to its ENI. If 2 pods are scheduled on the node, the controller will allocate 2 additional IP addresses
+  to the ENI, maintaining the 5 warm IP address target.
+
+- `windows-minimum-ip-target` → Defaults to 3 if unspecified or invalid. The minimum number of IP addresses, both in use
+  by running pods and available as warm IPs, that should be allocated to each Windows node at any given time. The
+  controller will attempt to ensure that this minimum threshold is always met.
+
+  Defaults to 3 if unspecified or invalid. Must be greater than or equal to 0.
+
+  For example, if no pods were running on a given Windows node, and if you set `windows-minimum-ip-target` to 10, the
+  VPC Resource Controller will aim to ensure that the total number of IP addresses on the Windows node should be at
+  least 10. Therefore, before pods are scheduled, there should be at least 10 IP addresses available. If 5 pods are
+  scheduled on a given node, they will consume 5 of the 10 available IPs. The VPC Resource Controller will keep 5 the
+  remaining available IPs available in addition to the 5 already in use to meet the target of 10.
+
+### Considerations while using the above configuration options
+
+- These configuration options only apply when the VPC Resource Controller is operating in the secondary IP mode. They do
+  not affect the prefix delegation mode. More explicitly, if `enable-windows-prefix-delegation` is set to false, or is
+  not specified, then the VPC Resource Controller operates in secondary IP mode.
+- Setting either `windows-warm-ip-target` or `windows-minimum-ip-target` to a negative value will result in the
+  respective default value being used.
+- If the values of `windows-warm-ip-target` or `windows-minimum-ip-target` are set such that the maximum node IP
+  capacity would be exceeded, the controller will limit the allocation to the maximum capacity possible.
+- The `warm-prefix-target` configuration option will be ignored when using the secondary IP mode, as it only applies to
+  the prefix delegation mode.
+- If `windows-warm-ip-target` is set to 0, the system will implicitly set `windows-warm-ip-target` to 1. This is
+  because on-demand IP allocation whereby an IP is allocated on the Windows node as the pods are scheduled is currently
+  not supported. Implicitly Setting `windows-warm-ip-target` to 1 ensures the minimum acceptable non-zero value is set
+  since the `windows-warm-ip-target` should always be at least 1.
+- The configuration options `warm-ip-target` and `minimum-ip-target` are deprecated in favor of the new
+  options `windows-warm-ip-target` and `windows-minimum-ip-target`.
+
+### Examples
+
+| `windows-warm-ip-target` | `windows-minimum-ip-target` | Running Pods | Total Allocated IPs | Warm IPs |
+|--------------------------|-----------------------------|--------------|---------------------|----------|
+| 1                        | 0                           | 0            | 1                   | 1        |
+| 1                        | 0                           | 5            | 6                   | 1        |
+| 5                        | 0                           | 0            | 5                   | 5        |
+| 1                        | 1                           | 0            | 1                   | 1        |
+| 1                        | 1                           | 1            | 2                   | 1        |
+| 1                        | 3                           | 3            | 4                   | 1        |
+| 1                        | 3                           | 5            | 6                   | 1        |
+| 5                        | 10                          | 0            | 10                  | 10       |
+| 10                       | 10                          | 0            | 10                  | 10       |
+| 10                       | 10                          | 10           | 20                  | 10       |
+| 15                       | 10                          | 10           | 25                  | 15       |
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -97,7 +97,7 @@ func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (wa
 	if vpcCniConfigMap.Data == nil {
 		warmIPTarget = IPv4DefaultWinWarmIPTarget
 		minIPTarget = IPv4DefaultWinMinIPTarget
-		log.V(1).Info(
+		log.Info(
 			"No ConfigMap data found, falling back to using default values",
 			"minIPTarget", minIPTarget,
 			"warmIPTarget", warmIPTarget,
@@ -126,46 +126,36 @@ func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (wa
 
 	// If warm IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundWarmIP {
-		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
+		warmIPTarget, err = strconv.Atoi(warmIPTargetStr)
 		if err != nil {
 			log.Info("Could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
-			warmIPTarget = 0
-		} else {
-			warmIPTarget = warmIPTargetInt
-
+		} else if !isPDEnabled && warmIPTarget == 0 {
 			// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
 			// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
-			if !isPDEnabled && warmIPTarget == 0 {
-				log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
-				warmIPTarget = 1
-			}
+			log.Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
+			warmIPTarget = 1
 		}
 	} else {
-		log.V(1).Info("could not find warm ip target in ConfigMap, defaulting to zero")
+		log.Info("could not find warm ip target in ConfigMap, defaulting to zero")
 		warmIPTarget = 0
 	}
 
 	// If min IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundMinIP {
-		minIPTargetInt, err := strconv.Atoi(minIPTargetStr)
+		minIPTarget, err = strconv.Atoi(minIPTargetStr)
 		if err != nil {
 			log.Info("Could not parse minimum ip target, defaulting to zero", "minimum ip target", minIPTargetStr)
-			minIPTarget = 0
-		} else {
-			minIPTarget = minIPTargetInt
 		}
 	} else {
-		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		log.Info("could not find minimum ip target in ConfigMap, defaulting to zero")
 		minIPTarget = 0
 	}
 
 	warmPrefixTarget = 0
 	if isPDEnabled && foundWarmPrefix {
-		warmPrefixTargetInt, err := strconv.Atoi(warmPrefixTargetStr)
+		warmPrefixTarget, err = strconv.Atoi(warmPrefixTargetStr)
 		if err != nil {
 			log.Info("Could not parse warm prefix target, defaulting to zero", "warm prefix target", warmPrefixTargetStr)
-		} else {
-			warmPrefixTarget = warmPrefixTargetInt
 		}
 	}
 
@@ -178,7 +168,7 @@ func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (wa
 			minIPTarget = IPv4DefaultWinMinIPTarget
 			warmIPTarget = IPv4DefaultWinWarmIPTarget
 		}
-		log.V(1).Info(
+		log.Info(
 			"Encountered zero values for warm-ip-target, min-ip-target and warm-prefix-target in ConfigMap data, falling back to using default values since on demand IP allocation is not supported",
 			"minIPTarget", minIPTarget,
 			"warmIPTarget", warmIPTarget,
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -28,13 +28,14 @@ const (
 	// Default Configuration for Pod ENI resource type
 	PodENIDefaultWorker = 30
 
-	// Default Configuration for IPv4 resource type
-	IPv4DefaultWorker  = 2
-	IPv4DefaultWPSize  = 3
-	IPv4DefaultMaxDev  = 1
-	IPv4DefaultResSize = 0
+	// Default Windows Configuration for IPv4 resource type
+	IPv4DefaultWinWorkerCount  = 2
+	IPv4DefaultWinWarmIPTarget = 3
+	IPv4DefaultWinMinIPTarget  = 3
+	IPv4DefaultWinMaxDev       = 0
+	IPv4DefaultWinResSize      = 0
 
-	// Default Configuration for IPv4 prefix resource type
+	// Default Windows Configuration for IPv4 prefix resource type
 	IPv4PDDefaultWorker               = 2
 	IPv4PDDefaultWPSize               = 1
 	IPv4PDDefaultMaxDev               = 0
@@ -70,26 +71,44 @@ func LoadResourceConfig() map[string]ResourceConfig {
 func LoadResourceConfigFromConfigMap(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) map[string]ResourceConfig {
 	resourceConfig := getDefaultResourceConfig()
 
-	warmIPTarget, minIPTarget, warmPrefixTarget := ParseWinPDTargets(log, vpcCniConfigMap)
+	warmIPTarget, minIPTarget, warmPrefixTarget, isPDEnabled := ParseWinIPTargetConfigs(log, vpcCniConfigMap)
 
 	// If no PD configuration is set in configMap or none is valid, return default resource config
 	if warmIPTarget == 0 && minIPTarget == 0 && warmPrefixTarget == 0 {
 		return resourceConfig
 	}
 
-	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmIPTarget = warmIPTarget
-	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.MinIPTarget = minIPTarget
-	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget
+	if isPDEnabled {
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmIPTarget = warmIPTarget
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.MinIPTarget = minIPTarget
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget
+	} else {
+		resourceConfig[ResourceNameIPAddress].WarmPoolConfig.WarmIPTarget = warmIPTarget
+		resourceConfig[ResourceNameIPAddress].WarmPoolConfig.MinIPTarget = minIPTarget
+	}
 
 	return resourceConfig
 }
 
-// ParseWinPDTargets parses config map for Windows prefix delegation configurations set by users
-func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTarget int, minIPTarget int, warmPrefixTarget int) {
-	warmIPTarget, minIPTarget, warmPrefixTarget = 0, 0, 0
-
+// ParseWinIPTargetConfigs parses Windows IP target configuration parameters in the amazon-vpc-cni ConfigMap
+// If all three config parameter values (warm-ip-target, min-ip-target, warm-prefix-target) are 0 or unset, or config map does not exist,
+// then default values for warm-ip-target and min-ip-target will be set.
+func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTarget int, minIPTarget int, warmPrefixTarget int, isPDEnabled bool) {
 	if vpcCniConfigMap.Data == nil {
-		return warmIPTarget, minIPTarget, warmPrefixTarget
+		warmIPTarget = IPv4DefaultWinWarmIPTarget
+		minIPTarget = IPv4DefaultWinMinIPTarget
+		log.V(1).Info(
+			"No ConfigMap data found, falling back to using default values",
+			"minIPTarget", minIPTarget,
+			"warmIPTarget", warmIPTarget,
+		)
+		return warmIPTarget, minIPTarget, 0, false
+	}
+
+	isPDEnabled, err := strconv.ParseBool(vpcCniConfigMap.Data[EnableWindowsPrefixDelegationKey])
+	if err != nil {
+		log.Info("Could not parse prefix delegation flag from ConfigMap, falling back to using secondary IP mode")
+		isPDEnabled = false
 	}
 
 	warmIPTargetStr, foundWarmIP := vpcCniConfigMap.Data[WarmIPTarget]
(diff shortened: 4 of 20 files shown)


pkg/config/loader.go
```diff
@@ -28,13 +28,14 @@ const (
...
-	// Default Configuration for IPv4 resource type
-	IPv4DefaultWorker  = 2
-	IPv4DefaultWPSize  = 3
-	IPv4DefaultMaxDev  = 1
-	IPv4DefaultResSize = 0
-
-	// Default Configuration for IPv4 prefix resource type
+	// Default Windows Configuration for IPv4 resource type
+	IPv4DefaultWinWorkerCount  = 2
+	IPv4DefaultWinWarmIPTarget = 1
```
tzifudzi, line 33: @haouc @jiechen0826 Any thoughts on optimal default values for `minIPTarget` and `warmIPTarget` for secondary IP mode? I am currently leaning towards having the same default values as Linux uses for consistency.

tzifudzi:
@haouc @jiechen0826 I considered creating a validating webhook as part of this change to validate the configuration values and prevent invalid values from being set. I chose to leave it out to keep the change minimal and do a follow up change separately. Any thoughts on this?

pkg/config/loader.go
jiechen0826: The thing is Windows customers have been getting 3 warm IPs as default in the past, so I would argue for `warmIPTarget`=3 for consistency.

jiechen0826 (commented):
I'm halfway through the changes, but just posting these comments first

pkg/pool/pool.go
```diff
@@ -715,6 +735,80 @@ func (p *pool) getPDDeviation() int {
...
+	resourcesShort = max(resourcesShort, p.warmPoolConfig.MinIPTarget-numAssignedResources)
+
+	// Calculate how many IPs we're over the warm target
+	resourcesOver := max(availableResources-p.warmPoolConfig.WarmIPTarget, 0)
+
+	// Adjust over to not go below the minimum IP target
+	resourcesOver = max(min(resourcesOver, numAssignedResources-p.warmPoolConfig.MinIPTarget), 0)
+
+	// The final deviation is the difference between short and over
+	deviation := resourcesShort - resourcesOver
```
jiechen0826, line 794: Very nice and succinct logic!

jiechen0826 (commented):
Reviewed all files. Ready to approve once the above comments are addressed.

controllers/core/configmap_controller.go
```diff
@@ -116,21 +116,34 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
-		isPDConfigUpdated = true
+	warmIPTarget, minIPTarget, warmPrefixTarget, isPDEnabled := config.ParseWinIPTargetConfigs(r.Log, configmap)
+	var winMinIPTargetUpdated = r.curWinMinIPTarget != minIPTarget
+	var winWarmIPTargetUpdated = r.curWinWarmIPTarget != warmIPTarget
+	var winPDWarmPrefixTargetUpdated = r.curWinPDWarmPrefixTarget != warmPrefixTarget
+	if winWarmIPTargetUpdated || winMinIPTargetUpdated {
+		r.curWinWarmIPTarget = warmIPTarget
+		r.curWinMinIPTarget = minIPTarget
+		r.curWinPDWarmPrefixTarget = warmPrefixTarget
```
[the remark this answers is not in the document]
tzifudzi: Yes correct, nice catch. Resolved now

pkg/config/loader.go
```diff
@@ -105,36 +123,59 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
+			log.Error(err, "failed to parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
+			warmIPTarget = 0
 		} else {
 			warmIPTarget = warmIPTargetInt
 		}
+	} else {
+		log.V(1).Info("warm IP configuration not found in ConfigMap, will use default value")
+		if isPDEnabled {
+			warmIPTarget = IPv4PDDefaultWarmIPTargetSize
```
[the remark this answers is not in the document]
tzifudzi: Thats actually a good consideration. That ensures that the behaviour for prefix delegation mode is preserved as initially intended as specified in https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/docs/windows/prefix_delegation_config_options.md. Will make this code change.
> _The settings for these values would depend upon your use case. If set, warm-ip-target and/or minimum-ip-target will take precedence over warm-prefix-target._

pkg/config/loader.go
tzifudzi: I updated to  `warmIPTarget=3`

pkg/config/loader.go
```diff
@@ -105,36 +123,69 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		minIPTarget = 0
 	}
-	if foundWarmPrefix {
+
+	// If PD is enabled and warm prefix target config value is not found, or there is an error parsing it, the value will be set to zero
+	if !isPDEnabled && foundWarmPrefix {
+		log.V(1).Info("warm prefix configuration not supported in secondary IP mode, will ignore warm prefix configuration")
```
jiechen0826, line 156: Should we set `warmPrefixTarget` to 0 here, as it is needed to make L177 work? In the case of secondary IP mode where `warmIPTarget == 0 && minIPTarget == 0 && warmPrefixTarget != 0` , the code will reset `warmPrefixTarget=0`, and then user will get default values for `warmIPTarget` and `minIPTarget` correctly.

jiechen0826 (commented):
Just one more comment. Btw, there seems to be a unit test failure.

pkg/config/loader.go
tzifudzi: Yes it defaults to zero based on the value set in L96. But probably better to explicitly set to zero. Made the change.

jiechen0826: Ohh gotcha, it was a bit far back so I lost track. Thanks!

jiechen0826 (approved):
LGTM. Thanks for implementing the feature and addressing my comments!

tzifudzi:
@jiechen0826 I have added documentation and integration tests. Please take another look? The logic you reviewed has not changed. 

Note: New integration tests for secondary IP mode are all passing, currently investigating a few failures in PD tests. Once all integration tests are passing I will drop a note on the test output. In the meantime please proceed to review.

jiechen0826 (approved):
LGTM, thanks for adding the integration tests and doc update!

controllers/core/configmap_controller.go
```diff
@@ -116,21 +116,34 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			config.WinMinimumIPTarget, r.curWinMinIPTarget,
+			config.WinWarmPrefixTarget, r.curWinPDWarmPrefixTarget,
+			config.EnableWindowsPrefixDelegationKey, isPDEnabled,
+		)
 	}
 
-	// Flag is updated, update all nodes
-	if isIPAMFlagUpdated || isPrefixFlagUpdated || isPDConfigUpdated {
+	var nodesRequireUpdate = isIPAMFlagUpdated || isPrefixFlagUpdated || isWinIPConfigsUpdated
+	if nodesRequireUpdate {
```
haouc, line 146: Why can't put them inline if the variable is not used anywhere else?

pkg/config/loader.go
```diff
@@ -70,26 +71,45 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
-
+// ParseWinIPTargetConfigs parses Windows IP target configuration parameters in the amazon-vpc-cni ConfigMap
+// If all three config parameter values (warm-ip-target, min-ip-target, warm-prefix-target) are 0 or unset, or config map does not exist,
+// then default values for warm-ip-target and min-ip-target will be set.
+func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTarget int, minIPTarget int, warmPrefixTarget int, isPDEnabled bool) {
 	if vpcCniConfigMap.Data == nil {
-		return warmIPTarget, minIPTarget, warmPrefixTarget
+		warmIPTarget = IPv4DefaultWinWarmIPTarget
+		minIPTarget = IPv4DefaultWinMinIPTarget
+		log.V(1).Info(
```
haouc, line 101: Let's not use v(1) for new logs since log levels wouldn't be configurable anyway.

pkg/config/loader.go
```diff
@@ -70,26 +71,45 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
+			"No ConfigMap data found, falling back to using default values",
+			"minIPTarget", minIPTarget,
+			"warmIPTarget", warmIPTarget,
+		)
+		return warmIPTarget, minIPTarget, 0, false
+	}
+
+	isPDEnabled, err := strconv.ParseBool(vpcCniConfigMap.Data[EnableWindowsPrefixDelegationKey])
```
haouc, line 109: Is this `isPDEnabled` same as the one assigned at ln 74?

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-
+	// If warm IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundWarmIP {
 		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
+			log.V(1).Info("could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
+			warmIPTarget = 0
 		} else {
 			warmIPTarget = warmIPTargetInt
```
haouc, line 135: Looks like this line can be removed.

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-	if !foundWarmIP && !foundMinIP && !foundWarmPrefix {
-		return warmIPTarget, minIPTarget, warmPrefixTarget
-	}
-
+	// If warm IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundWarmIP {
 		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
+			log.V(1).Info("could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
```
haouc, line 132: Same as the other log level. If the log info is important, let;s just use log.Info().

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+			// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
+			// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
+			if !isPDEnabled && warmIPTarget == 0 {
+				log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
+				warmIPTarget = 1
+			}
 		}
+	} else {
+		log.V(1).Info("could not find warm ip target in ConfigMap, defaulting to zero")
+		warmIPTarget = 0
```
haouc, line 146: Is `warmIPTarget` a passed in argument? Why do we change it to `0` here?

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 	}
+
+	// If min IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundMinIP {
 		minIPTargetInt, err := strconv.Atoi(minIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse minimum ip target", "minimum ip target", minIPTargetStr)
+			log.V(1).Info("could not parse minimum ip target, defaulting to zero", "minimum ip target", minIPTargetStr)
```
haouc, line 153: If it is not an error, let's remove the log line.

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		minIPTarget = 0
 	}
-	if foundWarmPrefix {
+
+	// If PD is enabled and warm prefix target config value is not found, or there is an error parsing it, the value will be set to zero
+	if !isPDEnabled && foundWarmPrefix {
+		log.V(1).Info("warm prefix configuration not supported in secondary IP mode, will ignore warm prefix configuration")
```
haouc, line 165: V(1) logs don't help at all. I think this log is not critical for operation and removable?

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 			minIPTarget = minIPTargetInt
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		minIPTarget = 0
 	}
-	if foundWarmPrefix {
+
+	// If PD is enabled and warm prefix target config value is not found, or there is an error parsing it, the value will be set to zero
+	if !isPDEnabled && foundWarmPrefix {
```
haouc, line 164: Look like this can be simplified to 
```
needUpdate := isPDEnabled && foundWarmPrefix
if needUpdate {
    warmPrefixTarget = 0
} else {
    if warmPrefixTargetInt, err := strconv.Atoi(warmPrefixTargetStr); err == nil {
        warmPrefixTarget = warmPrefixTargetInt
    } else {
        warmPrefixTarget = 0
    }
}
```

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 		} else {
 			warmPrefixTarget = warmPrefixTargetInt
 		}
+	} else if isPDEnabled && !foundWarmPrefix {
+		log.V(1).Info("could not find warm prefix target in ConfigMap, defaulting to zero")
+		warmPrefixTarget = 0
 	}
-	return warmIPTarget, minIPTarget, warmPrefixTarget
+
+	if warmIPTarget == 0 && minIPTarget == 0 {
```
haouc, line 180: What if customers intentionally set them to `0`? Does this overwrite their settings?

pkg/config/loader.go
```diff
@@ -105,36 +125,74 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+		if isPDEnabled && warmPrefixTarget == 0 {
+			minIPTarget = IPv4PDDefaultMinIPTargetSize
+			warmIPTarget = IPv4PDDefaultWarmIPTargetSize
+		} else if !isPDEnabled {
+			minIPTarget = IPv4DefaultWinMinIPTarget
+			warmIPTarget = IPv4DefaultWinWarmIPTarget
+		}
+		log.V(1).Info(
+			"No valid configuration values for warm-ip-target, min-ip-target and warm-prefix-target found in ConfigMap, falling back to using default values",
```
haouc, line 189: May not be not valid? Do we explicitly ask customers not to set them to `0`s?

pkg/pool/pool.go
```diff
@@ -423,6 +424,20 @@ func (p *pool) ReconcilePool() *worker.WarmPoolJob {
 		len(p.usedResources), "pending create", p.pendingCreate, "pending delete", p.pendingDelete,
 		"cool down queue", len(p.coolDownQueue), "total resources", totalCreatedResources, "capacity", p.capacity)
 
+	p.log.V(1).Info(
```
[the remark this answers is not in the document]
haouc: Same here, V(1) log doesn't help operations. If this is critical for operations, please set them to log.Info().

pkg/pool/pool.go
```diff
@@ -715,6 +736,73 @@ func (p *pool) getPDDeviation() int {
...
+
+	// warm pool is in draining state, set targets to zero
+	if p.warmPoolConfig.DesiredSize == 0 {
+		p.log.V(1).Info("DesiredSize is zero, warmPool is in draining state")
+		p.warmPoolConfig.WarmIPTarget = 0
+		p.warmPoolConfig.MinIPTarget = 0
+		p.warmPoolConfig.WarmPrefixTarget = 0
+	}
+
+	isMinIPTargetInvalid := p.warmPoolConfig.MinIPTarget < 0
```
haouc, line 754: Why should we have invalid values be passed with p.warmPoolConfig? We should make sure our managed fields valid before set them up.

pkg/pool/pool.go
```diff
@@ -715,6 +736,73 @@ func (p *pool) getPDDeviation() int {
...
+	// Calculate how many IPs we're over the warm target
+	resourcesOver := max(availableResources-p.warmPoolConfig.WarmIPTarget, 0)
+
+	// Adjust over to not go below the minimum IP target
+	resourcesOver = max(min(resourcesOver, numAssignedResources-p.warmPoolConfig.MinIPTarget), 0)
+
+	// The final deviation is the difference between short and over
+	deviation := resourcesShort - resourcesOver
+
+	p.log.V(1).Info(
```
haouc, line 790: I feel instead logging those important settings we can emit them as prometheus gauge metrics.

scripts/test/run-integration-tests.sh
```diff
@@ -42,7 +42,7 @@ function run_integration_tests(){
   TEST_RESULT=success
   (cd $INTEGRATION_TEST_DIR/perpodsg && CGO_ENABLED=0 ginkgo --skip=LOCAL $EXTRA_GINKGO_FLAGS -v -timeout=35m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID) || TEST_RESULT=fail
   if [[ -z "${SKIP_WINDOWS_TEST}" ]]; then
-    (cd $INTEGRATION_TEST_DIR/windows && CGO_ENABLED=0 ginkgo --skip=LOCAL $EXTRA_GINKGO_FLAGS -v -timeout=120m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID) || TEST_RESULT=fail
+    (cd $INTEGRATION_TEST_DIR/windows && CGO_ENABLED=0 ginkgo --skip=LOCAL $EXTRA_GINKGO_FLAGS -v -timeout=180m -- -cluster-kubeconfig=$KUBE_CONFIG_PATH -cluster-name=$CLUSTER_NAME --aws-region=$REGION --aws-vpc-id $VPC_ID) || TEST_RESULT=fail
```
haouc, line 45: We probably need to understand why we have to keep adding time to this test...

controllers/core/configmap_controller.go
tzifudzi: Indeed the variable is not used anywhere else. I separated it into a variable since the conditional expression was becoming more lengthier with the condition I added. I thought it might help to make the code readable by clearly communicating the purpose of the condition. If you would insist, I am okay with changing back to a conditional expression and using a comment.

pkg/config/loader.go
tzifudzi: I acknowledge the concern raised. However in `main.go` I noticed the application allows specifying of the the debug log levels but I do acknowledge that this is not configurable in the control plane by EKS customers.

I would remove the log statements entirely if they had no use, but from my own experience, these debug logs helped me during my debugging while writing code, and also the test suite uses a debug log level which helped me debug failing test cases. I also thought the logs could provide additional context to the code to make it readable though I acknowledge they could be replaced by comments. 

After hearing this response what are your thoughts on removing `log.V(1).Info(...)` logs that I added? Do you still insist they be removed?

pkg/config/loader.go
tzifudzi: Yes that is correct. The returned value will be used in L74. It is used to determine how to appropriately set the `resourceConfig` in `LoadResourceConfigFromConfigMap(...)` depending on whether its PD or secondary IP mode.

pkg/config/loader.go
tzifudzi: Yes it potentially could. Will monitor your response in thread https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1688880904 so I can apply the same precedence in other log statements.

pkg/config/loader.go
tzifudzi: Acknowledged. See response in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690063506

pkg/config/loader.go
tzifudzi: Good question as the behaviour intended is something I would want your feedback on. The workflow will be as follows which is why zero is set where you highlighted.

1. Retrieve the configuration parameters from the ConfigMap.
1. Check if config map does exists. If exists proceed otherwise return default values
1. Check if all three config parameter values (warm-ip-target, min-ip-target, warm-prefix-target) are 0 or unset. 
   1. If some values set, then only use the set values and set all other values to zero hence why this statement sets the value to zero. This ensures that the behaviour for prefix delegation mode is preserved as initially intended as specified in https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/docs/windows/prefix_delegation_config_options.md. See thread where @jiechen0826 recommended this as discussed in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1683360844. 
   1. If all values are set to zero then use default values
1. Return the results.

pkg/config/loader.go
tzifudzi: https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690077969

pkg/config/loader.go
tzifudzi: Acknowledged. Please see comment in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690049393

pkg/config/loader.go
tzifudzi: Agreed, your code suggestion is more concise. However it will impact the ability to add the log statements that currently exist. Based on your response to thread https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690063506, I could make the suggested change.

pkg/config/loader.go
tzifudzi: Correct. If customers intentionally set them to 0 their change will be overwritten. As undesirable as this sounds, take note that this is actually how prefix delegation currently works. It does not work with zero values. If you specify zero values they will be replaced with default values even with the current implementation.

tzifudzi: With this PR change, in the case of secondary IP mode, this condition `warmIPTarget == 0 && minIPTarget == 0` will never be true because of this line which forces warm IP to never be configurable to 1. This is because the code base workflow currently relies on warmpool always being populated with at least 1 IP.

```
// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
if !isPDEnabled && warmIPTarget == 0 {
  log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
  warmIPTarget = 1
}
```

pkg/config/loader.go
tzifudzi: Good point! The log statement is misleading that zero values are 'invalid'. Zero values are actually 'valid' values, but the VPC Resource Controller has a limitation that 'on demand' IP allocation where there are no warm IPs is _not_ currently supported. Updated the log statement as such. 

Shameless plug: The way you read this log statement to assume and understand the behaviour is what I meant when I said in this thread https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690049393 that the debug log statements might help both code readability and debugging.

pkg/pool/pool.go
tzifudzi: Indeed the invalid negative values should never make their way into this code logic sequence. It seems redundant to recheck this, but I think its a reasonable defensive coding precaution to avoid unintended behaviour. 

Another suggestion I considered was to change the type of IP target configurations e.g. `MinIPTarget` from `int` to `uint` which only allows positive values but this would require refactoring in multiple places and I wanted to keep this change minimal.

pkg/pool/pool.go
tzifudzi: I appreciate the suggestion, I actually think its a good suggestion as it would improve observability of the configuration value settings. However this would be additional work that I would need to thoroughly test and given my current time constraints I may not be able to get to this immediately. May I please suggest adding this suggestion as a technical debt item in the backlog to tackle at a later time?

scripts/test/run-integration-tests.sh
tzifudzi: I also hate to increase the testing time. In this case I am increasing the time because I added new integration tests which add about 20 minutes give or take or testing time.

What would you suggest as an appropriate timeout? If left as is there is a risk of the tests failing with a timeout error since the integration test suite has been increased with the new tests I added.

tzifudzi:
@sushrk Can you kindly join the review? I don't have access to edit the list of reviewers to add you.

scripts/test/run-integration-tests.sh
sushrk: If it is a 20 min increase in test time, let's limit it to 150m total which also gives 10m buffer. Would that be more appropriate?

pkg/config/loader.go
sushrk: > this condition warmIPTarget == 0 && minIPTarget == 0 will never be true

Do we need this code block then if we never expect to set warm IP and min IP target to be 0?

sushrk: Seems this is added for future cases when warm IP & min IP can be set to zero, please confirm.

test/README.md
```diff
@@ -63,7 +63,7 @@ The Integration test suite provides the following focuses.
    echo "Running Security Group for Pods Integration Tests"
    (cd perpodsg && CGO_ENABLED=0 GOOS=$OS ginkgo -v --timeout 40m -- --cluster-kubeconfig=$KUBE_CONFIG_PATH --cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id=$VPC_ID)
    echo "Running Windows Integration Tests"
-   (cd windows && CGO_ENABLED=0 GOOS=$OS ginkgo -v --timeout 40m -- --cluster-kubeconfig=$KUBE_CONFIG_PATH --cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id=$VPC_ID)
+   (cd windows && CGO_ENABLED=0 GOOS=$OS ginkgo -v --timeout 180m -- --cluster-kubeconfig=$KUBE_CONFIG_PATH --cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id=$VPC_ID)
```
sushrk, line 66: Let's also update this to 150m if it is sufficient time to run all tests, thanks.

pkg/pool/pool.go
sushrk: +1 for adding it as Prometheus metrics. If you want to create a backlog and make the changes in a follow-up PR, that would be appreciated, thanks!

sushrk (commented):
Thank you for the PR and implementing this feature. I had a few minor comments, but would like to sync up with you offline to understand when and how we set the default values for warm IP target and minimum IP target.

pkg/config/loader.go
tzifudzi: Adding more context to this thread to mention that setting zero values for both `warmIPTarget` and `minIPTarget` is not supported in VPC Resource Controller and will result in default values taking effect. This was the behavior prior to my change and will remain the behavior after the change. Having zero values to result in pod IPs allocated 'on demand' is not yet supported in VPC RC. This is also the case for [Amazon VPC CNI for Linux](https://github.com/aws/amazon-vpc-cni-k8s). 

Adding this code was part of my refactoring to consolidate the logic for this check for prefix delegation and secondary IP mode in one code block.

tzifudzi: > Do we need this code block then if we never expect to set warm IP and min IP target to be 0?

> Seems this is added for future cases when warm IP & min IP can be set to zero, please confirm.

@sushrk Good question and good guess. The need for that code code is twofold
1. For the code to act defensively and handle the edge case where the intended behaviour (to enforce minimum `warmIPTarget` of at least one) does not correctly get effected. VPC RC relies on the concept of having a warmpool of at least 1, so having both `warmIPTarget` and `minIPTarget` set as zero would result in pods never getting allocated an IP and this is to defend against this scenario.
2. As you rightly guessed, in the future, the intention would be to allow customers to have `warmIPTarget` and `minIPTarget` settable to zero so that on demand IP allocation (without warm IPs) for pods is possible. This would change is in preparation for such a future change.

pkg/pool/pool.go
tzifudzi: @sushrk Thanks for the feedback. I created this issue to track the work and am willing to take it up https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/456

scripts/test/run-integration-tests.sh
tzifudzi: I think 150m is a good compromise. Setting it to 150m.

controllers/core/configmap_controller_test.go
```diff
@@ -127,7 +141,7 @@ func Test_Reconcile_ConfigMap_PD_Disabled_If_IPAM_Disabled(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	mock := NewConfigMapMock(ctrl, mockConfigMapPD)
```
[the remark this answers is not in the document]
tzifudzi: > Seems mockConfigMapPD is no longer used in this file, let's delete it if not used.

Correct. I removed it

pkg/config/loader.go
```diff
@@ -105,36 +125,77 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-		return warmIPTarget, minIPTarget, warmPrefixTarget
-	}
-
+	// If warm IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundWarmIP {
 		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
+			log.Info("Could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
+			warmIPTarget = 0
```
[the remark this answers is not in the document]
tzifudzi: The reason to set to zero is to ensure a specific requirement continues to work as expected for prefix delegation. See this thread https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1683360844 where this was discussed. In summary if we set the default of 3 at this point, intended behavior for prefix delegation to allow setting of either the preefix target or the warm IP won't take effect as intended.

pkg/config/loader.go
```diff
@@ -105,36 +125,77 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+			// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
+			// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
+			if !isPDEnabled && warmIPTarget == 0 {
+				log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
+				warmIPTarget = 1
+			}
 		}
+	} else {
+		log.V(1).Info("could not find warm ip target in ConfigMap, defaulting to zero")
+		warmIPTarget = 0
```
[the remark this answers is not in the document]
tzifudzi: See response https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1728404991

controllers/core/configmap_controller.go
```diff
@@ -46,8 +46,8 @@ type ConfigMapReconciler struct {
 	Condition                         condition.Conditions
 	curWinIPAMEnabledCond             bool
 	curWinPrefixDelegationEnabledCond bool
-	curWinPDWarmIPTarget              int
-	curWinPDMinIPTarget               int
+	curWinWarmIPTarget                int
```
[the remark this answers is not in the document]
tzifudzi: I tried correcting it using both IDE auto linting and also manually removing the whitespace. The extra whitespace in the code is necessary to maintain the consistent alignment of the struct field variable names. If I remove the space the fields lose alignment. Marking this as resolved.

pkg/pool/pool.go
```diff
@@ -715,6 +736,73 @@ func (p *pool) getPDDeviation() int {
...
+	numUsedResources := len(p.usedResources)
+	numAssignedResources := numUsedResources + numWarmResources + p.pendingCreate + len(p.coolDownQueue)
+
+	// warm pool is in draining state, set targets to zero
+	if p.warmPoolConfig.DesiredSize == 0 {
+		p.log.V(1).Info("DesiredSize is zero, warmPool is in draining state")
+		p.warmPoolConfig.WarmIPTarget = 0
+		p.warmPoolConfig.MinIPTarget = 0
+		p.warmPoolConfig.WarmPrefixTarget = 0
+	}
```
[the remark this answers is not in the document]
tzifudzi: When warmpool is in draining state its any scenario whereby the warm pool needs to delete all its resources. An example is if toggling from secondary IP to PD mode and we want to delete all prefixes.

pkg/config/loader.go
```diff
@@ -105,36 +125,77 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 		if err != nil {
-			log.Error(err, "failed to parse minimum ip target", "minimum ip target", minIPTargetStr)
+			log.Info("Could not parse minimum ip target, defaulting to zero", "minimum ip target", minIPTargetStr)
+			minIPTarget = 0
 		} else {
 			minIPTarget = minIPTargetInt
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		minIPTarget = 0
```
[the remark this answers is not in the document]
tzifudzi: See response https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443/files#r1728405335

pkg/config/loader.go
```diff
@@ -70,26 +71,45 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
-	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.MinIPTarget = minIPTarget
-	resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget
+	if isPDEnabled {
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmIPTarget = warmIPTarget
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.MinIPTarget = minIPTarget
+		resourceConfig[ResourceNameIPAddressFromPrefix].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget
+	} else {
+		resourceConfig[ResourceNameIPAddress].WarmPoolConfig.WarmIPTarget = warmIPTarget
+		resourceConfig[ResourceNameIPAddress].WarmPoolConfig.MinIPTarget = minIPTarget
+		resourceConfig[ResourceNameIPAddress].WarmPoolConfig.WarmPrefixTarget = warmPrefixTarget // ignore warm prefix in secondary IP mode
```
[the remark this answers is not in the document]
tzifudzi: It doesn't need to be set. For completeness I explicitly added it. To avoid unintended behaviour I have since removed this line.

tzifudzi:
@sushrk I have responded to all your recent comments, so please take another look and re-review. Also take note of the following
- I rebased from master to get latest code changes. 
- I reran all unit tests and they are still all passing 
- I re-ran integration tests. I am currently investigating some failures and might need to do another code push as a follow up but please still proceed in the meantime.
- Locally running make presubmit results in changes to the auto generated files such as CRDs and RBAC configurations. I committed these changes so the presubmit works but I see this PR still is failing. This issue is also occurring in other PRs such as this https://github.com/aws/amazon-vpc-resource-controller-k8s/actions/runs/10189822604/job/28188550200?pr=450 so if you have time it would be good if this can be resolved outside this PR

controllers/core/configmap_controller_test.go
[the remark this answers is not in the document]
tzifudzi: > On a side note, we are missing tests cases for when IPAM is enabled, PD is disabled, but IPAM config is updated. Can we add this test case?

Good catch, thanks for that suggestion. I have since added a test for the scenario you descrived.

pkg/config/loader.go
```diff
@@ -28,13 +28,14 @@ const (
...
-	// Default Configuration for IPv4 resource type
-	IPv4DefaultWorker  = 2
-	IPv4DefaultWPSize  = 3
-	IPv4DefaultMaxDev  = 1
-	IPv4DefaultResSize = 0
-
-	// Default Configuration for IPv4 prefix resource type
+	// Default Windows Configuration for IPv4 resource type
+	IPv4DefaultWinWorkerCount  = 2
+	IPv4DefaultWinWarmIPTarget = 3
```
sushrk, line 33: Seems the warm IP target default is updated from 1 to 3. Could we update the doc [here](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/docs/windows/prefix_delegation_config_options.md?plain=1#L22) too? 

And could you please add a blurb on the update in the PR description for future references?

pkg/config/loader.go
sushrk: I agree this can be further simplified, maybe like
```
warmPrefixTarget = 0
if isPDEnabled && foundWarmPrefix {
  if warmPrefixTargetInt, err := strconv.Atoi(warmPrefixTargetStr); err == nil {
    warmPrefixTarget = warmPrefixTargetInt
  }
}
```
I'm not convinced we need to have a log for each scenario. We can add a comment at the top and `log.Info` at the end of the function to print the values of each config variable. Let me know what you think.

docs/windows/secondary_ip_mode_config_options.md
```diff
@@ -0,0 +1,63 @@
+# Configuration options when using secondary IP addresses Windows
```
sushrk, line 1: Nice doc!

pkg/config/loader.go
sushrk: Thanks for the explanation, makes sense to me, we're good to resolve this thread.

sushrk:
General feedback regarding the logs- please ensure all important logs are at `log.Info` level, I'm okay to have `log.V(1).Info` for debug purpose as you mentioned above. 
After confirming, feel free to resolve the comments. Thank you.

pkg/config/loader.go
tzifudzi: I decided to accept this feedback. Acknowledged on not needing to have a log for each scenario. Resolving code after recent code push resolves this.

tzifudzi:
@sushrk I have responded to all your recent comments and pushed two new commits so please review the new commits. At merge I will squash the commits into one, but left the commits so you can review them. 

Other notes:
- Reran all unit tests and they are still all passing
- Reran all Windows integration tests and they are still all passing

sushrk (approved):
Thanks for addressing the comments. Looks good to me!

pkg/pool/pool.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
tzifudzi: @haouc I am resolving the conversation to unblock the PR. Let me know if any concerns.

pkg/config/loader.go
```diff
@@ -105,36 +124,70 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
+			log.Info("Could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
+			warmIPTarget = 0
 		} else {
 			warmIPTarget = warmIPTargetInt
+
+			// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
+			// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
+			if !isPDEnabled && warmIPTarget == 0 {
+				log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
```
haouc, line 139: If these logs are for V(1), let's just remove them since V(1) doesn't help when the controller is running in EKS managed control plane.

pkg/config/loader.go
haouc: If you think logs are critical for locating customers' issue, we should use info(). For dev debugging and developping, we can always add temporary log lines to expedite the process but not adding them in prod code for future possible dev/debug purpose.

pkg/config/loader.go
```diff
@@ -105,36 +124,70 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 	}
 
-	// If no configuration is found, return 0
-	if !foundWarmIP && !foundMinIP && !foundWarmPrefix {
-		return warmIPTarget, minIPTarget, warmPrefixTarget
-	}
-
+	// If warm IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundWarmIP {
 		warmIPTargetInt, err := strconv.Atoi(warmIPTargetStr)
```
haouc, line 129: Looks like no need to create a new local variable `warmIPTargetInt`.

pkg/config/loader.go
```diff
@@ -105,36 +124,70 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+			}
 		}
+	} else {
+		log.V(1).Info("could not find warm ip target in ConfigMap, defaulting to zero")
+		warmIPTarget = 0
 	}
+
+	// If min IP target config value is not found, or there is an error parsing it, the value will be set to zero
 	if foundMinIP {
 		minIPTargetInt, err := strconv.Atoi(minIPTargetStr)
```
haouc, line 150: Same here. This local variable can be avoided.

pkg/config/loader.go
```diff
@@ -105,36 +124,70 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
+		minIPTarget = 0
 	}
-	if foundWarmPrefix {
+
+	warmPrefixTarget = 0
+	if isPDEnabled && foundWarmPrefix {
 		warmPrefixTargetInt, err := strconv.Atoi(warmPrefixTargetStr)
```
haouc, line 164: Also this one `warmPrefixTargetInt`

pkg/config/loader.go
```diff
@@ -105,36 +124,70 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+	if warmIPTarget == 0 && minIPTarget == 0 {
+		if isPDEnabled && warmPrefixTarget == 0 {
+			minIPTarget = IPv4PDDefaultMinIPTargetSize
+			warmIPTarget = IPv4PDDefaultWarmIPTargetSize
+			warmPrefixTarget = IPv4PDDefaultWarmPrefixTargetSize
+		} else if !isPDEnabled {
+			minIPTarget = IPv4DefaultWinMinIPTarget
+			warmIPTarget = IPv4DefaultWinWarmIPTarget
+		}
+		log.V(1).Info(
```
haouc, line 181: Looks like a log valid case for log.Info()?

pkg/config/type.go
```diff
@@ -155,6 +155,7 @@ type ResourceConfig struct {
 
 // WarmPoolConfig is the configuration of Warm Pool of a resource
 type WarmPoolConfig struct {
+	// TODO: Deprecate DesiredSize in favour of using WarmIPTarget since historically they served the same purpose
```
haouc, line 158: Is that possible to handle the conversion automatically instead of exposing two vraiables for same purpose?

pkg/pool/pool.go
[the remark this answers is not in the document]
haouc: This looks like a one-off log when a CRUD event happens. I think we can log this as log.Info() for better visibility especially customers reach out.

pkg/pool/pool.go
```diff
@@ -715,6 +736,73 @@ func (p *pool) getPDDeviation() int {
...
+// calculateSecondaryIPDeviation calculates the deviation required to meet the desired state for secondary IP mode
+// Returns a number of IPv4 addresses by taking into account the MinIPTarget and WarmIPTarget
+func (p *pool) calculateSecondaryIPDeviation() int {
+	numWarmResources := numResourcesFromMap(p.warmResources)
+	numUsedResources := len(p.usedResources)
+	numAssignedResources := numUsedResources + numWarmResources + p.pendingCreate + len(p.coolDownQueue)
+
+	// warm pool is in draining state, set targets to zero
+	if p.warmPoolConfig.DesiredSize == 0 {
+		p.log.V(1).Info("DesiredSize is zero, warmPool is in draining state")
```
haouc, line 748: Not sure if these V(1) logs are purely dev purpose. If yes, we can remove them. If they are critical to denug the flow, we should use `log.Info()`.

pkg/pool/pool.go
```diff
@@ -715,6 +736,73 @@ func (p *pool) getPDDeviation() int {
...
+	// Handle scenario where WarmIPTarget is configured to negative integer which is invalid
+	if isWarmIPTargetInvalid {
+		p.log.V(1).Info(
+			"WarmIPTarget value is invalid negative integer, setting warmIPTarget to default",
+			"IPv4DefaultWinWarmIPTarget", config.IPv4DefaultWinWarmIPTarget,
+		)
+		p.warmPoolConfig.WarmIPTarget = config.IPv4DefaultWinWarmIPTarget
+	}
+
+	availableResources := numWarmResources + p.pendingCreate - p.pendingDelete
```
haouc, line 773: `pendingDelete` may not be highly reliable. They could be raced in large scale. By the way, are all those calculations in single goroutine? If the controller support multiple workers, how the atomicity could be guaranteed?

pkg/provider/provider.go
```diff
@@ -46,3 +50,21 @@ type ResourceProvider interface {
 	IntrospectSummary() interface{}
 	ReconcileNode(nodeName string) bool
 }
+
+// GetWinWarmPoolConfig retrieves Windows warmpool configuration from ConfigMap, falls back to using default values on failure
+func GetWinWarmPoolConfig(log logr.Logger, w api.Wrapper, isPDEnabled bool) *config.WarmPoolConfig {
```
haouc, line 55: Why should we have this shared function in the provider interface file?

pkg/config/loader.go
tzifudzi: This particular discussion point and similar comments have been ongoing for a while now in this PR. I gave feedback for this in comment https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#discussion_r1690049393. Also see https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443#issuecomment-2318561034 where a consensus was reached regarding this.

I will reach out to you for a offline chat to hopefully speedily resolve this and reach a consensus rather than going back and forth via comments in threads. My reasons have been summarized in separate threads on why I thought they still have value. Ultimately, as the maintainer of the repo if you insist there is no value in these logs I will respect that and remove them.

pkg/config/loader.go
tzifudzi: Correct thanks. After some previous refactoring it can be a local variable. Same goes for other two occurrences.

pkg/config/loader.go
tzifudzi: I agree. This could be very useful to debug some customer issues i.e. why certain values were enforced.

pkg/config/type.go
tzifudzi: Yes, I think its possible to remove the variable completely actually. I actually did attempt to do this and refactor during my change but I realised the scope of the change would expand. The reason is that `DesiredSize` is used in multiple places throughout the codebase, and fully removing it would require refactoring across several files and tests. This increased scope will increase the complexity of the current change, which could in turn raise the risk of introducing unintended bugs or regressions. To expedite the change I propose making this a technical debt item rather than addressing it in this pull request.

pkg/provider/provider.go
tzifudzi: I have since migrated the function to a `util.go` file in the pool package.

pkg/pool/pool.go
tzifudzi: > `pendingDelete` may not be highly reliable. They could be raced in large scale.

Noted. I have removed it from the calculation

tzifudzi: > ...are all those calculations in single goroutine? 

Yes, all calculations are performed within a single goroutine so a race condition isn't expected except with another worker.

> If the controller support multiple workers, how the atomicity could be guaranteed?

I like this kind of PR feedback - its a valid question to ask to help me implement the change correclty. I am going to assume when you say multiple workers you aren't referring to node workers but rather a worker being a controller deployment worker pod.

Short answer is that if the VPC RC changes moves to support multiple workers, then atomicity won't be guaranteed for my changes. In that case, we would need to make some modifications to ensure atomicity so concurrent updates from multiple workers doesn't happen. Some ideas to achieve will likely be some form of distributed locking when making EC2 API requests that mutate the state of the node/ENI.

That being said, as far as I know, **the current implementation uses a single worker scenario**. This is also how prefix delegation mode currently works as it assumes a single worker. The PD mode code will also succumbs to a similar limitation if multiple workers are supported. When using a single leader model, atomicity should not be an issue. 

To provide further confidence on this change is not being negatively impactful, note that in the event of an interrupted worker, the remaining worker can still read the current current state by using the configmap values and ENI state as the source of truth so there wont be an issue with a leader failing mid-update.

Thanks for bringing up the good question. Let me know if you have any other feedback or concerns.

pkg/config/type.go
tzifudzi: Resolving this after speaking to Hao offline. The consensus is that regardless of logs being potentially useful for troubleshooting, 
- if the log is deemed critical use log.info(...)
- if the log is only useful for debugging, it should be removed

pkg/pool/pool.go
tzifudzi: Resolving this after speaking to Hao offline. The consensus is that regardless of logs being potentially useful for troubleshooting,

if the log is deemed critical use log.info(...)
if the log is only useful for debugging, it should be removed

pkg/pool/pool.go
```diff
@@ -715,6 +721,51 @@ func (p *pool) getPDDeviation() int {
 	return deviationPrefix * NumIPv4AddrPerPrefix
 }
 
+// calculateSecondaryIPDeviation calculates the deviation required to meet the desired state for secondary IP mode
+// Returns a number of IPv4 addresses by taking into account the MinIPTarget and WarmIPTarget
+func (p *pool) calculateSecondaryIPDeviation() int {
```
haouc, line 726: To be honest, I am not convinced yet this is thread safe. This function is invoked from a queued work which is executed by async workers.

pkg/pool/pool.go
tzifudzi: Spoke with Hao on this. He wasn't referring to the controller deployments but rather was referring to the struct fields being accessed and in particular the map that is not safe for concurrent use across different goroutines. 

Currently this risk is mitigated by a mutex lock wrapping the entire reconciliation operation but we also discussed that a better implementation might be to wrap map operations with a lock instead rather than making the entire operation lock access to the object. This requires more thought and further refactoring so it has been tabled for a future implementation. 

Resolving conversation.

tzifudzi:
@haouc I have responded to your latest feedback and pushed new code changes in commit https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/443/commits/e466622ec6ba76b87944080164db1f2cbfe3406b. On merge I will squash changes into one commit. Please  re-review.

pkg/config/loader.go
```diff
@@ -70,26 +71,44 @@ func LoadResourceConfig() map[string]ResourceConfig {
...
-
+// ParseWinIPTargetConfigs parses Windows IP target configuration parameters in the amazon-vpc-cni ConfigMap
+// If all three config parameter values (warm-ip-target, min-ip-target, warm-prefix-target) are 0 or unset, or config map does not exist,
+// then default values for warm-ip-target and min-ip-target will be set.
+func ParseWinIPTargetConfigs(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTarget int, minIPTarget int, warmPrefixTarget int, isPDEnabled bool) {
 	if vpcCniConfigMap.Data == nil {
-		return warmIPTarget, minIPTarget, warmPrefixTarget
+		warmIPTarget = IPv4DefaultWinWarmIPTarget
+		minIPTarget = IPv4DefaultWinMinIPTarget
+		log.V(1).Info(
```
haouc, line 100: nit: can we make this to log.Info(). Sounds to me nil `Data` indicates either misconfiguration or customer's intended setting. We need some way to trace back to historic settings.

pkg/config/loader.go
```diff
@@ -105,36 +124,60 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+		warmIPTarget, err = strconv.Atoi(warmIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse warm ip target", "warm ip target", warmIPTargetStr)
-		} else {
-			warmIPTarget = warmIPTargetInt
+			log.Info("Could not parse warm ip target, defaulting to zero", "warm ip target", warmIPTargetStr)
+		} else if !isPDEnabled && warmIPTarget == 0 {
+			// Handle secondary IP mode scenario where WarmIPTarget is explicitly configured to zero
+			// In such a case there must always be 1 warm IP to ensure that the warmpool is never empty
+			log.V(1).Info("Explicitly setting WarmIPTarget zero value not supported in secondary IP mode, will override with 1")
```
haouc, line 135: nit: Is this qualified for a historic settings record?

pkg/config/loader.go
```diff
@@ -105,36 +124,60 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
+			warmIPTarget = 1
 		}
+	} else {
+		log.V(1).Info("could not find warm ip target in ConfigMap, defaulting to zero")
```
haouc, line 139: nit: same here. It could be hard for us to track back historic config in the configmap audit logs if we need help customers to identify the cause.

pkg/config/loader.go
```diff
@@ -105,36 +124,60 @@ func ParseWinPDTargets(log logr.Logger, vpcCniConfigMap *v1.ConfigMap) (warmIPTa
...
-		minIPTargetInt, err := strconv.Atoi(minIPTargetStr)
+		minIPTarget, err = strconv.Atoi(minIPTargetStr)
 		if err != nil {
-			log.Error(err, "failed to parse minimum ip target", "minimum ip target", minIPTargetStr)
-		} else {
-			minIPTarget = minIPTargetInt
+			log.Info("Could not parse minimum ip target, defaulting to zero", "minimum ip target", minIPTargetStr)
 		}
+	} else {
+		log.V(1).Info("could not find minimum ip target in ConfigMap, defaulting to zero")
```
haouc, line 150: nit: same as the above comment.

pkg/config/loader.go
tzifudzi: Yes precisely so. Changed logging level as suggested

haouc:
lgtm, thanks.

haouc (approved):

Issue #442, opened 2024-07-16
Show correct vpc-resource-controller version in the userAgent


ClareCat:
Brought this up with @orsenthil over slack

**What would you like to be enhanced**:
The cloudwatch EKS cluster logs for vpc-resource-controller actions shows the userAgent field as:
```"userAgent": "controller/v0.0.0 (linux/amd64) kubernetes/$Format",```

It would be better if instead of always showing v0.0.0 the logs showed which version of the vpc-resource-controller made the request.

**Why is the change needed and what use case will it solve**:

It would have been simpler for us to track and report the v1.5.0 trunk initialization issues if the userAgent version mapped to the version of the vpc-resource-controller that made the request.

haouc:
@ClareCat thanks for reporting this. We have correct agent version showing in EC2 dashboards. I am not sure why CloudWatch doesn't have version correctly showing. I will take a look.

orsenthil:
@haouc - I pointed out to the GitVersion that is displayed during the bootstrap ( https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/964a3cce8281e25c185369f63859ed63bedd491f/main.go#L179) . this is different from that log; the request is if the controller version can be set in user agent.

ClareCat:
Yeah, this is the userAgent that is set in the manager client. It's probably an option value in there.

ClareCat:
Right now it seems like it's just using this default: https://github.com/kubernetes/client-go/blob/master/rest/config.go#L491-L505

orsenthil:
I have updated the userAgent to reflect the vpc-resource-controller and gitVersion in this PR https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/455

PR #455, opened 2024-08-21
Set controller user-agent to vpc-resource-controller/git-version

orsenthil:
*Issue #, if available:*

https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/442

```
The cloudwatch EKS cluster logs for vpc-resource-controller actions shows the userAgent field as:
"userAgent": "controller/v0.0.0 (linux/amd64) kubernetes/$Format",
```
*Description of changes:*

The controller user agent is set in the kubeConfig / [client-go](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L491-L505) in this controller's case.

And it can be customized by updating `.UserAgent` of the kubeconfig used in making the call.

## Testing.

1. Built the VPC RC Image with this change.
2. Deployed it to Beta Clusters
3. Verified it Audit Logs

```
fields @timestamp, @message, @logStream

| filter @logStream like /audit/

| filter requestURI like /cninodes/
```

image

After this change the userAgent will set to

```
        "@message": {
            "kind": "Event",
            "apiVersion": "audit.k8s.io/v1",
            "level": "Metadata",
            "auditID": "2e4b4b35-1aa0-4733-89c4-4xxx",
            "stage": "ResponseStarted",
            "requestURI": "/apis/vpcresources.k8s.aws/v1alpha1/cninodes?allowWatchBookmarks=true&resourceVersion=36132&timeoutSeconds=421&watch=true",
            "verb": "watch",
            "user": {
                "username": "eks:vpc-resource-controller",
                "groups": [
                    "system:authenticated"
                ]
            },
            "sourceIPs": [
                "10.0.97.217"
            ],
            "userAgent": "vpc-resource-controller/v1.5.0-11-gaabe5fd",
```
- there is no unit-test for the _main_ module for this controller at the moment; this will require some refactor.
 

### DIFF
--- a/main.go
+++ b/main.go
@@ -204,6 +204,7 @@ func main() {
 	// Set the API Server QPS and Burst
 	kubeConfig.QPS = config.DefaultAPIServerQPS
 	kubeConfig.Burst = config.DefaultAPIServerBurst
+	kubeConfig.UserAgent = fmt.Sprintf("%s/%s", config.ControllerName, version.GitVersion)
 
 	setupLog.Info("starting the controller with leadership setting",
 		"leader mode enabled", enableLeaderElection,
--- a/main.go
+++ b/main.go
@@ -204,7 +204,7 @@ func main() {
 	// Set the API Server QPS and Burst
 	kubeConfig.QPS = config.DefaultAPIServerQPS
 	kubeConfig.Burst = config.DefaultAPIServerBurst
-	kubeConfig.UserAgent = fmt.Sprintf("%s/%s", config.ControllerName, version.GitVersion)
+	kubeConfig.UserAgent = fmt.Sprintf("%s/%s", ec2API.AppName, version.GitVersion)
 
 	setupLog.Info("starting the controller with leadership setting",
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -204,6 +204,7 @@ func main() {
 	// Set the API Server QPS and Burst
 	kubeConfig.QPS = config.DefaultAPIServerQPS
 	kubeConfig.Burst = config.DefaultAPIServerBurst
+	kubeConfig.UserAgent = fmt.Sprintf("%s/%s", config.ControllerName, version.GitVersion)
```
haouc, line 207: Can we use the same AppName at https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/aws/ec2/api/wrapper.go#L42? The name is used for EC2 dashboard as client user. We should use the same name if possible.

orsenthil: Updated this to use `ec2API.AppName`

haouc (approved):
lgtm

PR #466, opened 2024-09-11
Update aws-sdk-go and change way to get regional sts endpoint

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This updates the aws-sdk-go and the way to get regional sts endpoint

### DIFF
--- a/pkg/aws/ec2/api/wrapper.go
+++ b/pkg/aws/ec2/api/wrapper.go
@@ -510,19 +510,19 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion, roleARN, clusterN
 	}
 	e.log.Info("created rate limited http client", "qps", qps, "burst", burst)
 
-	// Get the regional sts end point
-	regionalSTSEndpoint, err := endpoints.DefaultResolver().
-		EndpointFor("sts", aws.StringValue(userStsSession.Config.Region), endpoints.STSRegionalEndpointOption)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get the regional sts endoint for region %s: %v",
-			*userStsSession.Config.Region, err)
-	}
-
+	// GetPartition ID, SourceAccount and SourceARN
 	roleARN = strings.Trim(roleARN, "\"")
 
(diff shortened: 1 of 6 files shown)


jayanthvn:
Need to fix -

```
go vet ./...
# github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils
# [github.com/aws/amazon-vpc-resource-controller-k8s/pkg/utils]
Error: vet: pkg/utils/helper_test.go:545:20: assignment mismatch: 3 variables but GetSourceAcctAndArn returns 4 values
make: *** [Makefile:30: verify] Error 1
```

pkg/aws/ec2/api/wrapper.go
```diff
@@ -892,3 +892,35 @@ func (e *ec2Wrapper) DisassociateTrunkInterface(input *ec2.DisassociateTrunkInte
 	}
 	return err
 }
+
+func (e *ec2Wrapper) getRegionalStsEndpoint(partitionID, region string) (endpoints.ResolvedEndpoint, error) {
```
jayanthvn, line 896: We don't have UTs for this?

jaydeokar: Added

pkg/aws/ec2/api/wrapper.go
```diff
@@ -892,3 +892,35 @@ func (e *ec2Wrapper) DisassociateTrunkInterface(input *ec2.DisassociateTrunkInte
...
+	var partition *endpoints.Partition
+	var stsServiceID = "sts"
+	for _, p := range endpoints.DefaultPartitions() {
+		if partitionID == p.ID() {
+			partition = &p
+			break
+		}
+	}
+	if partition == nil {
+		return endpoints.ResolvedEndpoint{}, fmt.Errorf("partition %s not valid", partitionID)
```
haouc, line 906: nit: ideally we should return `nil, err` when error is not nil though it is ok to leave it as is for now.

haouc (approved):
lgtm

pkg/aws/ec2/api/wrapper.go
jaydeokar: Ack, I'll update this as a follow up

PR #467, opened 2024-09-12
Remove hard failure for not getting global STS endpoint

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
We don't use global STS endpoint so in case global endpoint is not found, ignore and move ahead with regional STS provider 

### DIFF
--- a/pkg/aws/ec2/api/wrapper.go
+++ b/pkg/aws/ec2/api/wrapper.go
@@ -537,22 +537,23 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion, roleARN, clusterN
 	// TODO: we should revisit the global sts endpoint and check if we should remove global endpoint
 	// we are not using it since the concern on availability and performance
 	// https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
+
 	globalSTSEndpoint, err := endpoints.DefaultResolver().
 		EndpointFor("sts", aws.StringValue(userStsSession.Config.Region))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/aws/ec2/api/wrapper.go
```diff
@@ -537,22 +537,23 @@ func (e *ec2Wrapper) getClientUsingAssumedRole(instanceRegion, roleARN, clusterN
...
-		globalProvider := &stscreds.AssumeRoleProvider{
-			Client:   e.createSTSClient(userStsSession, client, regionalSTSEndpoint, sourceAcct, sourceArn),
-			RoleARN:  roleARN,
-			Duration: time.Minute * 60,
+		e.log.Info("failed to get the global STS Endpoint, ignoring", "roleARN", roleARN)
+	} else {
+		// If the regional STS endpoint is different than the global STS endpoint then add the global sts endpoint
+		if regionalSTSEndpoint.URL != globalSTSEndpoint.URL {
+			globalProvider := &stscreds.AssumeRoleProvider{
+				Client:   e.createSTSClient(userStsSession, client, regionalSTSEndpoint, sourceAcct, sourceArn),
```
jayanthvn, line 549: Don't we need to pass `globalSTSEndpoint` instead of `regionalSTSEndpoint` ?

jayanthvn (approved):
We will review the regional -> global EP post merge.. need more context

orsenthil:
>  We don't use global STS endpoint so in case global endpoint is not found, ignore and move ahead with regional STS provider

This needs to be tested for other regions.

I think, we rely on global endpoint if the regional endpoint is not available. But post merge is fine.

PR #485, opened 2024-10-21
Skipping health check on nodes if EC2 returns throttling errors

haouc:
*Issue #, if available:*

*Description of changes:*
The controller shouldn't health check on resources when the upstream API returns some specific errors, such as EC2 throttling, since those errors are not recoverable by failing the controller.

### DIFF
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -57,6 +57,7 @@ type manager struct {
 	worker            asyncWorker.Worker
 	conditions        condition.Conditions
 	controllerVersion string
+	stopHealthCheckAt time.Time
 }
 
 // Manager to perform operation on list of managed/un-managed node
@@ -66,6 +67,7 @@ type Manager interface {
 	UpdateNode(nodeName string) error
 	DeleteNode(nodeName string) error
 	CheckNodeForLeakedENIs(nodeName string)
+	SkipHealthCheck() bool
 }
 
 // AsyncOperation is operation on a node after the lock has been released.
@@ -95,6 +97,8 @@ type AsyncOperationJob struct {
 	node     node.Node
 	nodeName string
 }
+
+const pausingHealthCheckDuration = 10 * time.Minute
 
 // NewNodeManager returns a new node manager
 func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager,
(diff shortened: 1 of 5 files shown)


yash97 (approved):
lgtm 🚀

pkg/node/manager/manager.go
```diff
@@ -425,6 +429,10 @@ func (m *manager) performAsyncOperation(job interface{}) (ctrl.Result, error) {
 		utils.SendNodeEventWithNodeName(m.wrapper.K8sAPI, asyncJob.nodeName, utils.VersionNotice, fmt.Sprintf("The node is managed by VPC resource controller version %s", m.controllerVersion), v1.EventTypeNormal, m.Log)
 		err = asyncJob.node.InitResources(m.resourceManager)
 		if err != nil {
+			if pauseHealthCheckOnError(err) && !m.SkipHealthCheck() {
```
sushrk, line 432: Any reason why we only check for failures in `Init` op? 
`Update` and `Delete` operations also call EC2 APIs, IMO we can extend this check for all cases.

pkg/node/manager/manager.go
```diff
@@ -96,6 +98,8 @@ type AsyncOperationJob struct {
 	nodeName string
 }
 
+const pausingHealthCheckDuration = 30 * time.Minute
```
sushrk, line 101: Would it better to try a exponential back off strategy here rather than wait for a fixed amount of time? 

Most customers would recover from the failure pretty quickly I assume, we might not want to block health check for 30 mins.

haouc: I would try to avoid more complicate exponential back off for this simple and safe skipping health check. How about 10 minutes?

pkg/node/manager/manager.go
haouc: hmm, Add/Update both go through `Init` op and call into the same path in terms of node resource initialization. I would make this as simple as we can instead of extending them to all EC2 APIs at this moment. wdyt?

sushrk: Makes sense, ok to keep this. We can extend this on an as-needed basis for other APIs in the future.

yash97 (approved):
lgtm 🚀

PR #502, opened 2024-12-13
Centralized leaked ENI cleanup- CNINode CRD changes

sushrk:
*Description of changes:*
PR 1, adding support for Centralize leaked ENI cleanup in the controller to delete leaked ENIs provisioned by the controller and VPC-CNI:

* Add a new field in Spec to add Tag: This is used to specify tags that need to added to network interfaces provisioned by the controller and VPC-CNI. VPC-CNI will read the CNINode to get the tags.
* Added finalizer in CNINode: when kubernetes deletes an object that has finalizers specified for it, the Kubernetes API marks the object for deletion by populating .metadata.deletionTimestamp, and returns a 202 status code (HTTP "Accepted"). The target object remains in a terminating state while the control plane, or other components, take the actions defined by the finalizers. After these actions are complete, the controller removes the relevant finalizers from the target object. When the metadata.finalizers field is empty, Kubernetes considers the deletion complete and deletes the object.
* Added finalizer manager
* Fixed some go static check alerts.


pkg/k8s/utils.go
```diff
@@ -0,0 +1,23 @@
...
+// permissions and limitations under the License.
+
+package k8s
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NamespacedName returns the namespaced name for k8s objects
+func NamespacedName(obj client.Object) client.ObjectKey {
```
haouc, line 21: Looks like redundant func. The callers can directly call client?

pkg/k8s/finalizer.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+type defaultFinalizerManager struct {
+	k8sClient client.Client
+	log       logr.Logger
+}
+
+func (m *defaultFinalizerManager) AddFinalizers(ctx context.Context, obj client.Object, finalizers ...string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
haouc, line 43: Can we use reconciler retry/requeue mechanism here?

pkg/k8s/finalizer.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		if !needsUpdate {
+			return nil
+		}
+		return m.k8sClient.Patch(ctx, obj, client.MergeFromWithOptions(oldObj, client.MergeFromWithOptimisticLock{}))
+	})
+}
+
+func (m *defaultFinalizerManager) RemoveFinalizers(ctx context.Context, obj client.Object, finalizers ...string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
haouc, line 65: Same here, can we use reconciler retry/requeue mechanism here?

pkg/k8s/finalizer.go
```diff
@@ -0,0 +1,84 @@
...
+		if err := m.k8sClient.Get(ctx, NamespacedName(obj), obj); err != nil {
+			return err
+		}
+
+		oldObj := obj.DeepCopyObject().(client.Object)
+		needsUpdate := false
+		for _, finalizer := range finalizers {
```
haouc, line 72: I am just boldly thinking if we should remove all finalizers although we should have only one so far, in stead of checking every one. We get into here because we are cleaning up resources and removing the blocker of deletion. I don't see any reason we can tolerate finalizer(s) leftover after successfully making this call. WDYT?

sushrk: Do you mean we need not check against the `finalizers` passed to the function and instead remove all finalizers here?

I don't think we should do that within the generic finalizer manager. If we want to remove all finalizers in the CNINode reconciler, then we should pass in all finalizers to the function.

yash97 (approved):
lgtm

pkg/k8s/finalizer.go
sushrk: Resolving the thread to merge the PR, I'll follow up on this offline.

PR #506, opened 2024-12-28
Upgrade to AWS SDK Go v2

### LINKED ISSUE #471: The controller need upgrade to AWS SDK Go v2
**What would you like to be enhanced**:
From July 31, 2024, the v1 SDK will not receive API updates for new or existing services, or be updated to support new regions.
https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

**Why is the change needed and what use case will it solve**:

### LINKED ISSUE #472: Investigate if we need to have global STS endpoint
**What would you like to be enhanced**:
Investigate if we need to have a global fallback sts endpoint as per this comment 
https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/dd3564c30c5da8233d3ab5737ce331ee2fe0d10d/pkg/aws/ec2/api/wrapper.go#L547-L549

Also this piece of code where if regional STS !=global STS we still end up using the regional STS endpoint. It should be global STS if regional is not available. 

https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/dd3564c30c5da8233d3ab5737ce331ee2fe0d10d/pkg/aws/ec2/api/wrapper.go#L547-L549

**Why is the change needed and what use case will it solve**:

phuhung273:
**Issue #, if available:**
Close #471
Close #472

**Description of changes:**
- Upgrade to AWS SDK Go v2

**Component that still using v1**
"pkg/aws/ec2/api/wrapper.go".ec2Wrapper.getRegionalStsEndpoint is the only function still use V1. Leave it as TODO. The only way I can think of is to clone v1 `endpoints` code into this repo. Please help give opinion on this.


pkg/utils/int.go
```diff
@@ -0,0 +1,42 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import "errors"
+
+func Int64ToInt32(value int64) (int32, error) {
```
yash97, line 18: why do we need this functions? Also i see int64 was changed to int32. Please provide reason this change.

phuhung273: In v2, field `Ipv4PrefixCount` is changed to int32 for some structs, eg:
- https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#CreateNetworkInterfaceInput
- https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#AssignPrivateIpAddressesInput

So decided to also change the interface

phuhung273: @yash97 can you help me check on this

haouc:
Close this PR due to a merged #523

PR #544, opened 2025-04-11
feat: support control plane flag to turn off the controller

haouc:
*Issue #, if available:*

*Description of changes:*
We want to support disable the controller if none of SGP and Windows are required. The disable flag should be passed from EKS managed services. After the controller is disabled, the lightly loaded controller will only run 
- Network Interface cleanup
- Webhooks uninstaller
- Possible CNINode leaked resources cleanup

This change will add a new metric for enabled/disabled controller.

The change has been tested with integrations and cleanup tests. 


yash97 (approved):
lgtm

PR #546, opened 2025-04-13
supported instances update

haouc:
*Issue #, if available:*

*Description of changes:*
g3/p2 has been removed by ec2 ENI trunking support and i7ie family was updated.


haouc:
g3/p2's removal is a concern for me since this may break some prod support if those types are in use. We need to sync up with ec2 on the timeline.

yash97 (approved):

PR #660, opened 2026-04-24
Getting connection tracking settings from primary ENI and reusing it for branch ENIs

jaydeokar:
*Issue #, if available:*
Feature. Adds a way to configre connection tracking settings on branch ENI ENI

*Description of changes:*
This reuses the connection timeout settings from primary ENI and uses it while creating Branch ENI on that instance. 
The setting is only required on branch ENIs and not trunk ENI
https://github.com/jaydeokar/connection-tracking/blob/main/CONNTRACK_TEST_CASES.md

As a part of this 
- Also updated image to use al23


pkg/aws/ec2/instance.go
```diff
@@ -161,6 +168,15 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
 				i.primaryENISecurityGroups = append(i.primaryENISecurityGroups, *group.GroupId)
 			}
 		}
+
+		// Get the connection tracking configuration from the primary ENI
+		if *index == 0 {
```
haouc: nit: index has been dereferenced at line 161 but still worth to consider if a nil check is necessary. It is unlikely nil since it is returned from EC2 and associated to an associated ENI. Just want to call out.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -239,7 +261,7 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
 		}
 
 		trunk, err := t.ec2ApiHelper.CreateAndAttachNetworkInterface(&instanceID, aws.String(t.instance.SubnetID()),
-			t.instance.CurrentInstanceSecurityGroups(), t.nodeIDTag, &freeIndex, &TrunkEniDescription, &InterfaceTypeTrunk, nil)
+			t.instance.CurrentInstanceSecurityGroups(), t.nodeIDTag, &freeIndex, &TrunkEniDescription, &InterfaceTypeTrunk, nil, t.getConnectionTrackingSpec())
```
haouc: Should we confirm with EC2 what would be expected for trunk connnection timeout set up? Branch ENI makes sense but I am not sure if trunk also need to be set?

jaydeokar: I'll confirm and update the PR

pkg/aws/ec2/instance.go
jaydeokar: Using aws.ToInt32 now

pkg/provider/branch/trunk/trunk.go
jaydeokar: We don't need to addd it to trunk ENI.

main.go
```diff
@@ -303,6 +303,7 @@ func main() {
 		instanceClientBurst, userClientQPS, userClientBurst, setupLog)
 	if err != nil {
 		setupLog.Error(err, "unable to create ec2 wrapper")
+		os.Exit(1)
```
jaydeokar, line 306: Exiting if we aren't able to get the ec2 wrapper since none of the commands work after that.

yash97 (approved):

PR #679, opened 2026-07-17
Near-zero-EC2 controller re-init via CNINode checkpoint, plus branch-ENI/VLAN correctness under restart and high churn

Issacwww:
## Summary

Beyond adding the CNINode status checkpoint, this PR makes the controller's branch-ENI / VLAN
management correct and cheap under the two conditions that matter at scale: **controller restart**
and **high pod churn**. Everything here is a change to the existing control-plane controller — there
is no new component.

## What this changes

- **Near-zero-EC2 restart.** On restart the controller rebuilds its branch-ENI / VLAN state by reading
  each node's CNINode `status` (a persisted checkpoint) instead of calling EC2
  `DescribeNetworkInterfaces` once per node. At large node counts this turns an O(nodes) burst of EC2
  calls on every restart into cheap Kubernetes reads. If a checkpoint is missing or unusable it falls
  back to EC2 discovery for that node.

- **Immediate capacity release on pod delete, with a VLAN reuse cooldown.** Deleting a branch-ENI pod
  frees its trunk capacity slot in the same reconcile pass, so a slow or failed teardown can't hold a
  slot hostage. The freed VLAN id, however, is withheld from reuse until a short cooldown elapses —
  the data plane needs time to forget the old VLAN/routing state before that id is safe to reuse.

- **Error-driven orphan reclaim.** When adding a branch ENI fails because EC2's real capacity
  disagrees with the controller's in-memory bookkeeping (an orphan branch ENI is occupying a
  slot/VLAN the bookkeeping doesn't know about), the controller does a single describe, reclaims the
  orphan through the normal delete path, and lets the pod's next reconcile succeed — instead of
  looping create/fail indefinitely.

- **Concurrency guards for the periodic orphan sweep.** Under high churn the low-frequency
  "reconcile unassigned branch ENIs" sweep can race with in-flight allocation/deletion. Three guards
  prevent it from doing harm: it skips a branch ENI whose creation is still in flight (not yet
  recorded), it de-duplicates an ENI that is already queued for deletion, and it refuses to free a
  VLAN id that has since been reassigned to a different ENI.

- **Recycled-instance detection on CNINode.** A CNINode is keyed by node name, so a recycled node
  name can leave a checkpoint that actually belongs to a different EC2 instance. The controller now
  compares the recorded instance id against the live one (on both the normal reconcile and the
  restart-hydrate path); on a mismatch it records it and safely falls back to EC2 rather than trusting
  stale data.

- **Observability.** New counters and structured VLAN-lifecycle log lines for all of the above, so
  these behaviors are visible in metrics and logs. (Exporting these counters to CloudWatch is a
  separate, companion change.)

- **Tests.** An integration test for the immediate-slot-release + VLAN-cooldown behavior, and explicit
  log messages on the VLAN-free guards.

## Validation (beta, 1500-node cluster unless noted)

| What was validated | Scale / method | Result |
|---|---|---|
| Restart rebuilds state with ~0 per-node EC2 calls | 1500 nodes; controller restart | **PASS** — every node hydrated from its CNINode checkpoint; EC2 describes ≈ 0 on the restart path |
| Slot freed immediately on delete; freed VLAN not reused during cooldown | node at full branch-ENI capacity; delete + replacement pod | **PASS** — replacement reached Running while the old VLAN was still cooling, and got a different VLAN/ENI |
| Recycled-instance / stale-checkpoint detection + self-heal | 3 nodes with injected mismatched instance ids, then a restart | **PASS** — detected on both the reconcile and hydrate paths; nodes stayed healthy; checkpoints self-corrected |
| Concurrency guards under churn | unit tests + high churn on 4 full-capacity nodes | **Logic PASS (unit tests).** In live churn the owner-aware VLAN-free guard was observed firing 18×; the in-flight-skip and delete-queue-dedup guards protect millisecond-scale races and did not reproduce live in the window (expected — the unit tests cover them) |
| Error-driven orphan reclaim (capacity blind-spot recovery) | observed under high churn | **PASS** — reclaimed real orphans produced by the churn; no create/fail wedge |
| Controller stable under extreme churn | 4 nodes at full capacity, sustained delete/recreate bursts | **PASS** — no error-level logs and no crash; elevated retry/error counters were expected EC2 capacity-contention, handled gracefully |

## Still being filled in

Tracked in a follow-up test plan; **not blocking review of the code in this PR** — these validate
operational robustness of the same changes, not additional code.

| Area | Why it matters | Status |
|---|---|---|
| No-regression + security-group correctness after churn / restart / reclaim | confirm the lifecycle changes never leave a pod on a stale or wrong branch ENI (i.e. wrong security group) | In progress |
| Multi-hour soak at scale | confirm no branch-ENI / VLAN / memory leak accumulates over time | In progress |
| VLAN-pool exhaustion edge | confirm that when nearly all VLAN ids on a trunk are in use or cooling, assignment never reuses a cooling or foreign VLAN | In progress |


pkg/node/node.go
```diff
@@ -152,16 +205,29 @@ func (n *node) UpdateResources(resourceManager resource.ResourceManager) error {
...
-			Message: "failed to load instance details",
-			Err:     err,
+
+	// Fast path: try to hydrate the instance from the persisted CNINode status
+	// snapshot, which skips the expensive EC2 LoadDetails on re-init. Only fall
+	// back to LoadDetails (cold-start / first-init path) when hydration fails.
+	if n.tryHydrateInstanceFromCNINodeStatus() {
+		// Hydrated from CNINode status: LoadDetails (EC2 DescribeInstances) is skipped.
+		nodeReinitEC2Skipped.WithLabelValues(reinitResultHydrated).Inc()
+	} else {
```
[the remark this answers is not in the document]
Issacwww: Fixed in e9133b8: the hydrated fast-path now records a 0s `nodeInitStageLatency{stage=load_details}` observation, so the summary carries a real datapoint proving LoadDetails was skipped instead of having zero samples. Added a unit test for the 0-observation.

pkg/k8s/wrapper.go
```diff
@@ -284,3 +310,36 @@ func (k *k8sWrapper) ListCNINodes() ([]*rcv1alpha1.CNINode, error) {
...
+
+	return retry.OnError(cniNodeStatusUpdateBackoff, retriable, func() error {
+		cniNode := &rcv1alpha1.CNINode{}
+		if err := reader.Get(k.context, request, cniNode); err != nil {
+			return err
+		}
+
+		newCNINode := cniNode.DeepCopy()
+		newCNINode.Status = status
+		return k.cacheClient.Status().Patch(k.context, newCNINode, client.MergeFrom(cniNode))
```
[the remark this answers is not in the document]
Issacwww: Fixed in e9133b8: `UpdateCNINodeStatus` now patches status with `MergeFromWithOptimisticLock` so conflicting status writes reliably surface as Conflict and are retried. The retry loop re-reads the object fresh via the non-cached apiReader each attempt and the backoff is bounded (4 steps, ~3s), so there is no hot loop.

pkg/provider/branch/provider.go
```diff
@@ -298,6 +428,81 @@ func (b *branchENIProvider) ReconcileNode(nodeName string) bool {
...
+	pi, di := persisted.Instance, desired.Instance
+	if pi.InstanceID != di.InstanceID ||
+		pi.InstanceType != di.InstanceType ||
+		pi.InstanceSubnetID != di.InstanceSubnetID ||
+		pi.CurrentSubnetID != di.CurrentSubnetID ||
+		pi.PrimaryNetworkInterfaceID != di.PrimaryNetworkInterfaceID {
+		return false
+	}
+	return true
+}
```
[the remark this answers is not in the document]
Issacwww: Fixed in e9133b8: `cniNodeStatusUpToDate` now compares exactly the fields `HydrateFromCNINodeStatus` validates - both security-group sets (order-independent via sameStringSet), subnet v4/v6 CIDR + masks, all identity fields, and connection-tracking - so a stale/partial snapshot is treated as out-of-date and re-patched. Added unit tests covering the full field set, set-order independence, and conn-tracking nil-equivalence.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -622,6 +701,34 @@ func (t *trunkENI) pushENIToDeleteQueue(eni *ENIDetails) {
...
+			continue
+		}
+
+		// Even though the ENI is going to be deleted, keep the VLAN reserved while it sits in the cool down queue.
+		t.markVlanAssigned(vlanId)
+		t.pushENIToDeleteQueue(&ENIDetails{
+			ID:                branchENIID,
+			VlanID:            vlanId,
+			deletionTimeStamp: time.Now(),
+		})
```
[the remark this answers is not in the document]
Issacwww: Fixed in e9133b8: `pushUnassignedBranchInterfacesToDeleteQueue` now validates the tag-parsed VLAN ID and falls back to the reserved id 0 when out of range, so `deleteENI`->`freeVlanId` can no longer panic on an out-of-bounds index. Added a unit test for the invalid-VLAN path.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -622,6 +733,50 @@ func (t *trunkENI) pushENIToDeleteQueue(eni *ENIDetails) {
...
+		// An attached branch ENI owned by no pod in the ledger is an orphan. Count each discovery so
+		// the real orphan rate is observable in Grafana (previously only the log line above existed).
+		branchENIOrphanReclaimedCount.WithLabelValues("discovered").Inc()
+
+		vlanId, err := t.getVlanIdFromTag(branchInterface.TagSet)
+		if err != nil {
+			trunkENIOperationsErrCount.WithLabelValues("get_vlan_from_tag").Inc()
+			t.log.Error(err, "failed to find vlan id", "interface", branchENIID)
+			continue
+		}
```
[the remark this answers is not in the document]
Issacwww: Fixed in fcf1a6d: on `getVlanIdFromTag` error we now fall back to the reserved VLAN ID 0 and still enqueue the discovered orphan (same handling as the out-of-range case), instead of `continue`-ing. This prevents leaking a real orphan in EC2 and keeps the orphan-discovered metric consistent with 'pushed to delete queue'. Added `TestTrunkENI_pushUnassignedBranchInterfacesToDeleteQueue_MissingVlanTag` covering the missing/unparseable-tag path (enqueued with VLAN 0, deleteENI does not panic).

pkg/node/node.go
```diff
@@ -197,6 +275,29 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+func (n *node) tryHydrateInstanceFromCNINodeStatus() bool {
+	if n.k8sAPI == nil {
+		return false
+	}
+
+	cniNode, err := n.k8sAPI.GetCNINode(types.NamespacedName{Name: n.instance.Name()})
+	if err != nil {
+		n.log.V(1).Info("could not get CNINode status snapshot, falling back to EC2 instance details",
+			"error", err)
+		return false
```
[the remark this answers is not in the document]
Issacwww: Fixed in fcf1a6d: added `GetCNINodeFromAPIServer` (non-cached apiReader read) and the hydrate fast path in `tryHydrateInstanceFromCNINodeStatus` now uses it, so a lagging informer cache on restart/leader change no longer spuriously misses and forces the EC2 LoadDetails fallback. Mirrors how UpdateCNINodeStatus already uses the apiReader.

pkg/provider/branch/provider.go
```diff
@@ -203,6 +245,93 @@ func (b *branchENIProvider) InitResource(instance ec2.EC2Instance) error {
...
+func (b *branchENIProvider) initTrunk(
+	instance ec2.EC2Instance,
+	trunkENI trunk.TrunkENI,
+	podList []v1.Pod,
+	log logr.Logger,
+) (bool, error) {
+	if instance.LoadedFromCNINodeStatus() {
+		statusStart := time.Now()
+		cniNode, err := b.apiWrapper.K8sAPI.GetCNINode(types.NamespacedName{Name: instance.Name()})
+		if err == nil {
```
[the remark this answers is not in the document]
Issacwww: Fixed in fcf1a6d: `initTrunk`'s CNINode-status fast path now reads via the new non-cached `GetCNINodeFromAPIServer`, same reasoning as the instance hydrate path - avoids a stale-cache false miss defeating the zero-EC2 re-init goal.

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +361,129 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+	}
+	if instanceStatus.InstanceType == "" {
+		return false, "missing_instance_type"
+	}
+	if instanceStatus.InstanceSubnetID == "" || instanceStatus.InstanceSubnetCIDRBlock == "" {
+		return false, "missing_instance_subnet"
+	}
+	if instanceStatus.CurrentSubnetID == "" || instanceStatus.CurrentSubnetCIDRBlock == "" {
+		return false, "missing_current_subnet"
+	}
```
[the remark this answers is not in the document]
Issacwww: Fixed in 15d5b58: HydrateFromCNINodeStatus now returns a cache-miss (`missing_subnet_mask`) when SubnetMask is empty, and `missing_subnet_v6_mask` when a v6 CIDR is present without its mask - so an incomplete/pruned snapshot falls back to EC2 instead of hydrating an instance that would build malformed pod CIDRs. Added TestEc2Instance_HydrateFromCNINodeStatus_MissingSubnetMask covering the valid hit plus both miss reasons.

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +361,140 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
 
 	return i.tcpEstablishedTimeout, i.udpStreamTimeout, i.udpTimeout
 }
+
+func (i *ec2Instance) HydrateFromCNINodeStatus(status rcv1alpha1.CNINodeStatus) (bool, string) {
```
haouc: Looks like this func is returning a boolean and a type. Should we use a ENUM instead of a string?

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +361,140 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+	i.subnetMask = instanceStatus.SubnetMask
+	i.subnetV6Mask = instanceStatus.SubnetV6Mask
+	i.primaryENIID = instanceStatus.PrimaryNetworkInterfaceID
+	i.primaryENISecurityGroups = slices.Clone(instanceStatus.PrimaryNetworkInterfaceSecurityGroups)
+
+	i.tcpEstablishedTimeout = nil
+	i.udpStreamTimeout = nil
+	i.udpTimeout = nil
+	if instanceStatus.ConnectionTracking != nil {
+		i.tcpEstablishedTimeout = cloneInt32Ptr(instanceStatus.ConnectionTracking.TCPEstablishedTimeout)
```
haouc, line 473: Any reason we want to use pointers on those timeouts?

pkg/k8s/wrapper.go
```diff
@@ -284,3 +331,40 @@ func (k *k8sWrapper) ListCNINodes() ([]*rcv1alpha1.CNINode, error) {
...
 }
+
+func (k *k8sWrapper) UpdateCNINodeStatus(nodeName string, status rcv1alpha1.CNINodeStatus) error {
+	request := types.NamespacedName{Name: nodeName}
+
+	// reader is the non-cached API server reader so we read back a CNINode that
+	// may have been created moments earlier (and is not yet in the informer
+	// cache) before patching its status. Fall back to the cache client only if
+	// no apiReader was wired.
+	reader := client.Reader(k.cacheClient)
```
haouc: Shouldn't this be check k.apiReader first before creating a new reader?

pkg/node/manager/manager.go
```diff
@@ -101,19 +106,32 @@ type AsyncOperationJob struct {
...
+// DefaultBranchENIOrphanSweepInterval is the fallback base cadence for the EC2 orphan branch-ENI
+// reclaim sweep when a non-positive interval is supplied. It matches the flag default in main.go.
+const DefaultBranchENIOrphanSweepInterval = time.Hour
+
 // NewNodeManager returns a new node manager
 func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager,
-	wrapper api.Wrapper, worker asyncWorker.Worker, conditions condition.Conditions, clusterName string, controllerVersion string, healthzHandler *rcHealthz.HealthzHandler) (Manager, error) {
+	wrapper api.Wrapper, worker asyncWorker.Worker, conditions condition.Conditions, clusterName string, controllerVersion string, orphanSweepInterval time.Duration, healthzHandler *rcHealthz.HealthzHandler) (Manager, error) {
+
+	// Guard against a mis-configured (zero or negative) interval so a slow sweep is always scheduled;
```
haouc: Any reason we want to make intervals configurable?

pkg/node/manager/manager.go
```diff
@@ -124,35 +142,100 @@ func NewNodeManager(logger logr.Logger, resourceManager resource.ResourceManager
...
-					}
-					cachedNode.SetReconciliationInterval(interval)
-				}
-				cachedNode.SetNextReconciliationTime(time.Now().Add(cachedNode.GetReconciliationInterval()))
-				m.Log.Info("reconciled node to cleanup leaked branch ENIs", "NodeName", nodeName, "NextInterval", cachedNode.GetReconciliationInterval(), "NextReconciliationTime", cachedNode.GetNextReconciliationTime())
+	if !reconcileDue && !sweepDue {
+		return
+	}
+
+	go func() {
```
haouc, line 226: Looks like this spawns a goroutine every time and loses control on them. I am not sure if this is a good way to trigger standalone goroutines to handle those tasks. Especially they keep reference to resources with nondeterministic lifetime.

pkg/k8s/wrapper.go
```diff
@@ -237,6 +268,22 @@ func (k *k8sWrapper) GetCNINode(namespacedName types.NamespacedName) (*rcv1alpha
 	return cninode, nil
 }
 
+// GetCNINodeFromAPIServer reads a CNINode directly from the API server via the non-cached
+// apiReader (falling back to the cache client only if no apiReader was wired). Used by the
+// re-init hydrate fast path so a lagging informer cache on restart / leader change does not
+// spuriously miss and force the EC2 LoadDetails fallback.
+func (k *k8sWrapper) GetCNINodeFromAPIServer(namespacedName types.NamespacedName) (*rcv1alpha1.CNINode, error) {
+	reader := client.Reader(k.cacheClient)
```
haouc: Looks a duplicated code and same order issue: `if k.apiReader != nil reader = k.apiReader else reader = cllient.Reader` ?

pkg/node/node.go
```diff
@@ -50,6 +102,12 @@ type node struct {
 	nextReconciliationTime time.Time
```
haouc, line 107: Is this still in use after this PR?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -214,11 +251,100 @@ func PrometheusRegister() {
...
+		t.uidToBranchENIMap[string(pod.UID)] = eniListFromPod
+	}
+
+	t.log.V(1).Info("successfully initialized trunk cache from CNINode status",
+		"trunk", t.trunkENIId, "branch interfaces", t.uidToBranchENIMap)
+	return nil
+}
+
+func (t *trunkENI) ReconcileUnassignedBranchENIs() (bool, error) {
```
haouc, line 569: >	// nextEC2SweepTime is the earliest time the EC2 orphan branch-ENI reclaim
	// (ReconcileUnassignedBranchENIs) may run for this node. It is an independent, low-frequency
	// timer, separate from nextReconciliationTime: the zero-EC2 Reconcile runs on the fast

I do have concerns on this approach. Since we are no longer list all branch ENIs on node's startup, the controller doesn't know branch ENIs which are not assigned to pods. This can cause a race condition: 
T0: the controller restarts / leader election change
T1: a pod is deleted and network is teared down with branch ENI A
T2: the controller hydrate its cache from CNINodes' trunk info and branch ENIs from pods' annotation
T3: a new pod is created
T4: the controller creates a branch ENI and assigns vlan ID to the ENI. Unfortunately the vlan ID happens to be same as the branch ENI A (orphaned) because the controller doesn't know it exists and doesn't care if they are existing since some process will clean them up on some time.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -214,11 +251,100 @@ func PrometheusRegister() {
...
+	defer t.lock.RUnlock()
+
+	return rcv1alpha1.TrunkENIStatus{
+		ID:             t.trunkENIId,
+		SubnetID:       t.instance.SubnetID(),
+		SecurityGroups: slices.Clone(t.instance.CurrentInstanceSecurityGroups()),
+	}
+}
+
+func (t *trunkENI) InstanceStatus() rcv1alpha1.InstanceStatus {
```
haouc: Do we need a func for this online return?

pkg/worker/jobs.go
```diff
@@ -87,6 +89,14 @@ func NewOnDemandProcessDeleteQueueJob(nodeName string) OnDemandJob {
 	}
 }
 
+// NewOnDemandReconcileUnassignedBranchENIsJob returns a job that discovers EC2 branch ENIs missing pod ownership.
+func NewOnDemandReconcileUnassignedBranchENIsJob(nodeName string) OnDemandJob {
+	return OnDemandJob{
+		Operation: OperationReconcileUnassignedBranchENIs,
```
haouc, line 97: I think this should be triggered on the controller starting runnables. But I can't say for sure we won't have the race condition on pods' deletion and branch ENI's vlan being reused.

pkg/provider/branch/trunk/trunk.go
Issacwww: Yeah you're right, this race is real. Looking at it more, the old describe in `InitTrunk` was actually doing two jobs - finding orphans to clean up, but also marking their vlans as used before we ever hand out a new one. I moved the cleanup part to the sweep but dropped the vlan part with it, and that part is correctness, not hygiene.

So rather than trying to win the timing game with the sweep, I'll gate the allocation path directly: each trunk gets an in-memory `branchLedgerVerified` flag - true if it was initialized through the EC2 path (already described), false if it was hydrated from CNINode status. The first time `CreateAndAssociateBranchENIs` runs on an unverified trunk, it does one describe before allocating. Your timeline with that in place:

```
T0: restart / leader change
T1: pod deleted, ENI A (vlan 5) was sitting in the in-memory delete queue
    -> queue lost, A stays attached in EC2 as an orphan
T2: hydrate from CNINode status + pod annotations
    -> ledger doesn't know A exists; trunk marked verified=false
T3: new pod lands on the node
T4: allocation sees verified=false -> one DescribeNetworkInterfaces first:
    vlan 5 (and every other existing branch ENI's vlan) marked used,
    A pushed to the delete queue, verified=true
    (vlan 5 stays marked used until the delete actually succeeds in EC2,
     so it can't be handed out in the meantime)
T5: allocation picks a vlan -> can't collide with A
T6: everything after this on the node is pure in-memory, no more describes
```

A few properties that fall out of this:
- No vlan is ever handed out before the ledger has been checked against EC2 once after a restart, so the race is closed no matter what the sweep timing is.
- Existing pods never go through allocation (their ENIs come back from annotations at hydrate), so they're unaffected either way.
- Re-init is still zero EC2: the describe only happens when a node actually gets a new pod - one call per node, paced by pod arrivals instead of bursting at startup. A node that never gets a pod just stays unverified, which is fine; its orphans are the slow sweep's job.
- The flag is only ever written by the allocation path (the sweep doesn't touch it), so there's a single writer and it flips false->true once per process lifetime. On the next restart it naturally resets and the whole thing replays.

The sweep stays, but purely as the cleanup layer - quiet nodes that never allocate, plus ENIs we gave up deleting at runtime. Will push this with tests.

pkg/worker/jobs.go
Issacwww: Should be covered by the allocation-path gate above (https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/679#discussion_r3635057193) - agree startup runnables would shrink the window but not close it. With the gate, sweep timing doesn't matter for correctness anymore, it's purely cleanup.

pkg/aws/ec2/instance.go
Issacwww: Good point, will switch to typed string constants in the next revision - they double as metric label values, so typing them also prevents label typos.

pkg/aws/ec2/instance.go
Issacwww: These mirror the EC2 SDK - `ConnectionTrackingConfiguration` fields are `*int32` there too (https://github.com/aws/aws-sdk-go-v2/blob/main/service/ec2/types/types.go, search `ConnectionTrackingConfiguration`). nil means "not set" and 0 isn't even a valid value (min is 60s), so a plain int32 can't represent unset. The instance struct fields these round-trip into were already `*int32` before this PR.

pkg/k8s/wrapper.go
Issacwww: Agreed, will flip it to check `k.apiReader` first and extract a shared helper (same fix as the other thread below).

pkg/k8s/wrapper.go
Issacwww: Yeah, same 4 lines twice - will extract a small `freshReader()` helper and use it in both places.

pkg/node/manager/manager.go
Issacwww: Honestly this was mostly for my own testing - will drop the flag and hardcode it. Planning 1h (jittered to 1-2h per node): orphans only come from rare failure paths so volume is low, correctness doesn't depend on this timer anymore (see the allocation-gate thread), and it's one DescribeNetworkInterfaces per node per interval. Does 1h sound right to you?

pkg/node/manager/manager.go
Issacwww: Fair point. The bare goroutine is the pre-existing pattern here (it ran ReconcileNode the same way before this PR), so I'd rather not refactor that part in this PR. But for what this PR adds: the EC2 sweep already just submits a job to the existing worker pool (the actual describe/delete runs there, with lifecycle and retries), and I'll move the status self-heal onto the same pool as another job type in the next revision. Then everything new runs in the managed pool, and the goroutine stays what it was - timer bookkeeping only.

pkg/node/node.go
Issacwww: Yes, still in use - the two timers check opposite directions of drift at very different costs. Fast one (1-15min, zero EC2): catches "ledger has it, pod is gone" - missed pod deletion events, common at scale, needs to run often to free up IPs. Slow one (1h, one describe per node): catches "EC2 has it, ledger doesn't" - orphans from rare failure paths, low volume, expensive to check. Merging them means either describing every 15min per node (the burst this PR removes) or leaking pod IPs for an hour. They don't race either: both fields are behind the node lock, and both paths converge on the trunk's own lock - worst case is a duplicate idempotent pass. Will make the field comments spell this out.

pkg/provider/branch/trunk/trunk.go
Issacwww: You're right, it's not pulling its weight - single caller, and the persist path already grabs the instance snapshot directly instead of going through the trunk. Will drop this method and make the self-heal path consistent with persist.

Issacwww:
@haouc this is ready for another look. Since your last review:

1. **All review feedback addressed** (typed hydrate reasons, `freshReader` helper, self-heal moved into the worker pool, timer docs, dropped the one-line forwarder, sweep interval hardcoded - flag removed).
2. **The VLAN-reuse race you found is closed** with the allocation-path gate we discussed ([thread](https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/679#discussion_r3635057193)) - and proven on a live 1500-node cluster with an injected orphan: gate fired before first allocation, marked the orphan's vlan, new pod allocated a different vlan, zero AssociateTrunkInterface conflicts, orphan reclaimed.
3. **Rebased onto merged #681**: cherry-picked, and the status schema is now unified - shared `trunkInterface` (v1 writes `id`/`subnetID`/`securityGroups`, never `branches`) + v1-private `reinitCheckpoint`. Old-format snapshots degrade to a clean EC2 fallback (tested live: the CRD apply prunes them, one-time full-fleet fallback, then every restart is zero-EC2).
4. **E2E on 1500 nodes**: restart with checkpoints populated = `hydrated`=1500, zero EC2 describes, 2m30s leader-to-all-trunk-ready on default config (remaining wall clock is apiserver-qps queueing, orthogonal to EC2 - tuned config predicts ~37s). Sweeper, CEL rules, bad-snapshot fail-safe all exercised live - details in the updated PR description.

The PR description now includes the deployment notes (CRD+controller as one upgrade unit; drop the old sweep-interval flag from manifests).

Issacwww:
Rebased onto `release-1.7.16-onwards` now that #682 (the #681 backport) is merged. The branch is now a clean 12-commit delta with no schema commits duplicated; the tree is unchanged from the E2E-validated build (verified byte-identical).

PR #680, opened 2026-07-22
docs: correct stale 5-security-group limit note in README

### LINKED ISSUE #664: README error about SGP.
The README has a note in "Security Group for Pods" chapter that is now false.

This note speak about the limit of 5 Security Groups per `SecurityGroupPolicy` and how to bypass it by creating several `SecurityGroupPolicy` targeting the same pods to have the controller to merge them and apply more than 5 Security Groups if the AWS account quota permit this.

However, this limit of 5 was set on the CRD and as been removed since this merged PR: https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/583

I can confirm that this limit does not exists anymore, as I have created a `SecurityGroupPolicy` with 6 Security Groups with success.

As such, all the note should be removed. Perhaps just remember the readers that the real limit comes from the service quota (which is 5 by default and hard limited to 16).

jupdec:
The CRD-enforced limit of 5 security groups per SecurityGroupPolicy was removed in #583. Update the README note to reflect that the effective limit now comes from the security-groups-per-ENI service quota (default 5, hard limit 16) rather than the CRD.

Fixes #664

*Issue #, if available:*

*Description of changes:*


haouc (approved):
lgtm

PR #681, opened 2026-07-23
Add trunk/branch ENI status and managedBy ownership field to CNINodeCninode trunk status

haouc:
### What type of PR is this?
feature

### Which issue does this PR fix?
N/A — API proposal, opened for discussion.

Relates to #679 — that PR fills CNINodeStatus with a re-init checkpoint (instance + trunkENI id/subnet/SGs); this PR adds branch persistence and ownership semantics. The trunk fields should unify — happy to fold trunkInterface into an extended TrunkENIStatus (branches as an optional SSA map-list) so both writers share one schema." That signals flexibility up front and frames the merge path without blocking either PR.

### What does this PR do / Why do we need it?

Proposes a structured `status` for the CNINode CRD plus an ownership marker, so the controller managing a node's trunk ENI can persist trunk/branch interface state on the object:

1. **`status.trunkInterface`** — trunk ENI (id, MAC, device index) and its associated branch ENIs (id, vlanID, associationID, MAC, subnet CIDRs, security groups, assigned IPv4/IPv6 CIDRs). The `branches` list is an SSA map-type list keyed by `id` (`x-kubernetes-list-map-keys`), so distinct field managers can own individual entries without conflicts. The CNINode status today is an empty object with a TODO; this fills it in for the security-groups-for-pods use case (trunk ENI persistence for visibility and recovery across controller restarts).

2. **Status subresource enabled** — status writers can be granted RBAC on `cninodes/status` only, without spec/metadata/finalizer access.

3. **`spec.managedBy`** — declares which controller owns a CNINode and its status. It is also declared in the CRD's `selectableFields` (CustomResourceFieldSelectors, GA since 1.32), so a controller can filter its informer cache server-side to only the objects it manages. Empty means `vpc-resource-controller` (backward compatible: all existing objects have no value and behave exactly as before). The node controller already skips nodes with `eks.amazonaws.com/compute-type: auto` and never creates CNINodes for them; EKS Auto Mode networking controllers will create and reconcile their own CNINodes marked `managedBy: eks-auto-mode`. The CNINode reconciler now skips objects managed by another controller — importantly it does **not** add `networking.k8s.aws/resource-cleanup` to them, since its finalizer routine never runs for those objects and would deadlock deletion.

This gives v1 (this controller) and EKS Auto Mode a single shared schema for trunk ENI persistence, with disjoint object ownership by node type rather than shared-object writes.

### Notes

- Purely additive: no existing field changed or removed; empty `managedBy` preserves current behavior end to end.
- Nothing writes `status.trunkInterface` in this PR; the persistence write path is a follow-up. This PR is the API shape + ownership semantics.
- `make generate/manifests` output (controller-gen v0.18.0) included.

### Why `apis/vpcresources/crds/` duplicates `config/crd/bases/`

`go:embed` cannot reference files outside the embedding package's directory (`../../config/...` is a compile error, and symlinks are ignored), while `config/crd/bases` must stay where it is for the kustomize deploy chain. Neither copy is hand-maintained source — both are controller-gen output of `cninode_types.go`. Two mechanisms keep them in sync: `make verify` regenerates and copies in one step (its existing `git diff --quiet` gate fails CI on drift), and `TestEmbeddedCRDsMatchGenerated` byte-compares the embedded copies against `config/crd/bases` on every `go test`.

### Testing done on this change

- `go build ./...`, `go vet`, `go test ./controllers/crds/...` — pass.
- New unit tests: reconciler skips `managedBy: eks-auto-mode` objects (no tags/labels/finalizer added); empty `managedBy` keeps existing behavior (finalizer added).
- Regenerated CRD applied to a live EKS 1.36 cluster (server dry-run: accepted); status write of `trunkInterface` validated against the new schema.

### Will this PR introduce any new dependencies?
No

### Will this break upgrades or downgrades? Has updating a running cluster been tested?
No breaking change — additive schema + optional enum field. Existing objects and controllers are unaffected. Downgrade note: an older controller ignores `managedBy` and would reconcile auto-mode-owned CNINodes; however older controllers never encounter such objects since they only exist on Auto Mode clusters running the newer stack.

### Does this PR introduce any user-facing change?
Yes

```release-note
Add status.trunkInterface (trunk and branch ENI details) and spec.managedBy to the CNINode CRD, and enable the status subresource. Empty managedBy is equivalent to vpc-resource-controller; controllers skip CNINodes managed by another controller.
```


apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -31,20 +31,102 @@ type Feature struct {
...
 	Features []Feature `json:"features,omitempty"`
 	// Additional tag key/value added to all network interfaces provisioned by the vpc-resource-controller and VPC-CNI
 	Tags map[string]string `json:"tags,omitempty"`
+	// ManagedBy identifies the controller that owns this CNINode and its
+	// status. Empty is equivalent to "vpc-resource-controller" for backward
+	// compatibility with objects created before this field existed.
+	// Controllers must ignore CNINodes managed by another controller.
+	// +optional
+	// +kubebuilder:validation:Enum=vpc-resource-controller;eks-auto-mode
+	ManagedBy CNINodeManager `json:"managedBy,omitempty"`
```
Issacwww, line 66: Should managedBy be immutable once set? A node can never legitimately flip between EC2 and auto mode, rejecting the mutation at the API level would be cheap insurance.

haouc: Agreed — done in 1a1453e via CRD CEL transition rules (field immutable once set + cannot be removed), with first-time set on existing objects still allowed for adoption of pre-existing CNINodes. Validated on a live 1.36 cluster.

Issacwww (approved):

Issue #634, opened 2026-01-05
VPC-RC taking upwards of 10 minutes, to hydrate its trunk cache when restarted at larger node size


yash97:
**What would you like to be enhanced**:

**Why is the change needed and what use case will it solve**:

Did experiment with 1700 nodes, restarted vpc-rc and track time it took to send event to that it manages this node, and calculated summary statistic for latency

Some raw numbers

=== Summary (seconds since T0 -> latest NodeTrunkInitiated per node) ===
nodes_total=1700 nodes_with_event=1700 nodes_missing=0
mean=384.334  p50=384.500  p90=667.200  p95=703.000  p99=732.000  max=737.000

SaranBalaji90:
## Root Cause Analysis

After investigating this issue, there are **two independent bottlenecks** causing the slow trunk cache hydration on restart at scale. Thanks to @oliviassss for performing the benchmark tests.

### Bottleneck 1: Global Lock in `AddNode` / `UpdateNode` Serializes Node Processing

`pkg/node/manager/manager.go` — `AddNode()` and `UpdateNode()` acquire an **exclusive lock** (`manager.lock`) and hold it across multiple Kubernetes API calls:

```go
func (m *manager) AddNode(nodeName string) error {
    m.lock.Lock()
    defer m.lock.Unlock()

    K8sAPI.GetNode(nodeName)               // network I/O under lock
    CreateCNINodeIfNotExisting(node)       // network I/O under lock (GetCNINode + maybe CreateCNINode)
    isSelectedForManagement(node)          // network I/O under lock (GetCNINode)
    updateSubnetIfUsingENIConfig(node)     // network I/O under lock (GetENIConfig)

    dataStore[name] = newNode              // only this needs the lock
    worker.SubmitJob(...)
}
```

Even with `--max-node-reconcile=10` (10 concurrent reconcilers), only **one** can hold the exclusive lock at a time. The others block on `m.lock.Lock()`. Each `AddNode` takes ~200-400ms (3-4 K8s API calls), so nodes are effectively submitted to the worker queue at ~2-3/sec regardless of reconciler concurrency.

For 1700 nodes: `1700 × ~300ms ≈ 8.5 minutes` just to get all Init jobs into the worker queue.

### Bottleneck 2: EC2 Client-Side Rate Limiter Caps Worker Throughput

[...]

PR #690, opened 2026-08-21
Zero-EC2 controller restart: CNINode checkpoint re-init + onboarding observability

Issacwww:
**Stack 1/3** (supersedes the long-running #679 with a cleaner, reviewable stack — happy to close #679 in favor of this series).

**Issue: a controller restart / leader change makes O(nodes) EC2 describe calls (#634), causing throttling storms and pod startup spikes at fleet scale.**

Contents:
- The CNINode checkpoint re-init work from #679 (M0): persist a controller-private restart-recovery snapshot in `status.reinitCheckpoint`, hydrate in-memory state from it with EC2 as fallback truth (any invalid/missing field = cache miss -> EC2 path).
- Merged with current `release-1.7.16-onwards` (#688/#689 — the #687 subnetID schema this branch had already aligned to byte-for-byte; the two CRD-yaml conflicts resolved as the union, mirror-consistency enforced by `TestEmbeddedCRDsMatchGenerated`).
- Observability to prove the zero-EC2 property at scale: `node_onboarding_latency{result}` (both Init submission paths), `node_manager_lock_wait_latency{operation}`, `worker_queue_depth{resource_name}` — observation points pinned by unit tests.

Scale evidence (1500-node beta cluster): restart hydrates 1500/1500 via the fast path with zero per-node describes.

**Review scope: the two commits after the merge base.** Stack: **1/3 (this)** <- stack-2 (orphan/VLAN lifecycle) <- stack-3 (collision metric).


Issacwww:
Recreating as a properly stacked PR series (in-repo branches, chained bases) so each PR's diff shows only its own slice and merge dependency is enforced — links to follow.

Issue #515, opened 2025-02-06
CNINode resources' name should refecting to node's instance ID


haouc:
**What would you like to be enhanced**:
Currently the CNINode resources' name are node hostname which is likely reused by EC2. To be more resilient, using instance ID should be a better way to prevent collision.

**Why is the change needed and what use case will it solve**:
Hypothetically, if the resources were leaked, the name collision would cause breaking behavior on this controller.

zshen-figma:
Hi do we have timeline on solving this issue?

cdirubbio:
A workaround is using https://awslabs.github.io/amazon-eks-ami/nodeadm/doc/examples/#using-instance-id-as-node-name in the meantime
```
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="BOUNDARY"

--BOUNDARY
Content-Type: application/node.eks.aws

---
apiVersion: node.eks.aws/v1alpha1
kind: NodeConfig
spec:
  featureGates:
    InstanceIdNodeName: true

--BOUNDARY--
```

Note, this requires using an Access Entry of type `EC2` rather than `EC2_LINUX` which is the default. This is because the username needs to be the session name (instance id), rather than the private IPv4 DNS Name. If you are using Managed Node Groups, you will need to delete the auto-created Access Entry of type `EC2_LINUX` and create a new one for `EC2`. For Karpenter, similarly you need to create the `EC2` access entry, and here is example nodeclass configuration:
```
apiVersion: karpenter.k8s.aws/v1beta1
kind: EC2NodeClass
metadata:
  name: al2023
spec:
  amiFamily: AL2023
...
  userData: |
    MIME-Version: 1.0
    Content-Type: multipart/mixed; boundary="//"

    --//
    Content-Type: application/node.eks.aws

    apiVersion: node.eks.aws/v1alpha1
    kind: NodeConfig
    spec:
      featureGates:
        InstanceIdNodeName: true

    --//
```

PR #692, opened 2026-08-21
Measure CNINode name collisions after EC2 private-IP reuse

Issacwww:
**Stack 3/3 — depends on stack-2 (contains its commits; review ONLY the last commit: `feat(metrics): Measure CNINode name collision after private-IP reuse (#515)`).**

**Issue: a leaked CNINode + EC2 private-IP reuse hands a new instance a stale CNINode (#515). The eventual fix (instance-id naming) is a disruptive schema migration — ship measurement first so that decision is data-driven.**

- `cninode_instance_id_collision_total{source=bind|hydrate}`: bind-time detection in the CNINode reconciler (live instance ID parsed from `spec.providerID` with strict shape validation) plus the existing hydrate-mismatch site feeding the same counter.
- Detection/measurement only: never fails a reconcile, silently skips on missing/unparseable providerID, hydrate fallback behavior unchanged, no CRD/schema change.

Scale evidence: collision scenario constructed on a 1500-node beta cluster — detected, EC2 fallback taken, counter incremented (S5 acceptance PASS).


Issacwww:
Recreating as a properly stacked PR series (in-repo branches, chained bases) so each PR's diff shows only its own slice and merge dependency is enforced — links to follow.

Issue #13, opened 2020-12-04
VLAN ID is not freed from cache becasuse as EC2 is evenutally consistent


abhipth:
**Describe the Bug**:
When a pod using Branch ENI is created and the controller is immediately restarted since EC2 is eventually consistent the EC2 API call can return ENI list without the newly created ENI leading to the controller not marking the VLAN ID allocated to the ENI as assigned in it's internal cache. 

This can lead to new Branch ENI created on the node to fail since the controller will try to reuse the existing VLAN ID which it's not aware of.

**Proposed Fix**:
Check the error message, if the error message says that the VLAN ID is still in use, add the VLAN ID to cache.
```
_, err = t.ec2ApiHelper.AssociateBranchToTrunk(&t.trunkENIId, nwInterface.NetworkInterfaceId, vlanID)
	if err != nil {
                // Check error here if VLAN is already used mark it in cache.
		trunkENIOperationsErrCount.WithLabelValues("associate_branch").Inc()
		break
	}
```

**How to reproduce it (as minimally and precisely as possible)**:
Can be reproduced on trying multiple times to create Branch ENIs for new pods and kill the controller just after new Branch ENIs are created.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.0.5
- CNI Version: v1.7.5

PR #698, opened 2026-08-21
Restore stable node state and recover branch inventory on demand (#634)

Issacwww:
## Goal

Reduce controller restart recovery from eager EC2 discovery on every managed
Node to checkpoint restore plus work required by active allocations.

## Scope

This PR addresses Issue 634 only.

- Stable instance and trunk state is restored from
  `CNINode.status.nodeNetworkState`.
- Complete Branch ENI/VLAN state is recovered once, immediately before the
  first new allocation on each restored trunk.
- CNINode owner freshness and same-name Node/IP reuse are handled separately in
  Issue 515, for both restart and non-restart paths.
- This PR does not add duplicate-VLAN quarantine or a separate orphan-cleanup
  queue.

The detailed boundary is documented in
`docs/restart-recovery-and-ip-reuse.md`.

## Restart recovery

At the checkpoint consumption boundary, the controller compares the persisted
instance ID with the current Kubernetes Node provider ID.

- A valid checkpoint restores stable instance and trunk state without eager
  Branch ENI discovery.
- A missing or invalid checkpoint, including an instance-ID mismatch, falls
  back to authoritative EC2 initialization and rewrites fresh status.
- Custom-networking subnet CIDRs use a process-wide cache with concurrent
  lookup coalescing.

A trunk restored from the checkpoint starts with:

```text
branchStateVerified = false
```

Existing workloads can continue using their restored ownership state. Before a
new Branch ENI allocation can select a VLAN, the controller:

1. Lists Pods on the Node.
2. Describes all controller-managed Branch ENIs for the trunk, without a
   current-subnet filter.
3. Reserves every VLAN observed in EC2.
4. Maps ENIs claimed by usable Pod annotations into the local ledger.
5. Places remaining unowned ENIs on the existing delete queue.
6. Atomically commits the complete ledger and marks the trunk verified.

A malformed Pod annotation is not accepted as ownership. Every VLAN observed
in EC2 remains reserved until its ENI is deleted. If recovery fails, the trunk
remains unverified and the Pod allocation retry performs a later attempt.

A per-trunk gate coalesces concurrent Pod requests, so only one successful
recovery Describe runs for that trunk.

## EC2 call behavior

| Path | EC2 behavior |
|---|---|
| Standard-networking checkpoint hit | No eager per-Node instance/trunk/branch discovery |
| Custom-networking checkpoint hit | At most one subnet lookup per unique subnet per controller process |
| First new allocation on a restored trunk | One paginated Branch ENI discovery for that trunk |
| Missing or invalid checkpoint | Existing authoritative cold initialization |

## Testing

- `make presubmit`, including full race tests.
- 100-Node warm restart:
  - 100 checkpoint hits and 100 restored trunks;
  - zero `DescribeInstances` and `DescribeNetworkInterfaces` calls;
  - four `DescribeSubnets` calls for four unique custom-networking subnets.
- Two concurrent first allocations on one restored trunk shared one recovery.
  A later allocation on the verified trunk performed no additional recovery
  Describe.
- Yash transitional-ENI scenario:
  - six old ENIs still occupied VLANs during leader failover;
  - the new leader recovered the trunk inventory and found all six unowned ENIs
    before creating replacement ENIs;
  - all replacement Pods became Ready;
  - no `InvalidVlanId.Duplicate` or duplicate-triggered allocation failure
    occurred;
  - all six old ENIs were deleted after cooldown.
- SGP allow/deny canaries passed.
- Final EC2 inventory returned to 1115 healthy Branch ENIs, with zero orphan and
  zero invisible-to-restore ENIs.


apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,82 @@ spec:
...
+                  checkpoint: the instance fields needed to rebuild in-memory state on
+                  controller restart or leader change without synchronous EC2 describes.
+                  Written only by the vpc-resource-controller; other controllers must
+                  neither write nor depend on it.
+                properties:
+                  instance:
+                    description: Instance stores the EC2 instance fields needed on
+                      hydrate.
+                    properties:
+                      connectionTracking:
```
SaranBalaji90: Why do we need connection tracking to be stored here?

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,82 @@ spec:
...
+                          type: string
+                        type: array
+                      currentSubnetCIDRBlock:
+                        type: string
+                      currentSubnetID:
+                        type: string
+                      currentSubnetV6CIDRBlock:
+                        type: string
+                      instanceID:
+                        type: string
```
SaranBalaji90: How do we use these? need better naming and description if used

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +384,145 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
 
 	return i.tcpEstablishedTimeout, i.udpStreamTimeout, i.udpTimeout
 }
+
+func (i *ec2Instance) HydrateFromCNINodeStatus(status rcv1alpha1.CNINodeStatus) (bool, HydrateResult) {
+	i.lock.Lock()
```
haouc, line 400: Is that possible to be handled concurrently per instance? My mental model is they are picked individually by workers.

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +364,54 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+	defer i.lock.Unlock()
+
+	status := cniNode.Status
+	trunk := status.TrunkInterface
+
+	i.instanceType = status.InstanceType
+	i.currentSubnetID = trunk.SubnetID
+	i.currentSubnetCIDRBlock = trunk.SubnetCIDR
+	i.currentSubnetV6CIDRBlock = trunk.SubnetV6CIDR
```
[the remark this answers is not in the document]
Issacwww: Ack, The checkpoint stores the derived values, but UpdateCurrentSubnetAndCidrBlock re-derives from instanceSubnet*/primaryENISecurityGroups, which hydrate never fills — so the first Update wipes everything.

Fix: checkpoint the source fields instead, and have hydrate fill them then derive, exactly like LoadDetails. Derivation/update flow untouched; adding a hydrate→Update regression test.

Trade-off: custom-networking nodes still need one GetSubnet on restart (ENIConfig has no CIDR). Persisting the derived CIDR too would avoid it but adds fields and can't help if the ENIConfig changed during downtime. my call is we can have fast follow to reduce the call for custom networking case in separate PR. @haouc @SaranBalaji90 wdyt?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -484,6 +560,10 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
 		if err != nil {
 			err = fmt.Errorf("associating branch to trunk, %w", err)
 			trunkENIOperationsErrCount.WithLabelValues("associate_branch").Inc()
+			// Associate can fail because an orphan branch ENI (attached in EC2 but
+			// tracked by no pod) already holds a VLAN or trunk slot. Reclaim orphans
+			// so a subsequent pod retry can allocate against the freed capacity.
+			t.reclaimOrphansOnAssociateFailure()
```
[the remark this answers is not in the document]
Issacwww: fair, but not blocking the correctness. handle as fast follow

pkg/provider/branch/trunk/trunk.go
```diff
@@ -504,6 +584,84 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
+	t.lock.RLock()
+	owned := make(map[string]struct{})
+	for _, enis := range t.uidToBranchENIMap {
+		for _, eni := range enis {
+			owned[eni.ID] = struct{}{}
+		}
+	}
+	for _, eni := range t.deleteQueue {
+		owned[eni.ID] = struct{}{}
+	}
```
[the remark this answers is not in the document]
Issacwww: have no data for occurrence, add metric for now and defer to next PR if fix needed

pkg/provider/branch/trunk/trunk.go
```diff
@@ -504,6 +584,84 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
+		if err != nil || vlanID <= 0 || vlanID >= MaxAllocatableVlanIds {
+			t.log.Info("skipping orphan branch eni with unusable vlan tag", "eni", id, "vlanID", vlanID)
+			continue
+		}
+		t.log.Info("reclaiming orphan branch eni owned by no pod", "eni", id, "vlanID", vlanID)
+		branchENIOrphanReclaimCount.WithLabelValues("reclaimed").Inc()
+		// Mark the VLAN used so the ledger does not hand it out again before the
+		// orphan is deleted, then enqueue the orphan for deletion.
+		t.markVlanAssigned(vlanID)
+		t.pushENIToDeleteQueue(&ENIDetails{ID: id, VlanID: vlanID, deletionTimeStamp: time.Now()})
```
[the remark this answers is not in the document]
Issacwww: have no data for occurrence, add metric for now and defer to next PR if fix needed

pkg/provider/branch/trunk/trunk.go
```diff
@@ -370,6 +400,39 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
...
+	for _, pod := range podList {
+		pod := pod
+		eniList := t.getBranchInterfacesUsedByPod(&pod)
+		if len(eniList) == 0 {
+			continue
+		}
+		for _, eni := range eniList {
+			t.markVlanAssigned(eni.VlanID)
+		}
+		t.uidToBranchENIMap[string(pod.UID)] = eniList
```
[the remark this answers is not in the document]
Issacwww: Adding validation for next commit

pkg/node/node.go
```diff
@@ -55,8 +60,34 @@ type node struct {
...
+	fastPathReasonUnsupportedType    = "unsupported_type"
 )
 
+var (
+	cniNodeStatusFastPathCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "cninode_status_fast_path_total",
+			Help: "Number of trunk re-init attempts served from the CNINode status snapshot (result=hit, zero EC2) or fallen back to EC2 discovery (result=miss), labeled by miss reason",
+		},
+		[]string{"result", "reason"},
```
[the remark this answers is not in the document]
Issacwww: will add one latency metric to prove the benefit: `node_init_duration_seconds`  — measures a single node's InitResources duration, from the manager's Init job entering until instance details + trunk + providers are ready .

pkg/aws/ec2/instance.go
Issacwww: Yeah, I think your mental model is right, so this lock is probably redundant. I can remove it.

but i think this added initially due to`LoadDetails()` does the same thing though — should we leave that as-is, or remove that lock too?

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
           status:
             description: CNINodeStatus defines the managed VPC resources.
             properties:
+              reinitCheckpoint:
+                description: |-
+                  ReinitCheckpoint is the vpc-resource-controller-private restart-recovery
+                  checkpoint: the instance fields needed to rebuild in-memory state on a
+                  controller restart or leader change without synchronous EC2 describes.
+                properties:
+                  connectionTracking:
```
Issacwww, line 97: This is needed when create branch eni

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                        description: UDPStreamTimeout is the timeout for UDP flows
+                          classified as streams.
+                        format: int32
+                        type: integer
+                      udpTimeout:
+                        description: UDPTimeout is the timeout for idle UDP flows.
+                        format: int32
+                        type: integer
+                    type: object
+                  currentInstanceSecurityGroups:
```
Issacwww: The reason we have current* fields is, they are needed for branch eni creation.  Most of the case the value will be same as instance* value, but for
custom networking use case. it will be override by the ENIConfig + a describe subnet call for CIDR

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                      InstanceID is the EC2 instance id this checkpoint belongs to; hydrate
+                      rejects the checkpoint if it does not match the live node (a reused name).
+                    maxLength: 19
+                    pattern: ^i-([0-9a-f]{8}|[0-9a-f]{17})$
+                    type: string
+                  instanceSubnetCIDRBlock:
+                    description: InstanceSubnetCIDRBlock is the IPv4 CIDR block of
+                      the instance's own subnet.
+                    type: string
+                  instanceSubnetID:
```
Issacwww: all instance* fields are original metadata when ec2 spin up.

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
           status:
             description: CNINodeStatus defines the managed VPC resources.
             properties:
+              reinitCheckpoint:
```
SaranBalaji90: For better readability - change it to NodeNetworkState

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                description: |-
+                  ReinitCheckpoint is the vpc-resource-controller-private restart-recovery
+                  checkpoint: the instance fields needed to rebuild in-memory state on a
+                  controller restart or leader change without synchronous EC2 describes.
+                properties:
```
SaranBalaji90: NodeNetworkState contains the controller-owned metadata needed to reconstruct a node's in-memory networking resources after it restart or leader change without an immediate EC2 describe.

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                        type: integer
+                      udpTimeout:
+                        description: UDPTimeout is the timeout for idle UDP flows.
+                        format: int32
+                        type: integer
+                    type: object
+                  currentInstanceSecurityGroups:
+                    description: |-
+                      CurrentInstanceSecurityGroups are the effective default security groups
+                      applied to a branch ENI when the pod does not specify its own.
```
SaranBalaji90: How is a branch ENI allocated for a pod that is not covered by a SecurityGroupPolicy (SGP) CR?

Also, a single instance can have branch ENIs allocated for pods covered by multiple SGPs, each with different security groups and CIDRs. If an instance has 10 such SGPs, what should these fields represent, and which values should they store?

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                    items:
+                      type: string
+                    type: array
+                    x-kubernetes-list-type: atomic
+                  currentSubnetCIDRBlock:
+                    description: CurrentSubnetCIDRBlock is the effective IPv4 CIDR
+                      block for branch ENIs.
+                    type: string
+                  currentSubnetID:
```
SaranBalaji90: "current" sounds odd, lets change all these to BranchENISubnetId, BranchENISubnetCIDR, etc

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                      instance's primary ENI (the source-of-truth value).
+                    items:
+                      type: string
+                    type: array
+                    x-kubernetes-list-type: atomic
+                  trunkENIID:
+                    description: TrunkENIID is the trunk ENI the branch ledger is
+                      rebuilt against.
+                    type: string
+                type: object
```
SaranBalaji90: There is already one under TrunkInterfaces

mocks/amazon-vcp-resource-controller-k8s/pkg/aws/ec2/mock_instance.go
```diff
@@ -119,6 +134,20 @@ func (mr *MockEC2InstanceMockRecorder) GetHighestUnusedDeviceIndex() *gomock.Cal
...
+// HydratedTrunkID mocks base method.
+func (m *MockEC2Instance) HydratedTrunkID() string {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "HydratedTrunkID")
+	ret0, _ := ret[0].(string)
+	return ret0
+}
+
+// HydratedTrunkID indicates an expected call of HydratedTrunkID.
+func (mr *MockEC2InstanceMockRecorder) HydratedTrunkID() *gomock.Call {
```
SaranBalaji90: Hydrated is bit internal, change it to something easier to grok when reading the code.

pkg/aws/ec2/instance.go
```diff
@@ -93,6 +101,10 @@ type EC2Instance interface {
 	GetCustomNetworkingSpec() (subnetID string, securityGroup []string)
 	UpdateCurrentSubnetAndCidrBlock(helper api.EC2APIHelper) error
 	GetConnectionTrackingSpec() (tcpEstablishedTimeout, udpStreamTimeout, udpTimeout *int32)
+	LoadFromCheckpoint(checkpoint rcv1alpha1.ReinitCheckpoint)
```
SaranBalaji90: LoadFromCNINode?

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +384,101 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
 
 	return i.tcpEstablishedTimeout, i.udpStreamTimeout, i.udpTimeout
 }
+
+// maskFromCIDR returns the prefix length ("16" for "10.0.0.0/16") of a CIDR, or
+// "" if the input is empty or not in CIDR form. Mirrors how LoadDetails derives
+// the subnet mask, so it is not persisted in the checkpoint.
+func maskFromCIDR(cidr string) string {
```
SaranBalaji90: prefixLengthFromCIDR

pkg/node/node.go
```diff
@@ -194,9 +259,126 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
 	n.ready = true
+	initResult = initResultOK
 	return errInit
 }
 
+// validateCheckpoint reports whether the reinit checkpoint may be hydrated for the
+// given live node, returning "" when it may be and the miss reason otherwise. It is
+// pure and Kubernetes-local: no EC2 call, and deliberately no EC2 freshness check.
+//
+// It asks three questions, and nothing else:
```
SaranBalaji90: Fix these writeup to read well - eg remove "and nothing else"

pkg/node/node.go
```diff
@@ -194,9 +259,126 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+}
+
+// tryHydrateFromCheckpoint rebuilds the instance details from the node's reinit
+// checkpoint without any EC2 call, returning true on success. On any miss it
+// leaves the instance untouched (so the EC2 LoadDetails fallback runs on a
+// pristine instance) and records the reason on cninode_status_fast_path_total.
+func (n *node) tryHydrateFromCheckpoint() bool {
+	nodeName := n.instance.Name()
+
+	cniNode, err := n.k8sAPI.GetCNINode(types.NamespacedName{Name: nodeName})
```
SaranBalaji90, line 317: Can't we use the CNINode watcher?

pkg/node/node.go
```diff
@@ -194,9 +259,126 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+		n.log.Info("CNINode is managed by another controller, falling back to EC2",
+			"managedBy", cniNode.Spec.ManagedBy)
+		cniNodeStatusFastPathCount.WithLabelValues(fastPathResultMiss, fastPathReasonNotManaged).Inc()
+		return false
+	}
+
+	// Best-effort Kubernetes Node instance-type label (cache read, no EC2), used
+	// only when present to catch a checkpoint that belongs to a different type.
+	nodeInstanceType := ""
+	if k8sNode, err := n.k8sAPI.GetNode(nodeName); err == nil && k8sNode != nil {
```
SaranBalaji90: NodeWatcher?

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
```diff
@@ -91,11 +91,98 @@ spec:
...
+                    type: object
+                  currentInstanceSecurityGroups:
+                    description: |-
+                      CurrentInstanceSecurityGroups are the effective default security groups
+                      applied to a branch ENI when the pod does not specify its own.
+                    items:
+                      type: string
+                    type: array
+                    x-kubernetes-list-type: atomic
+                  currentSubnetCIDRBlock:
```
SaranBalaji90: what happens when ENIConfig associated with the node changes? Would this make VPC RC to use old SecGp and Subnets?

SaranBalaji90 (commented):
initial review

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
Issacwww: Fair point, the original idea to add it is to avoid describe subnet call for custom networking, but I think this  is not the most high priority one, we can still have the ec2 call for only custom networking scenario to populate these values

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
Issacwww: this  is not branch eni level data, lets sync for alignment

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
Issacwww: > How is a branch ENI allocated for a pod that is not covered by a SecurityGroupPolicy (SGP) CR?

For non-SGP pod, branch ENI won't be allocated. it will use secondary IP from Primary/Secondary ENI or prefix if PD enabled。

For SGP, this value will be [a fall back](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/provider/branch/trunk/trunk.go#L421-L423) if the SGP policy is not specified

> Also, a single instance can have branch ENIs allocated for pods covered by multiple SGPs, each with different security groups and CIDRs. If an instance has 10 such SGPs, what should these fields represent, and which values should they store?

This does not store Branch ENI level data

----

I think multiple comments are related to this, my 2cents are we can drop `current*` fields, and the trade off is to call `DescribeSubnet` for custom networking scenario. let's sync for alignment

pkg/node/node.go
Issacwww: this already wrapped with  cache, https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/ce7277e063a4ac817aba4e13e0858a3c5dab1958/pkg/k8s/wrapper.go#L139 is that what you are referring to?

pkg/node/node.go
Issacwww: We already watch owned CNINodes via [Owns(&v1alpha1.CNINode{})](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/cc3fbcfd0dfa8bfa570f68268bea85ff754ff38e/controllers/core/node_controller.go#L122-L126). This [GetCNINode](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/cc3fbcfd0dfa8bfa570f68268bea85ff754ff38e/pkg/k8s/wrapper.go#L233-L238) reads from the same informer cache backed by mgr.GetClient(), not directly from the API server.

pkg/node/node.go
Issacwww: Same  as above. The Node is already watched via [For(&corev1.Node{})](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/cc3fbcfd0dfa8bfa570f68268bea85ff754ff38e/controllers/core/node_controller.go#L122-L126), and [GetNode](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/cc3fbcfd0dfa8bfa570f68268bea85ff754ff38e/pkg/k8s/wrapper.go#L141-L146) reads from that shared informer cache

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
Issacwww: This is needed when create branch eni, save it to avoid ec2 call

apis/vpcresources/crds/vpcresources.k8s.aws_cninodes.yaml
Issacwww: resolved in #703

pkg/aws/ec2/instance.go
```diff
@@ -118,6 +130,23 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
...
+	i.primaryENIID = ""
+	i.tcpEstablishedTimeout = nil
+	i.udpStreamTimeout = nil
+	i.udpTimeout = nil
+	i.currentSubnetID = ""
+	i.currentSubnetCIDRBlock = ""
+	i.currentSubnetV6CIDRBlock = ""
+	i.currentInstanceSecurityGroups = nil
+	i.restoredFromNodeNetworkState = false
+	i.restoredTrunkENIID = ""
```
SaranBalaji90: Have you thought about grouping replaceable state like

  type instanceNetworkState struct {
      subnetV6CIDR string
      primarySGs   []string
      primaryENIID string
      // ...
  }

  type restoreState struct {
      restored bool
      trunkID  string
  }

  Then rebuilding becomes:

  i.networkState = instanceNetworkState{}
  i.restoreState = restoreState{}

Adding new fields in future might be easier, wdyt

pkg/aws/ec2/instance.go
```diff
@@ -353,3 +381,87 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+// LoadFromNodeNetworkState loads the persisted EC2 values needed to restore the
+// instance. Effective subnet and security group values are derived separately
+// from the current ENIConfig. Device indexes and the primary ENI id remain
+// unset because restored nodes already have a trunk, and Windows nodes do not
+// use this restoration path.
+func (i *ec2Instance) LoadFromNodeNetworkState(state rcv1alpha1.NodeNetworkState, instanceType string, trunkENIID string) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	i.instanceType = instanceType
```
SaranBalaji90: Same like previous one - 

  type instanceSourceState struct {
  	instanceType       string
  	subnetID           string
  	subnetCIDR         string
  	subnetV6CIDR       string
  	primarySGs         []string
  	connectionTracking connectionTrackingState
  }

  type effectiveNetworkState struct {
  	subnetID       string
  	subnetCIDR     string
  	subnetV6CIDR   string
  	securityGroups []string
  }

  Restoration then becomes:

  func (i *ec2Instance) LoadFromNodeNetworkState(
  	state rcv1alpha1.NodeNetworkState,
  	instanceType, trunkENIID string,
  ) {
  	i.lock.Lock()
  	defer i.lock.Unlock()

  	i.source = instanceSourceState{
  		instanceType: instanceType,
  		subnetID:     state.SubnetID,
  		subnetCIDR:   state.SubnetCIDRBlock,
  		subnetV6CIDR: state.SubnetV6CIDRBlock,
  		primarySGs:   slices.Clone(state.PrimaryNetworkInterfaceSecurityGroups),
  		connectionTracking: connectionTrackingFrom(state.ConnectionTracking),
  	}

  	// Recomputed from the current ENIConfig.
  	i.current = effectiveNetworkState{}
  	i.restore = restoreState{
  		fromNodeNetworkState: true,
  		trunkENIID:            trunkENIID,
  	}
  }

pkg/aws/ec2/api/helper.go
```diff
@@ -195,13 +201,50 @@ func (h *ec2APIHelper) GetSubnet(subnetId *string) (*ec2types.Subnet, error) {
...
 	return &describeSubnetOutput.Subnets[0], nil
 }
 
+// GetSubnetCIDR returns the IPv4 CIDR for a subnet. Results are cached for the
+// lifetime of the controller process because a subnet's CIDR is immutable.
+func (h *ec2APIHelper) GetSubnetCIDR(subnetId *string) (string, error) {
+	if subnetId == nil || *subnetId == "" {
+		return "", fmt.Errorf("subnet id is empty")
+	}
+	id := *subnetId
```
haouc: nit: looks this local assignment is unnecessary.

pkg/aws/ec2/api/helper.go
```diff
@@ -195,13 +201,50 @@ func (h *ec2APIHelper) GetSubnet(subnetId *string) (*ec2types.Subnet, error) {
...
+	cidr, ok := h.subnetCIDRCache.Load(id)
+	if ok {
+		return cidr.(string), nil
+	}
+
+	value, err, _ := h.subnetCIDRLookupGroup.Do(id, func() (interface{}, error) {
+		cachedCIDR, cached := h.subnetCIDRCache.Load(id)
+		if cached {
+			return cachedCIDR.(string), nil
+		}
```
haouc: Is this lookup redundent if the line 221 doesn't return?

pkg/aws/ec2/instance.go
```diff
@@ -118,31 +139,37 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
...
 	}
 	if instanceSubnet == nil || instanceSubnet.CidrBlock == nil {
 		return fmt.Errorf("failed to find subnet or CIDR block for subnet %s for instance %s",
-			i.instanceSubnetID, i.instanceID)
+			i.source.subnetID, i.instanceID)
 	}
-	i.instanceSubnetCidrBlock = *instanceSubnet.CidrBlock
-	i.subnetMask = strings.Split(i.instanceSubnetCidrBlock, "/")[1]
+	i.source.subnetCIDRBlock = *instanceSubnet.CidrBlock
+	i.source.subnetMask = strings.Split(i.source.subnetCIDRBlock, "/")[1]
```
haouc: I think EC2 API ensure the lenght of their output. Just wonder if you want to do sanity check before indexing 1.

pkg/aws/ec2/instance.go
```diff
@@ -118,31 +139,37 @@ func (i *ec2Instance) LoadDetails(ec2APIHelper api.EC2APIHelper) error {
...
 		return err
 	}
 	if instanceSubnet == nil || instanceSubnet.CidrBlock == nil {
 		return fmt.Errorf("failed to find subnet or CIDR block for subnet %s for instance %s",
-			i.instanceSubnetID, i.instanceID)
+			i.source.subnetID, i.instanceID)
 	}
-	i.instanceSubnetCidrBlock = *instanceSubnet.CidrBlock
-	i.subnetMask = strings.Split(i.instanceSubnetCidrBlock, "/")[1]
+	i.source.subnetCIDRBlock = *instanceSubnet.CidrBlock
```
haouc: Do you want to do same nil check before deref this just same as what you do for v6 below?

pkg/aws/ec2/instance.go
haouc: I think async is ok for this hydrating but not ok for init, wdyt?

pkg/aws/ec2/instance.go
```diff
@@ -351,5 +378,93 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+	i.source = source
+	i.current = effectiveNetworkState{}
+	i.restore = restoreState{
+		fromNodeNetworkState: true,
+		trunkENIID:           trunkENIID,
+	}
+}
+
+// BuildNodeNetworkState returns the EC2 values needed to restore this instance.
+// It is called after authoritative EC2 initialization.
```
haouc, line 439: Sounds like these building work are not time sensitive and can be asynced? Do we need lock if we are not using shared sources? Looks like those data are built into instances?

pkg/aws/ec2/instance.go
```diff
@@ -351,5 +378,93 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+	}
+}
+
+// IsRestoredFromNodeNetworkState reports whether the instance details were
+// restored from NodeNetworkState.
+func (i *ec2Instance) IsRestoredFromNodeNetworkState() bool {
+	i.lock.RLock()
+	defer i.lock.RUnlock()
+
+	return i.restore.fromNodeNetworkState
```
haouc, line 464: Can you confirm `retore`/`fromNodeNetworkState` are per instance object? Just want to understand for these processes do we really need lock.

pkg/node/manager/manager.go
```diff
@@ -498,7 +499,17 @@ func GetNodeInstanceID(node *v1.Node) string {
 	return instanceID
 }
 
-// getNodeOS returns the operating system of a node.
+// GetNodeInstanceType returns the instance type from the stable label, falling
+// back to the beta label used by older Kubernetes versions.
+func GetNodeInstanceType(node *v1.Node) string {
```
haouc, line 504: I think instance type has been fetched, such as https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/aws/ec2/instance.go#L83. Can this be wired with existing code?

pkg/node/node.go
```diff
@@ -55,8 +63,48 @@ type node struct {
...
+	restoreReasonNotManaged         = "not_managed_by_controller"
+	restoreReasonMissingField       = "missing_field"
+	restoreReasonInvalidCIDR        = "invalid_cidr"
+	restoreReasonInstanceIDMismatch = "instance_id_mismatch"
+	restoreReasonSubnetLookupFailed = "subnet_lookup_failed"
+	restoreReasonUnsupportedType    = "unsupported_type"
+
+	// node_init_duration_seconds label values.
+	initResultOK    = "success"
+	initResultError = "error"
```
haouc: I think we should consider at least some of them as enums.

pkg/node/node.go
```diff
@@ -194,9 +254,100 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+		}
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
+			return restoreReasonInvalidCIDR
+		}
+	}
+
+	// Identity: a CNINode name can be reused by a new instance.
+	if state.InstanceID != instanceID {
+		return restoreReasonInstanceIDMismatch
+	}
```
haouc, line 305: This is one of our critical caveats using node name. I wonder how this error is handled? Failing here seems a bit late?

pkg/node/node.go
```diff
@@ -194,9 +254,100 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+
+	// Instance type determines branch ENI capacity. It is read from the
+	// Kubernetes Node label and is no longer persisted in the checkpoint.
+	if nodeInstanceType == "" {
+		return restoreReasonMissingField
+	}
+	if _, ok := vpc.Limits[nodeInstanceType]; !ok {
```
haouc: I think we have multiple places validating node type. Can you check if you have to check again here? I wonder if the code (node/cninode) even reach here if the type is not supported.

pkg/node/node.go
```diff
@@ -194,9 +254,100 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+	if err != nil {
+		cniNodeNetworkStateRestoreCount.WithLabelValues(restoreResultMiss, restoreReasonNoState).Inc()
+		return false
+	}
+
+	// Only the owning controller may consume or update this status.
+	if !cniNode.IsManagedByVPCResourceController() {
+		n.log.Info("CNINode is managed by another controller, falling back to EC2",
+			"managedBy", cniNode.Spec.ManagedBy)
+		cniNodeNetworkStateRestoreCount.WithLabelValues(restoreResultMiss, restoreReasonNotManaged).Inc()
```
haouc: I am not sure if we need a metric to report this since this is a normal operation instead of an error or failure.

pkg/node/node.go
```diff
@@ -194,9 +254,100 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+	state := cniNode.Status.NodeNetworkState
+	if reason := validateNodeNetworkState(cniNode.Status.TrunkInterface, state,
+		n.instance.InstanceID(), n.instanceType); reason != "" {
+		n.log.Info("NodeNetworkState is unusable, falling back to EC2", "reason", reason,
+			"nodeInstanceID", n.instance.InstanceID(), "nodeInstanceType", n.instanceType)
+		cniNodeNetworkStateRestoreCount.WithLabelValues(restoreResultMiss, reason).Inc()
+		return false
+	}
+
+	trunkENIID := cniNode.Status.TrunkInterface.ID
```
haouc, line 346: If the cninode doesn't have the trunk populated in status, if ID is not a pointer, we may have to handle empty, if ID is a pointer, we need to handle nullable field.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -86,15 +93,32 @@ var (
...
+	// trunkReinitCount records whether trunk initialization restored local state
+	// or used authoritative EC2 discovery.
+	trunkReinitCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "trunk_reinit_total",
+			Help: "The number of trunk re-initializations by path (restored|ec2) and result",
+		},
+		[]string{"path", "result"},
+	)
 	prometheusRegistered = false
```
haouc: The PrometheusRegister is unsafe when being processed async. Myaybe use sync.Once(), or atomic.boolean?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -491,19 +727,125 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
+}
+
+// reclaimOrphansOnAssociateFailure runs one successful EC2 check per restored
+// trunk per process. Failed describes remain retryable.
+func (t *trunkENI) reclaimOrphansOnAssociateFailure() {
+	if t.isOrphanCheckCompleted() {
+		branchENIOrphanReclaimCount.WithLabelValues("already_checked").Inc()
+		return
+	}
+	_, _, _ = t.reclaimGroup.Do(t.trunkENIId, func() (interface{}, error) {
```
haouc, line 763: Should we handle errors returned from `.DO()`?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -319,57 +371,212 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
...
+		return ErrInvalidRestoredLedger
+	}
+
+	for _, pod := range podList {
+		pod := pod
+		uid := string(pod.UID)
+		eniList, usable := t.decodeBranchInterfacesUsedByPod(&pod)
+		if !usable {
+			// The EC2 fallback preserves ENIs whose ownership cannot be decoded.
+			return reject("pod branch eni annotation is unusable", "pod", uid)
```
haouc, line 520: Can you measure how frequently this need to fallback to EC2 calls in tests, especially tests with large scale and/or restarting during nodes churning.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -665,6 +1085,37 @@ func (t *trunkENI) addBranchToCache(UID string, branchENIs []*ENIDetails) {
...
+
+// moveReservedBranchENIsToDeleteQueue atomically releases the request's
+// pending-create reservation and accounts for any ENIs that need cleanup.
+func (t *trunkENI) moveReservedBranchENIsToDeleteQueue(reserved int, eniList []*ENIDetails) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	t.pendingCreate -= reserved
+	t.log.Info("pushing ENIs to delete queue", "ENIs", eniList)
+	t.deleteQueue = append(eniList, t.deleteQueue...)
```
haouc, line 1105: Why don't we delete them immediately? From the caller, looks like this only clean up branch ENIs which failed on associate?

pkg/node/manager/manager.go
Issacwww: That instance type won't be populated when we restart as the CNINode does not store instance type.  
I want to save less duplicate info. So this new function is intentional to get this from node label.

Imk if you think using label has concerns and we can add instanceType to the CNINode for recover.

pkg/node/node.go
Issacwww: we need to do the check，this code is been invoked like below.  The label is mutable, so I think we should have the check. 
```
 if tryRestoreFromNodeNetworkState() {
      //restore from node label
      // validate here
  } else {
      LoadDetails() -> //existing behavior: load from ec2 and do validation
  }
```
But I agree we can have a helper function to validate instance types

pkg/aws/ec2/api/helper.go
Issacwww: fixed

pkg/node/node.go
Issacwww: removed metric，just keep log

pkg/aws/ec2/api/helper.go
Issacwww: The second lookup closes the race between the outer cache miss and entering singleflight.Do. A previous flight may complete and populate the cache during that interval. Without the recheck, this caller could start another flight and issue a duplicate DescribeSubnets request.

although the chance to happen is low, just to be defensive

pkg/node/node.go
Issacwww: in current implementation，when hit this scenario , it would fallback to ec2 call to reload the instance detail then proceed. I think we can add metric here for now at least we know how often it happen.

> Failing here seems a bit late?

are you referring we should perform this check at manager level when AddNode / UpdateNode?
I think move there would be better. 2 options
  1.keep using nodename as key, compare the instance id.
if mis-match, invalidate the old cache. and proceed as usual
  2. use instanceid+ip as new cache key to avoid duplication.

For this CR, I do not want to increase the scope, I think we can create a helper function to validate name + instance id and then emit metric. wdyt?

pkg/provider/branch/trunk/trunk.go
Issacwww: Good catch, will use sync Once for this

pkg/provider/branch/trunk/trunk.go
Issacwww: We also need to address the clean up error here, using async way would not blocking new pod allocation.

pkg/provider/branch/trunk/trunk.go
Issacwww: Simplify this to 
1. quarantine VLAN when hit duplicate vlan issue
2. reconcile pod
3. async clean up the orphan ENI and release vlan

pkg/aws/ec2/instance.go
Issacwww: the nil check is covered at line 154

pkg/node/node.go
Issacwww: The nil/empty check has been coverred in validateNodeNetworkState

pkg/node/node.go
Issacwww: new CNINode should have same feature flag for CNI to use - see if there is any reconcile from CNI
T0 node create - T1 cninode create - T2 cni update cninode with feature flag
T3 node controller should reconcile to update node  

option 1: change cache key + force delete 
option 2: new instance id as CNINode name

pkg/node/manager/manager.go
Issacwww: Will Add instance type to CNINode

pkg/node/node.go
Issacwww: Current implementation only handle the restart path when instance id mismatch. Which is when mismatch found between node and CNINode, it will fallback to EC2 calls to get the right metadata.

For non-restart path, when mismatch happen it remain the existing behavior. This does not make things worse. 

I will separate this issue in another PR, and it should not blocking current change to be merged.

I do think using instance id name would be simpler rather than change cache key + force delete. but for current change I will add another instance id check in node controller

Issacwww:
[Orphan clean up logic diagram](https://mermaid.live/edit#pako:eNp1VV1T2zAQ_CsazfSNUIJJKJkOnZAPoE2ZtnTaaR0ehHSOVWQpSHJDSvjvPclxag80eYml27273TvnkXIjgA5opsyK58x68nU81yR-hulQKcOZl0aTjEnlbkinc0rO0s8ls0x7qYF8mw2vCNOCgL4voQTCFTBdLm9qlrOIGaUzwwThpbWgPfG21HcRtgQtpF48g40ibJzOpPPkkxEuRgvpuPkNlhi7zJl2u_BxDJ-kY1Dg4dn1JF5PH7Gh-o5YcEb9BvHuqY6aYhS5MjH2PP0CVUMr6XNyy_idybId4XnVVQv5A1w8vUAoduMa4mRSS5cjuvq-ekXaguNBrfkFMOXz9dtb-_p0ZHStGNtZsXNhZIFhr4F_6JzhMjxNri5vtika-m-uS87BuQ0aMTJFIaOmxKw0WJfLIPsOUCk_fIFkXC6V5CFLaGyD7lwWBYiQV63JfXsoYgPX4MkX4NiGVOhyM03t2BQHCwQK7HkeqncRiC2jy_CA7ofpEJWt0ZAmydbX9AzFuXttt45pWDX1ioTfGfacGUuk7mRKLnJfpUTPvCEOvFct5ul2CsbguJW3EAjJGTa4rbIt8nlb5It06E2BUinUhXFuysrBKExVz6ReF6MzaVHFxszuaC9esKJKFVQjLAsiZQw3xIK3EjD1JVa8sEyAiGnCBq-INrrzzx7RGqYG9WVM9z69RllEqXCN0MvAvCa_zC2uqq0Wt4l5X3Uesmx3ZEM-pJNoX8P5WMzYMqkbBvynjA-V9K2uKyVm6aSeCObWmtdzsTL2DuzzwZg1TmZNi1BsXHQ_RWPEhnxsL6zMSKktZICLx0EgLd2jCysFHWRMOdijBdiChWf6GN4Ac-pzKGBOB_jTonQPHW6UsXM610-IXTL905iCDlA9RFtTLvIdV7kUuD9jydC1YneKqQXYURgcOuglB5GEDh7pAx10-8n-wXH_pN9LksODpHvY26NrPE66-8f9fv8k6XZ7h8lR8uZpj_6JeQ_2j49OjjC62zvp480RInBvvbEfq9d__Bd4-gvJGOQG)

yash97:
can you check this scenario, i was able to repro this:


Leader-change state gap
This is not a new InvalidVlanId.Duplicate bug. The failure is already tracked in [#13](https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/13).

The difference is how state is rebuilt after a leader change.

Current master:

The new leader starts with an empty in-memory cache.
It asks EC2 for all Branch ENIs attached to the trunk.
It restores ENIs used by running Pods.
For remaining ENIs, it marks their VLANs busy and queues them for deletion.
New Pods therefore do not reuse those VLANs.
PR #698 fast path:

The new leader starts with an empty in-memory cache.
It restores the trunk from CNINode.status.
It restores VLAN usage only from running Pod annotations.
It does not restore ENIs waiting for deletion or cooldown.
Their VLANs appear free.
A new allocation reuses one of those VLANs, and EC2 returns InvalidVlanId.Duplicate.
The PR then discovers and cleans up the old ENI reactively.
Therefore, this PR does not introduce the original duplicate-VLAN bug. It adds another path into that known failure by not restoring transitional Branch ENIs.

Reproduction
Tested against PR commit ae79f60; the reconstruction path appears unchanged at current head.

Create six long-running Pods selected by a SecurityGroupPolicy.

Pin them to one trunk-capable node, such as c5.xlarge.

Wait until all six Pods are Ready.

Delete three Pods and immediately create three replacements.

Wait for the replacements to become Ready, but continue before the old Branch ENIs finish cooldown and deletion.

Confirm the three old Branch ENIs still exist in EC2.

Find the active VPC RC leader:

kubectl -n kube-system get lease cp-vpc-resource-controller \
  -o jsonpath='{.spec.holderIdentity}{"\n"}'

Stop only the active leader's VPC RC container. Allow kubelet to restart it normally.

During leader failover, delete the other three original Pods and create three replacements.

After the new leader starts, observe:

It restores only the three currently running Pods.
The six old Branch ENIs still occupy their VLANs in EC2.
New allocation attempts reuse those VLANs.
EC2 returns InvalidVlanId.Duplicate.
Pods receive BranchAllocationFailed and FailedCreatePodSandBox.
Reactive cleanup eventually removes the old ENIs and the Pods become Ready.
Observed result:

PR running normally: replacements Ready in about 2 seconds.
PR with leader restart: six duplicate-VLAN errors; replacements Ready in 37–38 seconds.
Exact base 7f84cf8 with the same restart: no duplicate-VLAN errors; replacements Ready in 34–35 seconds, mostly due to the existing leader-election delay.
For the baseline run, start from a clean ENI state and repeat the same sequence. The EC2-backed initialization should find the old ENIs, reserve their VLANs, and queue them for deletion

yash97:
Follow-up on the leader-change state gap: why are we not populating and restoring status.trunkInterface.branches?

The schema already supports branch ENI id, vlanID, and associationID, but at the current head:

persistCNINodeStatus only writes NodeNetworkState and the trunk ID/subnet.
InitFromNodeNetworkState restores branch/VLAN usage only from running Pod annotations.
When a Pod is deleted, its branch ENI moves into the process-local cooldown/delete queue without updating CNINode status.
Consequently, after a leader change, attached branch ENIs that are cooling down or awaiting deletion disappear from the restored ledger. Their VLANs look free and may be reused while the old ENIs are still attached.

We reproduced this with 500 SGP Pods: scale 500→250, restart the active leader during deletion, then scale 250→500. The new leader restored only the 250 surviving Pod annotations while EC2 still had 500 branch ENIs attached. We observed 210 InvalidVlanId.Duplicate failures. Affected Pods remained Pending/ContainerCreating until reactive cleanup removed the old ENIs.

Should this PR persist every still-attached branch ENI in status.trunkInterface.branches, retain entries until EC2 confirms deletion, and reserve those VLANs during restart recovery? If this field is intentionally not used, what guarantees that the EC2-free checkpoint represents all VLANs still occupied outside running Pod annotations?

This is not claiming that PR #698 introduces the original duplicate-VLAN issue tracked in #13. The concern is that incomplete restart state creates a new leader-change path into that known failure.

yash97:
Key code:

  - pkg/config/loader.go:26 — retry limit is 5.
  - pkg/worker/worker.go:162-176 — job is forgotten after the limit.
  - pkg/provider/branch/provider.go:416-468 — cleanup does not requeue the Pod.



  Pod-ENI allocation has five retries after the initial attempt, giving six total
  attempts. Each failed VLAN is quarantined, so the next attempt generally selects
  another VLAN.
 If all six attempts hit duplicate VLAN errors before cleanup finishes, the create job
 is forgotten. Cleanup may later release the VLANs, but it does not requeue the
  affected Pod. The Pod can therefore remain Pending until another Pod update or the 30-
  minute resync.
 Can orphan-cleanup completion explicitly requeue the affected Pod allocation?

Issacwww:
@yash97 Thanks for catching this. Your reproduction is valid and exposed a gap in the previous recovery design.

The earlier approach restored live Pod ownership from annotations but relied on InvalidVlanId.Duplicate to discover transitional ENIs that were still cooling down or waiting for deletion. That is error-driven: the controller has already selected an occupied VLAN and failed a Pod allocation before recovery begins, which can increase Pod readiness latency and make recovery depend on retries.

The updated approach is on-demand, but not error-driven:

- During restart, the controller restores the stable trunk state and live Pod ownership from the checkpoint and Pod annotations. It does not perform Branch ENI discovery and marks the restored trunk as unverified.
- When the first post-restart allocation reaches that specific trunk, allocation is blocked only on that trunk while its restored ledger is verified against EC2.
- Any additional occupied VLANs are reserved, safely identified unowned ENIs are placed on the existing delete queue, and allocation proceeds after the trunk is marked verified.

The trade-off is an additional EC2 Describe for every restored trunk that later receives a new allocation, even when no stale ENI or VLAN conflict exists. I think this is acceptable because the call is no longer on the controller restart path: idle trunks make no additional EC2 calls, and the cost is proportional to restored trunks receiving new allocations rather than all Nodes.

I validated this with the reported 500→250→restart→500 scenario. The controller restarted while the old 250 ENIs were still cooling down, then the workload was scaled back to 500. All 500 Pods became Ready, no duplicate-VLAN allocation failures or duplicate trunk/VLAN pairs were observed, and the old ENIs eventually converged to zero.

pkg/aws/ec2/instance.go
```diff
@@ -351,5 +378,88 @@ func (i *ec2Instance) GetConnectionTrackingSpec() (tcpEstablished, udpStream, ud
...
+
+	source := instanceSourceState{
+		instanceType:          state.InstanceType,
+		subnetID:              state.SubnetID,
+		subnetCIDRBlock:       state.SubnetCIDRBlock,
+		subnetV6CIDRBlock:     state.SubnetV6CIDRBlock,
+		subnetMask:            subnetMask,
+		subnetV6Mask:          subnetV6Mask,
+		primarySecurityGroups: state.PrimaryNetworkInterfaceSecurityGroups,
+	}
```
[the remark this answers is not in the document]
haouc: This should be fixed since we shouldn't try to release primary ENI.

pkg/aws/ec2/api/helper.go
```diff
@@ -249,7 +297,6 @@ func (h *ec2APIHelper) DescribeNetworkInterfaces(nwInterfaceIds []string) ([]ec2
 		*describeNetworkInterfacesInput)
 }
 
-// TODO: Not used currently as the API is not publicly available with assumed role
 // DescribeTrunkInterfaceAssociation describes all the association of the given trunk interface id
 func (h *ec2APIHelper) DescribeTrunkInterfaceAssociation(trunkInterfaceId *string) ([]ec2types.TrunkInterfaceAssociation, error) {
```
haouc, line 301: Is this a dead func? I am not seeing reference anywhere except tests.

pkg/node/node.go
```diff
@@ -152,16 +218,26 @@ func (n *node) UpdateResources(resourceManager resource.ResourceManager) error {
...
+
+	start := time.Now()
+	initResult := initResultError
+	defer func() {
+		nodeInitDuration.WithLabelValues(string(initResult)).Observe(time.Since(start).Seconds())
+	}()
+
+	// Restore local state first. Any miss falls back to EC2 discovery.
+	if !n.tryRestoreFromNodeNetworkState() {
+		err := n.instance.LoadDetails(n.ec2API)
```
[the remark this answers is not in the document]
haouc: IP providers should be only initialized with Windows nodes. I don't think Windows support have same path and impact as trunk/branch.

pkg/aws/ec2/instance.go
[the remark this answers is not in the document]
Issacwww: Yes fixed

pkg/node/node.go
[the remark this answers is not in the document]
Issacwww: IPv4 or prefix provider are for Windows Node use case, and this is not in the scope for current change.

pkg/node/node.go
```diff
@@ -194,9 +270,88 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+	}
+
+	state := cniNode.Status.NodeNetworkState
+	if reason := validateNodeNetworkState(cniNode.Status.TrunkInterface, state,
+		n.instance.InstanceID()); reason != restoreReasonNone {
+		stateInstanceID := ""
+		stateInstanceType := ""
+		if state != nil {
+			stateInstanceID = state.InstanceID
+			stateInstanceType = state.InstanceType
```
haouc, line 325: Does this mean we are using a CNINode which has unreliable status/metadata since `reason != restoreReasonNone`?

pkg/node/node.go
```diff
@@ -194,9 +270,88 @@ func (n *node) InitResources(resourceManager resource.ResourceManager) error {
...
+		}
+		n.log.Info("NodeNetworkState is unusable, falling back to EC2", "reason", reason,
+			"nodeInstanceID", n.instance.InstanceID(), "stateInstanceID", stateInstanceID,
+			"stateInstanceType", stateInstanceType)
+		if reason == restoreReasonInstanceIDMismatch {
+			RecordInstanceIDMismatch()
+		} else {
+			recordNodeNetworkStateRestore(restoreResultMiss, reason)
```
haouc, line 333: Let's use the same naming pattern as `RecordInstanceIDMismatch`. I am also not sure why we just use one `else` to emit the metrics with (reason).

pkg/node/node.go
```diff
@@ -52,11 +56,72 @@ type node struct {
...
+	})
+}
+
+func recordNodeNetworkStateRestore(result restoreResult, reason restoreReason) {
+	cniNodeNetworkStateRestoreCount.WithLabelValues(string(result), string(reason)).Inc()
+}
+
+// RecordInstanceIDMismatch records a persisted or cached identity mismatch.
+func RecordInstanceIDMismatch() {
+	registerNodeMetrics()
```
haouc, line 121: Looks like using Once.DO() to lazily register the metric. I am not sure if this is better than register them all first since this lazy loading pattern on metrics could be harder to maintain and read for future.

pkg/provider/branch/provider.go
```diff
@@ -130,7 +139,8 @@ func prometheusRegister() {
 	if !prometheusRegistered {
```
haouc, line 139: I understand using a boolean here to gate the registration is not ideal. But if we use this pattern, we should use this for all metrics. I think the contralized pattern for metrics would be easier for read, extend and maintain.

pkg/provider/branch/provider.go
```diff
@@ -181,7 +199,12 @@ func (b *branchENIProvider) InitResource(instance ec2.EC2Instance) error {
 
 		utils.SendNodeEventWithNodeName(b.apiWrapper.K8sAPI, nodeName, utils.NodeTrunkFailedInitializationReason, "The node failed initializing trunk interface", v1.EventTypeNormal, b.log)
 		branchProviderOperationsErrCount.WithLabelValues("init").Inc()
-		return fmt.Errorf("initializing trunk, %w", err)
+		return fmt.Errorf("initializing trunk, %w", initErr)
+	}
+
+	// Persist best-effort state after EC2 initialization.
+	if ec2Path {
```
haouc, line 206: We should move line 171 here: `if !instance.IsRestoredFromNodeNetworkState()` since I am not seeing other usages for the variable.

pkg/provider/branch/provider.go
```diff
@@ -203,6 +226,37 @@ func (b *branchENIProvider) InitResource(instance ec2.EC2Instance) error {
...
+		return
+	}
+	if !cniNode.IsManagedByVPCResourceController() {
+		b.log.Info("skipping NodeNetworkState persistence because another controller manages the CNINode",
+			"node", nodeName, "managedBy", cniNode.Spec.ManagedBy)
+		return
+	}
+	base := cniNode.DeepCopy()
+
+	state := instance.BuildNodeNetworkState()
```
haouc, line 245: Why can't we add trunk info when we BuildNodeNetworkState()?

pkg/provider/branch/provider.go
```diff
@@ -203,6 +226,37 @@ func (b *branchENIProvider) InitResource(instance ec2.EC2Instance) error {
...
+	cniNode.Status.NodeNetworkState = &state
+
+	if cniNode.Status.TrunkInterface == nil {
+		cniNode.Status.TrunkInterface = &rcv1alpha1.TrunkInterface{}
+	}
+	cniNode.Status.TrunkInterface.ID = trunkENI.TrunkENIID()
+	cniNode.Status.TrunkInterface.SubnetID = trunkENI.TrunkSubnetID()
+
+	if err := b.apiWrapper.K8sAPI.UpdateCNINodeStatus(base, cniNode); err != nil {
```
haouc, line 254: Looks like this will not be retried? What is the plan to handle failures?

pkg/provider/branch/trunk/trunk.go
```diff
@@ -367,12 +404,211 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
...
+	if verified {
+		return nil
+	}
+
+	t.branchStateGate.Lock()
+	defer t.branchStateGate.Unlock()
+
+	if t.branchStateVerified {
+		return nil
+	}
```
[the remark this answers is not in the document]
haouc: I think this is a good mitigation on error storms. I do think this can be a fast follow up though since looks like there is no immediate path to fetch-share failure reason.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -209,18 +231,17 @@ func NewTrunkENI(logger logr.Logger, instance ec2.EC2Instance, helper api.EC2API
 }
 
 func PrometheusRegister() {
-	if !prometheusRegistered {
+	prometheusRegisterOnce.Do(func() {
```
haouc, line 234: I think I have seen multiple ways to register metrics. Let's unify them.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -267,8 +289,11 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
 		}
 
 		t.trunkENIId = *trunk.NetworkInterfaceId
+		t.trunkSubnetID = lo.FromPtr(trunk.SubnetId)
+		t.branchStateVerified = true
```
haouc, line 293: I think I saw this field be protected by locks and also no lock protection. Can you recheck if this is shared async and need a locking or not.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -322,28 +347,40 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
 	// From the list of pods on the given node, and the branch ENIs from EC2 API call rebuild the internal cache
 	for _, pod := range podList {
 		pod := pod // Fix gosec G601, so we can use &node
-		eniListFromPod := t.getBranchInterfacesUsedByPod(&pod)
+		eniListFromPod, usable := t.decodeBranchInterfacesUsedByPod(&pod)
+		if !usable {
+			trunkENIOperationsErrCount.WithLabelValues("unusable_pod_eni_annotation").Inc()
```
haouc, line 352: nit: I wonder if we need logging out here otherwise spikes in dashboard could be difficult to trace back through logs.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -367,12 +404,211 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
...
+		eniList, usable := t.decodeBranchInterfacesUsedByPod(&pods[i])
+		if !usable {
+			trunkENIOperationsErrCount.WithLabelValues("unusable_pod_eni_annotation").Inc()
+			continue
+		}
+		uid := string(pods[i].UID)
+		for _, eni := range eniList {
+			if _, found := branchByID[eni.ID]; !found {
+				continue
+			}
```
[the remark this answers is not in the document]
haouc: I am not sure if this is a valid comment. Vland IDs are managed by the controller, I don't think the boudnaries could be broken since we can(do?) have unit tests to guard the logic for unintentional bugs. If we don't have UTs, we should add which will be easier and cheaper than adding validation everywhere.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -367,12 +404,211 @@ func (t *trunkENI) InitTrunk(instance ec2.EC2Instance, podList []v1.Pod) error {
...
+
 // Reconcile reconciles the state from the API Server to the internal cache of EC2 Branch Interfaces, if the controller
 // missed some delete events the reconcile method will perform cleanup for the dangling interfaces
 func (t *trunkENI) Reconcile(pods []v1.Pod) bool {
+	t.branchStateGate.RLock()
+	defer t.branchStateGate.RUnlock()
+
 	// Perform under lock to block new pods being added/removed concurrently
 	t.lock.Lock()
 	defer t.lock.Unlock()
```
haouc, line 614: Let's be very careful on this two layers locking pattern. Can you double check to confirm they are safe.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -402,9 +638,22 @@ func (t *trunkENI) Reconcile(pods []v1.Pod) bool {
 	return leakedENIs > 0
 }
 
-// CreateAndAssociateBranchToTrunk creates a new branch network interface and associates the branch to the trunk
-// network interface. It returns a Json convertible structure which has all the required details of the branch ENI
+// CreateAndAssociateBranchENIs allocates ENIs and commits Pod ownership.
 func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []string, eniCount int) ([]*ENIDetails, error) {
+	t.branchStateGate.RLock()
+	defer t.branchStateGate.RUnlock()
```
haouc, line 644: Can you run deadlock and race validation on these sorts of locks since I think the called funcs in this code block / locking block already have their own locks.

pkg/provider/branch/trunk/trunk.go
```diff
@@ -490,22 +726,38 @@ func (t *trunkENI) CreateAndAssociateBranchENIs(pod *v1.Pod, securityGroups []st
...
+		ID: *nwInterface.NetworkInterfaceId, MACAdd: *nwInterface.MacAddress,
+		IPV4Addr: v4Addr, IPV6Addr: v6Addr, SubnetCIDR: t.instance.SubnetCidrBlock(),
+		SubnetV6CIDR: t.instance.SubnetV6CidrBlock(), VlanID: vlanID,
+	}
+}
+
 // DeleteBranchNetworkInterface deletes the branch network interface and returns an error in case of failure to delete
 func (t *trunkENI) PushBranchENIsToCoolDownQueue(UID string) {
+	t.branchStateGate.RLock()
+	defer t.branchStateGate.RUnlock()
```
haouc, line 759: I think I do have concerns on two mutex locks are operated together which can be easily messed up. Let's scan and test to make sure they are always ordered safely.

pkg/provider/branch/provider.go
Issacwww: As the info we need from BuildNodeNetworkState is coming from EC2Instance, which does not have the trunk info. Are looking for a separate helper to build all both EC2Instance + trunk info together?  IMO, we still need a getter from EC2Instance and a helper does not change too much

pkg/provider/branch/provider.go
Issacwww: Yes, I have added a retry in #711 for the CNINode state write path

pkg/aws/ec2/api/helper.go
Issacwww: good catch,  we can clean up as fast follow

PR #703, opened 2026-08-31
feat: add CNINode nodeNetworkState schema

Issacwww:
## Problem

A controller restart loses the EC2-derived node networking state and the in-memory trunk ledger. Rebuilding every node through EC2 causes a restart-time burst of `DescribeInstances`, `DescribeSubnets`, and `DescribeNetworkInterfaces` calls. This PR defines the CNINode status fields needed by a follow-up change to restore that state from Kubernetes. It also updates the Go directive from 1.26.5 to 1.26.6 because the base version contains reachable standard-library vulnerabilities fixed in 1.26.6.

## Field decisions

- `instanceID` binds the persisted state to the EC2 instance that produced it. The restore path will compare it with the instance ID parsed from `Node.spec.providerID` so state from a replaced instance is not reused.
- `subnetID`, `subnetCIDRBlock`, and optional `subnetV6CIDRBlock` describe the instance and primary ENI subnet. The Kubernetes Node does not provide these CIDRs, and ENIConfig describes the custom-networking subnet rather than the instance subnet.
- `primaryNetworkInterfaceSecurityGroups` supplies the default security groups when ENIConfig does not specify security groups.
- `connectionTracking` preserves optional primary ENI timeout settings needed when new branch ENIs are created after a controller restart.
- `trunkInterface.subnetID` remains separate from `nodeNetworkState.subnetID` because ENIConfig can place the trunk and branch ENIs in a subnet different from the instance and primary ENI subnet.
- Instance type is not persisted. The current value comes from the Kubernetes Node instance-type label, avoiding a second value that can become stale.
- `nodeNetworkState` does not duplicate branch ownership. Pod ENI annotations remain the recovery source for branch ENI and VLAN ownership, and EC2-only orphan ENIs are not represented by this schema.

## Recovery and EC2 behavior

| Scenario | Follow-up controller behavior |
| --- | --- |
| Valid state, standard networking | Restore from the Kubernetes Node, CNINode, and Pod ENI annotations. Skip the startup `DescribeInstances` calls used by instance loading and trunk discovery, the primary-subnet `DescribeSubnets` call, and the `DescribeNetworkInterfaces` call used to discover existing branch ENIs. |
| Valid state, custom networking | Skip instance, trunk, and branch discovery. Retain one `DescribeSubnets` call because ENIConfig contains the selected subnet ID but not its CIDR. |
| New node or missing state | Run the existing full EC2 initialization path, then persist the recovered node and trunk state for a later restart. |
| Invalid state | Do not initialize the node from the persisted state. Run the existing full EC2 initialization path when the instance identity does not match, required recovery inputs are unusable, the current effective subnet conflicts with the observed trunk subnet, or Pod annotations contain malformed or conflicting ENI/VLAN ownership. |
| EC2-only orphan after a valid restore | Do not add proactive EC2 discovery to the normal restart path. The follow-up runtime recovery will handle EC2-only orphans in response to relevant EC2 operation failures; its trigger and error classification are outside this schema PR. |

This schema does not remove normal EC2 mutation calls. New branch ENI allocation still requires `CreateNetworkInterface`, `CreateNetworkInterfacePermission`, and `AssociateTrunkInterface`.

## Review focus

- Is this the minimum persisted field set needed to restore node networking state without standard-networking startup discovery?
- Can any persisted value be obtained reliably from an existing Kubernetes source instead?
- Are the instance subnet and trunk subnet modeled separately where custom networking requires them to differ?
- Do the required and optional fields match the EC2 fallback boundary?
- Does the schema avoid constraining the later orphan-detection implementation?


apis/vpcresources/v1alpha1/cninode_types.go
```diff
@@ -73,6 +73,43 @@ type CNINodeStatus struct {
...
+	// +kubebuilder:validation:Pattern=`^subnet-([0-9a-f]{8}|[0-9a-f]{17})$`
+	SubnetID string `json:"subnetID"`
+	// +required
+	SubnetCIDRBlock string `json:"subnetCIDRBlock"`
+	// +optional
+	SubnetV6CIDRBlock string `json:"subnetV6CIDRBlock,omitempty"`
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=atomic
+	PrimaryNetworkInterfaceSecurityGroups []string `json:"primaryNetworkInterfaceSecurityGroups"`
```
haouc, line 101: Nit: since the SGs are populated by the controller, validation may not be a must. But failing loudly may not be a bad idea.

haouc, line 101: Do you also want secondaryENI's SGs as pods SGs in terms of custom networking?

Issacwww: good catch, added

haouc (approved):

